PR #467, opened 2016-03-17
Fixed payola initializer for migrations.

farukaydin:
When running rake db:migrate command, initializers are loaded before migrations executed. So payola
initializer can not find 'SubscriptionPlan' model and it breaks migrations. With this change we control
'SubscriptionPlan' is defined before.

Without this change, contribution to this project will be painful.


andrew:
Great, thanks @farukaydin, I suspect there will be more things like this where I've not had to start over with a fresh setup for ages. :+1:

PR #491, opened 2016-03-25
Revert quickfixes

andrew:
For merging after the queue has died back down


andrew:
Note to self, also remember to scale down the max in hirefire and reduce the size back to 1X in heroku dashboard

PR #549, opened 2016-04-18
Autofocus to search box: save some tabbing for the mouse-phobic

timruffles:


andrew:
Thanks @timruffles, will deploy that shortly :shipit:

PR #614, opened 2016-06-10
[WebHook] Fix explanation copy.

alloy:
Some words are missing in this sentence. I’m not sure these are the words you were going for, but this at least makes sense to me ¯\_(ツ)_/¯


andrew:
Ah nice catch, that makes much more sense, thanks!

PR #709, opened 2016-07-09
update oj excon rspec

royka:
Upgrade some libs
- https://github.com/librariesio/libraries.io/issues/708
- https://github.com/librariesio/libraries.io/issues/706
- https://github.com/librariesio/libraries.io/issues/705

Ran the specs locally and all seemed good 👍

Kinda off topic but would it be worth increasing the number of specs that libraries has?


royka:
CI has failed! but passes locally honest :) 

Running the specs with  `bundle exec rspec` is that the correct command?

andrew:
Yeah the specs fail sometimes due to elasticsearch being out of date, need to mock that out really.

If you fancy adding any more specs that would be wonderful!

PR #743, opened 2016-07-27
Improve API docs on pagination

zeke:
@jlord and I are working on [a project](https://github.com/zeke/repos-using-electron/blob/master/index.js) to collect information about repos using electron.

We noticed there's no documentation about how to change the number of results per request, so we guessed `per_page` and it works! (Convention over configuration!)

This PR adds a section to the API docs about pagination, and documents the default values.

What is the maximum allowable value of `per_page`?


andrew:
Wonderful, thanks!

andrew:
@zeke max per_page is 100 if you fancy adding that to the docs

PR #759, opened 2016-08-06
Add link in emails to opt out of emails.

blainesch:
I'd prefer not to get emails, but the link in the email was misleading.


andrew:
Thanks!

PR #822, opened 2016-10-04
Add syntax highlighting styles

bengourley:
Code in readmes was already annotated with useful classnames, so this just adds some github-like styles for those elements 🌈.


bengourley:
screen shot 2016-10-04 at 13 24 03

andrew:
Did you write the CSS yourself or get it from GitHub?

Issue #864, opened 2016-11-03
Render the tree view asynchronously 


andrew:
Dependent on #852

Tree views often take over 60 seconds to render for big trees, need to push the rendering to a background job and some javascript to update the page when it's ready so as not to cause timeouts.

- [x] Move tree renderer into a separate class
- [x] Cache tree object into memcached on completion 
- [x] Tree#show should check for cache otherwise enqueue job
- [x] If job enqueued, show spinner and check every few seconds for update

PR #907, opened 2016-11-21
Dependency Tree API endpoint

### LINKED ISSUE #839: Add an API for "resolved transitive dependency tree for any version of any library"
Picking up from [Twitter](https://twitter.com/teabass/status/788827835855175682) ...

This would be super-helpful to us in https://github.com/gratipay/gratipay.com/pull/4135.

andrew:
Fixes #839

Urls like: 
```ruby
'https://libraries.io/api/npm/express/tree?api_key=XXX' # latest version
'https://libraries.io/api/npm/express/4.14.0/tree?api_key=XXX' #specific version
```

Requires an API key, so no anonymous access as it's quite intensive if not cached.

Some big trees may time out if they've not loaded within 60 seconds, need to come up with some way to say "check back in a minute" once #864 is complete.

Not going to add it to the documentation yet as the format may need to change but gives @whit537 something to play with for now.

### DIFF
--- a/app/controllers/api/tree_controller.rb
+++ b/app/controllers/api/tree_controller.rb
@@ -1,0 +1,14 @@
+class Api::TreeController < Api::ApplicationController
+  before_action :require_api_key
+  before_action :find_project
+
+  def show
+    find_version
+    if @version.nil?
+      @version = @project.latest_stable_version
+      raise ActiveRecord::RecordNotFound if @version.nil?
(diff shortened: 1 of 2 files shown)


andrew:
Example response body here: https://gist.github.com/andrew/49da957ae5fe6b1051a3a919a83a2837

whit537:
💃

Issue #1009, opened 2016-12-21
Containing file is missing from dependencies output in the API


tomnatt:
If I look at the dependencies for one of the my projects in the API (eg https://libraries.io/api/github/tomnatt/tomnatt/dependencies?api_key=KEY) I see:

 ```{
"project_name": "jekyll",
"name": "jekyll",
"platform": "Rubygems",
"requirements": "2.5.3",
"latest_stable": "3.3.1",
"latest": "3.3.1",
"deprecated": false,
"outdated": true
},
```

and:
 
 ```{
"project_name": "jekyll",
"name": "jekyll",
"platform": "rubygems",
"requirements": ">= 0",
"latest_stable": "3.3.1",
"latest": "3.3.1",
"deprecated": false,
"outdated": false
},
```

One is (I think) from my Gemfile and the other from my Gemfile.lock however there is no way to tell the difference. 

Can we have the filename in the output please?

andrew:
Just deployed this 🚢

PR #1028, opened 2016-12-29
Added source of dependency to API output

tomnatt:
Fixes https://github.com/librariesio/libraries.io/issues/1009


andrew:
Just deployed this 🚢

Issue #1055, opened 2017-01-09
Bring back turbolinks


andrew:
We used to have turbolinks 2 on the site, was great but caused some issues with google analytics tracking. 

Turbolinks 5 is out and is wonderful, we should try it again and see how it works with GA and adsense.

samjacobclift:
Think we need to do something like this gist https://gist.github.com/esBeee/545653241530f8f2c2e16371bec56f20

PR #1056, opened 2017-01-10
enable turbolinks5

samjacobclift:
**headlines**

1) Enable turbolinks again, see (https://github.com/librariesio/libraries.io/issues/1055)

**TODOs**

- [ ] It seems that that event is fired multiple times per link click, so need to check that out

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -153,9 +153,5 @@
     <%= yield %>
   </div>
   <%= render 'layouts/footer' %>
-  <%= javascript_include_tag 'application', 'data-turbolinks-track' => true %>
-  <% if Rails.env.production? %>
-    <script>ga('send', 'pageview');</script>
-  <% end %>
 </body>
 </html>
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -4,7 +4,7 @@
   <meta http-equiv="X-UA-Compatible" content="IE=edge">
   <meta name="msvalidate.01" content="3667D2FEB238C1BF174F3CB1AE0D3C16" />
   <meta name="viewport" content="width=device-width, initial-scale=1">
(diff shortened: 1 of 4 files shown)


samjacobclift:
I might redo the branch rather that resolve the conflicts :)

andrew:
@samjacobclift need to remove the existing GA tracking in the header of application.html.erb I think

andrew:
Also got some conflicts to resolve but thanks for jumping on this so quickly! ✨

samjacobclift:
@andrew no worries I like learning about turbolinks! its an interesting thing!

andrew:
We may need to review all the other javascript that's loaded on doc ready to be sure it doesn't need to be fired again on turbolinks:load

samjacobclift:
I think so too, there are a few duplicate page views I've been noticing

app/views/layouts/application.html.erb
```diff
@@ -4,7 +4,7 @@
   <meta http-equiv="X-UA-Compatible" content="IE=edge">
   <meta name="msvalidate.01" content="3667D2FEB238C1BF174F3CB1AE0D3C16" />
   <meta name="viewport" content="width=device-width, initial-scale=1">
-  <%= stylesheet_link_tag 'application', media: 'all', 'data-turbolinks-track' => true %>
```
samjacobclift: I think these data-turbolinks-track attrs are causing extra track events, feels like there are a few more elsewhere too

samjacobclift: hmm now looking at it think they can be left in they shouldn't be doing any harm, that being said though something is causing extra 'pageview' events from being fired. The platform page produced some consistently for me

andrew:
@samjacobclift I found what the problem was, turbolinks assumes your javascript to be in the head of the page, adding `, 'data-turbolinks-eval': false` to the script tag at the bottom of the layout fixed the repeated calls to GA, going to deploy this now 🚂

PR #1075, opened 2017-01-16
Allow users to connect multiple accounts from settings page

### LINKED ISSUE #1074: Allow linking your GitHub/Gitlab/Bitbucket accounts together
Now that we have https://github.com/librariesio/libraries.io/pull/1046, we should allow a logged in user to connect all their accounts together from the account section so they can eventually build up a mega-profile

samjacobclift:
**headlines**

1) Lets make it so users can connect multiple bitbucket and gitlab accounts to their user profile!
(also why not multiple github accounts too)

Fixes #1074

**TODO's**

- [x] Can we redirect from the OAuth2 flow back to the settings page for that user (currently when connecting it takes you to the home page)

- [x] Better styling of the connect and connected accounts (any ideas? I can ask the UX guy at work but I'm open to all suggestions!)

- [x] Be able to delete a connected account (do we want this feature, we dont for now)

**latest screenshot**

screen shot 2017-01-16 at 21 51 21

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -2,7 +2,10 @@ class SessionsController < ApplicationController
   skip_before_action :verify_authenticity_token, only: [:create, :failure]
 
   def new
-
+    if !params[:source].nil?
+      session[:pre_login_destination] = '/account'
+      redirect_to "/auth/" + params[:source]
+    end
   end
 
   def enable_public
@@ -46,9 +49,10 @@ class SessionsController < ApplicationController
     end
 
     identity.user.update_repo_permissions_async
+    login_destination = pre_login_destination
 
-    redirect_to(root_path) && return unless pre_login_destination
-    redirect_to pre_login_destination
+    redirect_to(root_path) && return unless login_destination
+    redirect_to login_destination
   end
 
   def destroy
@@ -63,6 +67,8 @@ class SessionsController < ApplicationController
   private
 
   def pre_login_destination
-    session[:pre_login_destination]
+    destination = session[:pre_login_destination]
+    session.delete :pre_login_destination
+    return destination
   end
 end
--- a/app/views/accounts/show.html.erb
+++ b/app/views/accounts/show.html.erb
@@ -38,6 +38,29 @@
     </div>
 
     <div class="panel panel-default">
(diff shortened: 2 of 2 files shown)


andrew:
Sweet, that turned out to be easier than I thought!

I've got some code to redirect back to a location after oauth here: https://github.com/librariesio/libraries.io/blob/master/app/controllers/sessions_controller.rb#L9 I think we can send users through to sessions#new and set the return value there before redirecting them onto /auth/:whatever and the sessions#create action should take care of the rest.

Design wise, I'd combine the two panels and have the "Add" buttons in blue or green as an action

I think we can punt on removing/disconnecting accounts for now until it's requested by someone

samjacobclift:
@andrew hmm might be tricky sending them though sessions#new I'd have to send a param for the source I guess, something like this
`        <%= link_to '/login?source=gitlab', class: 'btn btn-primary' do %>
          <%= fa_icon('gitlab') %> Connect GitLab
        <% end %>`

Then in session#new I could do something like this
`
  def new
    if params[:source]
      session[:pre_login_destination] = request.fullpath
      redirect_to "/auth/" + [:source]
    end
  end
`
Is that what you where thinking of?

Also latest screenshot

screen shot 2017-01-25 at 10 04 18

andrew:
@samjacobclift yeah that's what I was thinking of, in fact it used to work that way but the functionality was removed when I initially added support for multiple logins.

Screenshot is looking great 👍

app/controllers/sessions_controller.rb
```diff
@@ -63,6 +67,8 @@ def failure
   private
 
   def pre_login_destination
-    session[:pre_login_destination]
+    destination = session[:pre_login_destination]
+    session.delete :pre_login_destination
+    return destination
```
samjacobclift: Need to do a bit of cleanup of the pre_login_destination, otherwise it will always redirect to the accounts page

andrew: 👍

andrew:
Thanks @samjacobclift!

andrew:
@samjacobclift I've made a few tweaks here: https://github.com/librariesio/libraries.io/commit/9e10a4fb7f57b7b3cf90f3d7c02c311063db8db6 deploying it now

PR #1112, opened 2017-01-27
Minor changes to readme

BenJam:
I found myself tripping up on the following. It might also be an idea to warn of the Elastic errors before providing the command as I did the classic 'wtf' at that point.


andrew:
👍

PR #1188, opened 2017-02-20
Refactor Repository host_type conditionals

### LINKED ISSUE #1173: Refactor Repository conditionals
Currently we have a lot of conditionals in the `Repository` class that look like this:

```ruby
  def update_from_repository(token = nil)
    case host_type
    when 'GitHub'
      update_from_github(token)
    when 'GitLab'
      update_from_gitlab(token)
    when 'Bitbucket'
      update_from_bitbucket(token)
    end
  end
```

They can be refactored out into a single function that knows which type of repo it is and which function to call.

maclover7:
Fixes #1173

Refactors internal `Repository` methods to `send` based on the `host_type`.

### DIFF
--- a/app/models/repository.rb
+++ b/app/models/repository.rb
@@ -160,7 +160,7 @@ class Repository < ApplicationRecord
   def download_owner
     case host_type
     when 'GitHub'
-      download_github_owner
+      send("download_#{host_type.downcase}_owner")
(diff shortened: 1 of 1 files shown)


andrew:
This is great, thanks so much @maclover7!

maclover7:
👍

PR #1190, opened 2017-02-21
Combine almost identical URL parsers

maclover7:
The Bitbucket and Gitlab URL parsers were almost identical, so it made
sense to just combine them, to avoid extreme code duplication.


andrew:
👍 This is great, I think we can combine this with the https://github.com/librariesio/github_urls project too

maclover7:
@andrew do we want to bring the parser from that gem into librariesio/libraries.io?

andrew:
@maclover7 it was originally extracted from this repo and has a nice set of tests with it, would make sense to have them all in the same place, do you fancy bringing it back in with the test suite?

maclover7:
👍 sure, I'll take a look!

maclover7:
Opened #1192

PR #1195, opened 2017-02-21
Create RepositoryHost classes

maclover7:
If 👍 on this architecture, I can begin to move code over to the different `RepositoryHost::Base` subclasses.

cc @andrew

### DIFF
--- a/app/models/concerns/bitbucket_repository.rb
+++ b/app/models/concerns/bitbucket_repository.rb
@@ -106,10 +106,6 @@ module BitbucketRepository
     nil
   end
 
-  def bitbucket_avatar_url(size = 60)
-    "https://bitbucket.org/#{full_name}/avatar/#{size}"
-  end
-
   def update_from_bitbucket(token = nil)
     begin
       r = Repository.map_from_bitbucket(self.full_name)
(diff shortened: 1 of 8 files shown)


andrew:
👍👍👍

maclover7:
@andrew if possible, can we start off with just Github, and see how that goes? I can continue porting over the different hosts after Github goes live for a short while.

andrew:
@maclover7 yeah we can do, would be good to get some test coverage on these changes, I also was imagining that we'd make a new instance of the RepositoryHost class, passing the Repository object and then using instance methods where it makes sense

maclover7:
@andrew updated to pass around the `Repository` object; what do you think about migrating over methods at a time, instead of "hosts" at a time? Also, I can try and add test coverage for the methods in a separate PR, to keep this one smaller.

andrew:
Methods rather than hosts sounds good to me, and tests can come in a later PR 👍

maclover7:
@andrew updated the PR to move all main business logic methods to the new classes. I'm happy to do a little more cleanup in a followup PR

andrew:
That's looking great!

maclover7:
👍

PR #1216, opened 2017-02-28
Add gascrolldown fork

BenJam:
In order to track elements of 

* homepage
* project page

Using fork rather than jquery extension but can amend if needed. 

? Doesn’t allow nesting of tags in _partial templates. Could I declare ga_scroll_ids as a variable within application_helper and append to that? Showing my lack of golden path knowledge here.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -34,7 +34,10 @@ document.addEventListener('turbolinks:load', function(){
     userTiming: false,
     percentage: false,
     pixelDepth: false,
-    nonInteraction: false
+    nonInteraction: false, 
+    eventHandler: function(data) {
+      console.log(data);
+    }
   });
 
 })
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -28,6 +28,18 @@
 document.addEventListener('turbolinks:load', function(){
   $('.tip').tooltip({placement: 'bottom'})
   stickFooter()
+  
+  gascrolldepth.init({         
+    elements: $('body').data('ga_scroll_ids').split(','),
+    userTiming: false,
+    percentage: false,
+    pixelDepth: false,
+    nonInteraction: false, 
+    eventHandler: function(data) {
+      console.log(data);
+    }
+  });
+
 })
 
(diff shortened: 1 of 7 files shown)


app/assets/javascripts/application.js
```diff
@@ -21,12 +21,25 @@
...
   stickFooter()
+  
+  gascrolldepth.init({         
+    elements: $('body').data('ga_scroll_ids').split(','),
+    userTiming: false,
+    percentage: false,
+    pixelDepth: false,
+    nonInteraction: false, 
+    eventHandler: function(data) {
+      console.log(data);
```
andrew: You can remove the console.log for production

andrew:
@BenJam I just pushed a tweak to make using the helper in partials work as expected 👍

BenJam:
https://github.com/googleanalytics/autotrack looks like it might be a little more consistent. Particularly https://github.com/googleanalytics/autotrack/blob/master/docs/plugins/event-tracker.md

andrew:
@BenJam that looks much nicer, and maintained by Google too

app/helpers/application_helper.rb
```diff
@@ -66,7 +66,7 @@ def title(page_title)
   end
 
   def ga_scroll_ids(ids)
-    content_for(:ga_scroll_ids) { ids.join(',') }
+    content_for(:ga_scroll_ids) { ids.join(',')+',' }
```
BenJam: I'm still unsure about the behaviour and scope here, possibly because of the naming. `ids.join()` will take`ids` param and turn it from an array obj into a string, and provide it back out on `content_for` in another template. Is it storing ids implicitly in the method scope? how does it not get overridden in a second call with an embedded partial template?

andrew: `content_for` concatenates everything sent to it into a single string, as mentioned in http://apidock.com/rails/v4.2.1/ActionView/Helpers/CaptureHelper/content_for, so it's not acting like a regular variable otherwise we should just send it many ids arrays and join them all together in application.html.erb

BenJam: fairs.

PR #1241, opened 2017-03-02
Add autotracker analytics

BenJam:
Adds autotracker.js to libraries.io

BUG: does not concatenate ‘footer’ in layouts/_footer. Fix appreciated! 

TODO: 

- [ ] Add definitions, custom metrics and calculated metrics to GA.


app/views/layouts/_footer.html.erb
```diff
@@ -1,3 +1,4 @@
+<% ga_tracked_els ['footer'] %>
 <footer>
```
BenJam: it *should* be able to track the el rather than the id. Can do though.

BenJam:
Once this is okay I'll replicate my local analytics setup on Libraries.io's

andrew:
Example:

    <div data-ga-track="dependencies"> 

Which can be selected with jquery using:

    $('[data-ga-track]')

BenJam:
Followed approach from #1216 ¯\_(ツ)_/¯

PR #1258, opened 2017-03-08
Move development setup documentation from Readme to /docs

andrew:
Now that we have a `/docs` folder, I'm moving all the technical details of setting up the project locally into a separate file and adding links to all the docs in the readme, which now looks something like this:

screen shot 2017-03-08 at 5 23 13 pm


andrew:
I figured that this repo is going to get a lot more attention that most, and also has the longest list of setup instructions, felt like we can do a much better job of explaining what the project is in this readme and then giving readers pointers to where to continue reading.

Issue #1174, opened 2017-02-16
REST API for enqueuing jobs


andrew:
Currently both `github-dispatch` and `watcher` write directly to the Libraries.io redis to enqueue sidekiq jobs, this has caused problems in the past where the job names have changed in Libraries but not in `github-dispatch` or `watcher`, also not great for security having the redis exposed to multiple services.

Instead Libraries should provide an authenticated REST endpoint for both `github-dispatch` and `watcher` to post to with details of the jobs, which then the rails app can enqueue in sidekiq under the correct names.

andrew:
Still need to deploy both updates and start piping github-dispatch to the new API.

PR #1278, opened 2017-03-13
Introduce GithubHookHandler

maclover7:
An abstract way of dealing with different kinds of GitHub webhooks. Will
be useful when solving #1174.

cc @andrew


andrew:
Thanks!

PR #1281, opened 2017-03-14
Add more events to GithubHookHandler

maclover7:
These should be all of the events that are consumed by [github-dispatch](https://github.com/librariesio/github-dispatch/blob/master/dispatch.rb).

cc @andrew


andrew:
Excellent progress 👍

maclover7:
ref https://github.com/librariesio/libraries.io/issues/1174

PR #1307, opened 2017-03-20
Add `/hooks/package` endpoint

maclover7:
To be used by "watcher". Called when new package versions are released,
and need to be processed.

cc @andrew

### DIFF
--- a/app/controllers/hooks_controller.rb
+++ b/app/controllers/hooks_controller.rb
@@ -13,6 +13,12 @@ class HooksController < ApplicationController
     render json: nil, status: :ok
   end
 
+  def package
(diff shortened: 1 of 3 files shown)


app/controllers/hooks_controller.rb
```diff
@@ -13,6 +13,18 @@ def github
     render json: nil, status: :ok
   end
 
+  def package
+    if params[:payload]
+      payload = JSON.parse(params[:payload])
```
andrew: do we need to have this here, can't we let rails handle the parsing of json?

maclover7: Had this for the `github` endpoint, will remove since we always control the input for the `package` endpoint

andrew:
🙏

PR #1322, opened 2017-03-28
Move JSON generation for API into serializers

andrew:
The current API implementation uses to_json and makes for some fairly horrible looking API controller code, switching to use `active_model_serializers` should make it easier to test and refactor and also useful for generating CSV dumps in the future.


andrew:
Putting this one on hold until the tests for the format of the API responses are improved as I'm not confident that this won't break the API in some subtle places

PR #1332, opened 2017-04-04
Add rake task for exporting dependency data to csv

andrew:
First pass at exporting data for research purposes, trying to match the data format that @AlexandreDecan is already using.


andrew:
Merging this early so I can get it on a production machine to try exporting some real data.

PR #1385, opened 2017-04-28
Add (broken) cii experiment

BenJam:
Todo

* fix pg query syntax
* double check render
* add test

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -149,7 +149,7 @@ class ProjectsController < ApplicationController
     orginal_scope = Project.unsung_heroes
     scope = current_platform.present? ? orginal_scope.platform(current_platform) : orginal_scope
     @projects = scope.order('projects.dependent_repos_count DESC').paginate(page: page_number)
-    @platforms = orginal_scope.group('projects.platform').count.reject{|k,_v|÷ k.blank? }.sort_by{|_k,v| v }.reverse.first(20)
+    @platforms = orginal_scope.group('projects.platform').count.reject{|k,_v| k.blank? }.sort_by{|_k,v| v }.reverse.first(20)
   end
 
   private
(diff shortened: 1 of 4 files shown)


BenJam:
@andrew would be good to understand what's happening with the db qry here. Tripping up on `projets_controller#digital_infrastructure` ln 144. But I'm certain this is due to `project.rb` ln 80 but I took the query for the rake task.

app/models/project.rb
```diff
@@ -77,6 +77,11 @@ class Project < ApplicationRecord
                              .where('repositories.stargazers_count < 100')
                              .where('projects.dependent_repos_count > 1000') }
 
+  scope :digital_infrastructure, -> { not_removed
+                             .with_dependent_repos
+                             .with_repo
+                             .where('projects.dependent_repos_count DESC NULLS LAST').limit(1000)}
```
andrew: This should be .order, rather than .where at the start of this line

app/controllers/projects_controller.rb
```diff
@@ -138,11 +138,18 @@ def sync
...
+    @platforms = orginal_scope.group('projects.platform').count.reject{|k,_v| k.blank? }.sort_by{|_k,v| v }.reverse.first(20)
+  end
+
   def unseen_infrastructure
     orginal_scope = Project.unsung_heroes
     scope = current_platform.present? ? orginal_scope.platform(current_platform) : orginal_scope
     @projects = scope.order('projects.dependent_repos_count DESC').paginate(page: page_number)
-    @platforms = orginal_scope.group('projects.platform').count.reject{|k,_v| k.blank? }.sort_by{|_k,v| v }.reverse.first(20)
+    @platforms = orginal_scope.group('projects.platform').count.reject{|k,_v|÷ k.blank? }.sort_by{|_k,v| v }.reverse.first(20)
   end
```
andrew: Weird percentage symbol on this line that needs to be removed

app/models/project.rb
BenJam: 👍

app/controllers/projects_controller.rb
BenJam: 🥇

BenJam:
Tak!

PR #1419, opened 2017-05-09
Feature/move slack into lib

samjacobclift:
**headlines**

Move [lib2slack](https://github.com/librariesio/lib2slack) into the libs repo

1) Add `slack_api_token` and `slack_channel` to user model
2) Allow users to fill in these fields from the account page
3) Copy in the [lib2slack](https://github.com/librariesio/lib2slack) slack notification method

**NOTES**
Very much a work in progress! feedback needed! This is more of an initial implementation/feeling out how this functionality would work

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -85,6 +85,8 @@ gem 'active_model_serializers'
 gem 'pghero'
 gem 'pg_query'
 gem 'schema_plus_pg_indexes'
+gem 'slack-ruby-client'
+gem 'semantic_range'
 
 group :development do
   gem 'spring'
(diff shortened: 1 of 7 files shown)


andrew:
@samjacobclift awesome start on this, I think we'll need to move it more towards setting the slack channel and token at the repo level rather than user, as different repos might need notifications to go to different channels. 

Also probably needs to be it's own model, something like `Integration`, so we can make a standardized interface for adding new integrations like GitHub Issues, Trello, Gitter etc

samjacobclift:
sounds like a plan I'll work on the models

samjacobclift:
Closing this for now until I can refactor it a bit :)

Issue #1366, opened 2017-04-18
UX Review: Searcher


BenJam:
## Help Needed!
If you use Libraries.io to do any of the following tasks we would *love* to hear from you. Please get in touch with us in the comments, by [email](mailto:hello@libraries.io) or in our [open chat channel](https://slack.libraries.io).

---

In the following we look at a specific audience, their goals or common tasks, map out entry points, user flows and potential drop of points. These flows are informed by analytics data and walk throughs with users. Where drop offs occur we highlight what the user needs and/or expects to see and how well these expectations are met. Finally we outline modifications to be made at each point of the user flow. 

## Audience/Persona
This ticket tackles common uses from the perspective of the [*searcher*](https://docs.libraries.io/personas.html). 

## User need
Ada is a developer who has been assigned a ticket to improve a slow-performing ruby application. She's heard that `pg-hero` is really useful for highlighting slow database queries. She would like to check it out and decide whether to use it. 

She's interested in whether the project is well used, documented and supported, who the maintainer(s) are, who has contributed, how many dependencies the project has, whether using the project would result in further restrictions on her own application, how responsive maintainers are to issues and improvements, whether there are any known vulnerabilities and finally what alternatives are available. 

## User Flows
* Google search > `/project`
* Chrome autocomplete > `/search` > `/project`(s)
* `/` > search input > project(s)
* `/project` > search input > `/search` > `/project`(s) 
* `/search` > `/search` > *

## Behaviour and traffic
* 98% of users flow straight into a `/project page  (e.g. `/github/jondot/awesome-react-native`) from Google. 
* For the ~2% using the search 16% search from `/` while 16% search from `/search?q=` presumably performing a second search or using autocomplete in chrome.
* Once a search is performed exit rate is 26%, stayed by a huge 60% on (entrance)
* 31% of searchers then add refinements
 with an average page depth of ~6 thereafter, avg. duration ~6 minutes indicating that users 

### Search
- [ ] add SourceRank score, to results to make it easier to compare packages (left in favour of #78)
- [ ] consider breaking SourceRank score into sections (code, documentation, community, dependencies) (left in favour of #78)
- [x] make refinements clearer
- [x] use ordinal dates for 'updated by'
- [x] make it more obvious that the number is the current version

### Project/Repository
- [x] Make a clearer distinction between metadata (licence, version, etc) and content (readme)
- [x] Only show the topmost section of the readme
[...]

PR #1459, opened 2017-05-25
Simplify project pages

BenJam:
### General
Brings in some simplifications and reprioritises content following #1366. 

### Specfics
* separation of ‘content’ (left) from ‘data’ (right) where sensible. 
* Makes clearer distinction of third party content i.e. readme.md’s
* makes more of the source rank and brings this above other data
* replaces a lot of tabular content with definition lists which are arguable more semantically valid (but happy to concede they can be annoying)
* adds ‘release’ selector for mobile devices.
* moves a lot of extraneous content for mobile devices (i.e. dates, version number etc)
* remove button classes from things that are not tasks
* brings ‘something wrong’ and export content up the page in sidebar 

### TODO (later)
* replace links names with icons
* work out how to get the source rank under the total/version without having it op right down the page on md devices
* cleaner interface for ‘releases’ 
* show only last major, minor and patch release in releases

Note: this will not be a simple merge :/ 

screencapture-localhost-3000-rubygems-subroutine-1495730942299

screencapture-localhost-3000-rubygems-subroutine-1495731004205


BenJam:
might be nicer to have actions on a row above the title/sourcerank actually...

screen shot 2017-05-25 at 17 55 27

maybe not.

BenJam:
@andrew if you want to review this one and propose any modifications I'll roll it into repo's too. I'll get on with some homepage stuff for now.

app/views/projects/_statistics.html.erb
```diff
@@ -1,115 +1,93 @@
...
-        </td>
-      </tr>
-    <% end %>
+      </dt>
+      <dd class='col-sm-4'>
+        <%= link_to number_to_human(@project.contributions.count), @project.repository.contributors_url %>
+      </dd>
+      <% if @project.repository.size.present? %>
+        <dt class='col-sm-8'>
+          Size:
```
andrew: This size label could be a little misleading without mentioning it's the size of the repository, for long running projects the repo will get bigger even if the size of the published package doesn't, a big deal for css/js frontend libraries.

PR #1481, opened 2017-06-05
Refine homepage

### LINKED ISSUE #1246: Homepage Update
When I read about [Rosehub](https://news.ycombinator.com/item?id=13769727) (Google's auto-PR bot for patching vulnerabilities) on Hacker News and saw how well received it was, I thought, "but libraries.io is all that and more". Yet a newcomer to the homepage would know that neither from a glance, nor a longer read.

So updating a homepage is a big thing, it's the primary symbolic totem expressing and conserving the vision and essence of a project. Its success is therefore dictated by a preexisting expression of those values -- the homepage does not create them. So discussion about this issue is more than just logistics, it's an exploration of what's most meaningful about libraries.io to you.

Personally I would argue there are some easy and obvious changes that could be made, all based around the fact that libraries.io is not just an, "Open Source Discovery Service". For example:
- "be notified of new releases" -- should be an independent statement, its own sentence.
- The API, license compatibility checking, "Help Wanted/First Pull Request" and Bus Factor features should all be above the fold.
- Make [`lib2issues`](https://github.com/librariesio/lib2issues) a toggleable service for every repo and advertise this above the fold. (There's a bigger, but separate, conversation here about doing what Rosehub does but on a much more comprehensive scale, personally I think auto-PRs should be opt-in, not opt-out.)

But there are some bigger questions that I might venture to propose, namely regarding its prevailing identity as a discovery service. To give a concrete suggestion, what if libraries.io strapline was, "Open Source's Immune System" (I don't think this is actually a good idea, it's just something to work with). From it we get more of an immediate view into the grander and altruistic vision of maintaining and improving the overall health of open source software. However it leaves out the discovery features, so perhaps a strapline in between would work.

Here's a reminder of the vision from the About page:

> Our goal is to raise the quality of all software, by raising the quality and frequency of contributions to free and open source software; the services, frameworks, plugins and tools we collectively refer to as libraries.

So anyway, the point is, yes there are some quick wins to improving the homepage, but there is also the fact that libraries.io could possibly benefit from shifting the centre of gravity of its brand identity. Let me suggest some actionable tasks:

- [ ] Identify consensus on strapline/core vision, does it need changing?
- [ ] If so, to what?
- [ ] Identify the minimal core features that should be above the fold.

BenJam:
* remove ‘search by’ sections
* add persona-focussed info/CTAs
* reintroduce /explore

I have not included btn-groups in the two ‘success’ CTAs as they cannot reliably be set to width:100%. Could remove centre-block from the buttons however and re-introduce this as it does save a click (and the login page is very minimal). Open to comments. 

screen shot 2017-06-05 at 10 31 48

Fixes #1246

### DIFF
--- a/app/views/projects/index.html.erb
+++ b/app/views/projects/index.html.erb
@@ -1,6 +1,7 @@
 <div class='row'>
-  <div class="col-md-12 leaderboard">
-    <h1>Libraries.io <small>The Open Source Discovery Service. <%= link_to 'Find out more &raquo;'.html_safe, about_path, class: 'visible-xs' %></small></h1>
+  <div class="col-md-12 headline">
+    <h1>Libraries.io</h1>
+    <%= link_to 'Find out more &raquo;'.html_safe, about_path, class: 'visible-xs' %>
 
     <%= form_tag search_path, method: :get, class: 'search-form', enforce_utf8: false do |f| %>
       <div class="input-group input-group-lg">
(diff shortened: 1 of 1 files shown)


andrew:
🚢

BenJam:
toot toot!

PR #1482, opened 2017-06-05
Add search to nabber for logged in users (who will be looking at a list of repos)

BenJam:
Not sure of the syntax here but it appears valid. In irb.


andrew:
LGTM 👍

PR #1497, opened 2017-06-12
Modifications to /open-data page

BenJam:
These are more ideas than a firm stake in the ground on how I think the page should be. 

To the description top left tI have added 

* first para: very brief and functional
* second para: something more context setting and narrative
* third para: something of an overview about the data contained within the release

I do not think all these need to be there. Maybe something for jphill to look at. 

screen shot 2017-06-12 at 17 50 33


BenJam:
I think it needs a little more v space. Very closed up around titles atm.

andrew:
Can we bump the download and documentation partials back out into a col-md-12 in another row below?

BenJam:
I was thinking about that. Could the documentation live on another page like `/data/documentation`. Or perhaps we have the documentation describing the relationships between tables and how to use the BQ database and then a `/data/dictionary` page that describes each field in each dataset?

BenJam:
I was also thinking about having a supported package manager list in the sidebar a la `/explore` in order to encourage some more contribution...

andrew:
Ok sounds good 👍

BenJam:
modified the view a fair whack. Actually taking up slightly more room on the page but removing a lot of the detail on load. Happy with a combined page like this now. But not happy with the way of toggling the fields:

screen shot 2017-06-13 at 16 50 01

BenJam:
Added collapse arrows: 

screen shot 2017-06-14 at 10 34 08

andrew:
👍

andrew:
I can fix the merge conflicts if you like?

BenJam:
if you know which bits you updated, might be easier. Manual grok for me else.

PR #1519, opened 2017-06-19
Fixup Docker configuration for local dev

brynary:
:wave: This PR attempts to fix and clean up the Docker configuration needed for local dev, as it seemed to be broken. (Running `docker build` failed due to missing system dependencies). Key changes:

* Ignore some unneeded paths when building Docker image
* Parallelize gem installation during `docker build` for speed
* Add missing system deps (libicu-dev and cmake)
* Reduce verbosity of Sidekiq logging as it gets quite overwhelming quickly (let me know if this is not desired)
* Revamp docker-compose with the following improvements:
  * Durable volume stores for the databases
  * Pin versions of database images to avoid inadvertent drift
  * Add Sidekiq container
 
Possible future improvements include adding Mailcatcher (easy), and supporting running tests via Docker (there are multiple implementation paths for this, so I avoided it for now).

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,9 @@
+.env
+.git
+*.csv
+*.dump
+coverage
+data
+log
+node_modules
+tmp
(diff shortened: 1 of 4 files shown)


andrew:
Thanks @brynary!

PR #1534, opened 2017-06-26
Update homepage to prioritise /data release

BenJam:
Just a quick update.  Played around with a few other options including:

- having two buttons
- having a fifth panel
- creating a combined data page

Didn’t like em. Rethink later.


andrew:
🚢

PR #1566, opened 2017-07-10
Off white header spike

andrew:
Opening this PR up for as a discussion than a fix, I was playing around with a lighter header last week, came up with this:

screen shot 2017-07-04 at 12 46 38 pm

The black logo SVG really needs recreating to make it sharper, alternatively we could inline the current white svg into the page which would allow us to style it with css and do `fill: #000`

Love it, hate it, any other thoughts?


M-Zuber:
Overall I like it very much, but the second to last color (not sure what pm it represents) on the line underneath blends it a bit

andrew:
@M-Zuber true, that was just a random selection of colours that I chose a couple years ago, we should probably make the bar a little bit more meaningful, perhaps colours and percentages of the most popular languages on the site

e-beach:
I think the light header looks a little better, mostly because the search field and login button look more appropriate with the light header. If you look at GitHub, the site has a dark navbar, but because the text-fields are also dark, things still look natural.

andrew:
I'm going to close this as I suspect there will be a more involved, proper redesign in the next couple months.

Issue #1603, opened 2017-07-17
Dependency activity on repositories


andrew:
As part of https://github.com/librariesio/supporters/issues/21 and feeding into https://github.com/librariesio/metrics, we want measure usage of libraries over time. 

Currently we have a total for the number of open source repositories that depend on a given library but we don't have any information on the trends of usage.

We want to be able to see: 

- What time/date did the library get added/modified/removed as a dependency to a repo 
- Who added/modified/removed the dependency
- Why was the dependency added/modified/removed 

Git commit data has the answer to those three questions:

- The actual time the dependency was added, not just when we indexed it, going back through the history of the repository
- Both the author and the committer email addresses for each commit show who made the change
- The commit message gives a qualitative reason as to why a dependency change was made

With this data we'll be able to graph change in usage in open source over time for a library and each individual version for all package managers that have a concept of a dependency manifest/lockfile.

Pieces of the puzzle:

- [ ] Code to walk through the history of a repository: https://github.com/librariesio/repo_miner
- [ ] Code to parse dependency manifests and lockfiles: https://github.com/librariesio/bibliothecary
- [ ] Background job for mining dependency activity from a repository
- [ ] Database table for storing dependency change events
- [ ] Time series database for indexing and graphing events over time

PR #1609, opened 2017-07-19
Dependency activity collection

andrew:
First attempt at collecting and storing dependency activity in the database for https://github.com/librariesio/libraries.io/issues/1603 and ultimately https://github.com/librariesio/supporters/issues/21

Adds a method to clone a repository, analyse its history (using [repo_miner](https://github.com/librariesio/repo_miner)) and then store each action as a `DependencyActivity` record, including finding the project id by it's name in a process (if possible)

Includes a few indexes on the table but not 100% sure what we're going to want to do with the data yet so keeping them to a minimum for now.

~~Partially blocked by https://github.com/librariesio/repo_miner/issues/2 right now~~


andrew:
@BenJam gunna hold off on merging this until the open data release prep is done, mostly because this will put an unknown amount of load on the queue 🙈

BenJam:
👍

PR #1615, opened 2017-07-21
Update links to open data release for v1.0.1

andrew:
Updated links to new zenodo release which should fix the zip errors people have been reporting: https://zenodo.org/record/833207

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -78,7 +78,7 @@ h3,h4{
   }
 }
 
-dt {
+dt, dd {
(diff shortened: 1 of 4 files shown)


BenJam:
maybe tighten up the leftmost div because the date is cutting in two. Might also be better to retain direct links to previous version(s) too.

andrew:
@BenJam updated, how's this:

screen shot 2017-07-21 at 4 07 03 pm

BenJam:
👍

andrew:
Deploying now

PR #1633, opened 2017-07-31
Add experiments page

### LINKED ISSUE #1140: Group and list the experimental sections of the site together
Things like /bus-factor and /unseen-infrastructure don't fit with the general use cases of the site, we should have a top-level section called experiments or labs that lists them all and tucks them away from the main flow

BenJam:
* Adds Experiments page:
screen shot 2017-07-31 at 17 12 35

* updates footer:
screen shot 2017-07-31 at 17 12 40

TODO:

- [ ] Fix fa icons
- [ ] Add SVG test-tube libraries logo
- [ ] Check copy

Closes #1140

### DIFF
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -15,6 +15,11 @@ class PagesController < ApplicationController
 
   end
 
+  def experiments
+    @platforms = Project.popular_platforms(:facet_limit => 40).first(28)
+
+  end
+
   def data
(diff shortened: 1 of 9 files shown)


app/controllers/pages_controller.rb
```diff
@@ -15,6 +15,11 @@ def compatibility
 
   end
 
+  def experiments
+    @platforms = Project.popular_platforms(:facet_limit => 40).first(28)
+
```
andrew: No need for this new line here

Issue #1026, opened 2016-12-29
Support Puppet Forge


petems:
Add libraries.io support for the Puppet Forge

URL: https://forge.puppet.com/modules?utf-8=%E2%9C%93

API Documentation: https://forgeapi.puppetlabs.com/#!/module/getModule

petems:
@andrew Is there a rough pointer on how to add a new package type? Like what rake tasks I need to create the new objects and stuff? Wouldn't mind trying to implement this myself if I get some slack time :

andrew:
@petems most of the code to add a new package manager is adding a new class to this folder: https://github.com/librariesio/libraries.io/tree/master/app/models/package_manager

Rubygems is a good example of some of the functions that need to be implemented: https://github.com/librariesio/libraries.io/blob/master/app/models/package_manager/rubygems.rb

andrew:
This has shipped now 🎉

petems:
Thanks to @andrew and @garethr for getting this done, super cool! 👍

PR #1691, opened 2017-08-31
Initial work on supporting Puppet Forge packages

garethr:
This is pretty much just the minimal methods implemented, but lots of
the others are possible based on the capabilities of the API.

This aims to address #1026 

I'll follow up with additional commits to flesh out the other methods when I get a moment but this imports Puppet package details locally.

### DIFF
--- a/app/models/package_manager/puppet.rb
+++ b/app/models/package_manager/puppet.rb
@@ -1,7 +1,7 @@
 module PackageManager
   class Puppet < Base
-    HAS_VERSIONS = false
-    HAS_DEPENDENCIES = false
+    HAS_VERSIONS = true
+    HAS_DEPENDENCIES = true
     BIBLIOTHECARY_SUPPORT = false
     BIBLIOTHECARY_PLANNED = true
     URL = 'https://forge.puppet.com'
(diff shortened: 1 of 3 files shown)


app/models/package_manager/puppet.rb
```diff
@@ -0,0 +1,41 @@
...
+    end
+
+    def self.project(name)
+      get_json("https://forgeapi.puppetlabs.com/v3/modules/#{name}")
+    end
+
+    def self.mapping(project)
+      {
+        name: project['slug'],
+        homepage: "https://forge.puppet.com/#{project['slug']}",
```
andrew: This should be the the `project_page` attribute, we can generate the link to the forge page from the name by adding a `package_link` method like: https://github.com/librariesio/libraries.io/blob/e38e012b4b9dd49e0b20ff242279d37edd5940ed/app/models/package_manager/rubygems.rb#L10-L12

andrew:
Sweet, thanks for kicking this off @garethr! 

Looks like adding `versions`, `dependencies`, `recent_names`, `package_link` and `download_url` methods will be pretty easy, I can do this for you if you'd like?

The only slightly odd thing is the difference between name and slug, we only have one field for storing and using the name, which seems to work best as slug, is it ok to display the name on each project page as the dashed version rather than slashed?

Looks like there's a puppetfile parser gem [here](https://rubygems.org/gems/puppetfile-parser) that we can use for Bibliothecary to avoid having to execute the file.

I also need to update the section about watcher in https://github.com/librariesio/libraries.io/blob/master/docs/add-a-package-manager.md as it's now moved to https://github.com/librariesio/dispatch

garethr:
@andrew thanks for the quick sanity check. I have most of the rest of the interface implemented locally, but need to do something else this afternoon. I'll finish up all being well tomorrow/early next week and update this PR. I wanted to get something up for visibility.

I was going to get working without bibliothecary and then hopefully move on to that too if that works.

I also want to tidy up the code more generally and write a few tests. This was basically me seeing how quickly I could get something working (very quickly as it turned out).

andrew:
Great, looking forward to rolling out Puppet support 🎉

BenJam:
Huzzah! 🙌

garethr:
With the code in this PR I have this working locally I think.

libraries io

I'd love for someone else to try it out and confirm. I'll add a set of unit tests for the package manager class when I get a moment too.

app/models/package_manager/puppet.rb
```diff
@@ -0,0 +1,75 @@
...
+      project['releases'].map do |release|
+        {
+          number: release['version'],
+          published_at: release['created_at']
+        }
+      end
+    end
+
+    def self.dependencies(name, version, project)
+      metadata = project['current_release']['metadata']
```
andrew: This should be the dependencies for the given version, looks like it's only setting dependencies for the current version on every version here

andrew:
@garethr I'm just running an import locally now 👍

app/models/package_manager/puppet.rb
garethr: Unfortunately the API doesn't expose this information as part of getModule: https://forgeapi.puppetlabs.com/#!/module/getModule

I've opened an issue against the Forge API track whether we can expose that information.
https://tickets.puppetlabs.com/browse/FORGE-394

andrew: Looks like we can use the release endpoint instead for that: https://forgeapi.puppetlabs.com/v3/releases/puppetlabs-ntp-6.1.0

garethr: Ah ha. Excellent. I'd missed that. Will look to update to use that endpoint.

garethr: I've pushed a new commit which uses this endpoint. dependencies should now be per version.

garethr:
@andrew I've added a few basic tests based on those from the rubygems model, and updated to get dependency information on a per-version basis. Anything else just let me know and I'll try and take a look.

andrew:
@garethr looks great, I'm going to deploy this shortly 👍

PR #1726, opened 2017-09-14
Change bower registry to new one

sheerun:
We changed Bower's registry so it's behind a CDN, please change it as well :)


andrew:
Thanks!

PR #1827, opened 2017-11-15
Update button copy for subscribed projects

### LINKED ISSUE #1752: UX Suggestion "Subscribe to releases" button
The subscription button currently only changes by one letter, when subscribing to a package: 

image

Subscribe**d** to releases: 
image

This change is very subtle and I was not sure right away, if the subscription succeed.
One could also change the color, but I am not sure in which direction. Red seems to be to appropriate here.

Better: 
image

djpowers:
- [x] Is there a corresponding ticket for your pull request?
- [x] Have you successfully run the project with your changes locally?

- Fixes #1752

This change adds a checkmark to the "subscribe" button and updates the copy to read "Already subscribed" when a user is viewing a project they are subscribed to. As mentioned in the issue, this more clearly differentiates between the two states.

Please note that this removes the conditional that would show the word "stable." Also, the Bootstrap grid widths were adjusted to accommodate the larger button size.

### DIFF
--- a/app/views/projects/_actions.html.erb
+++ b/app/views/projects/_actions.html.erb
@@ -38,7 +38,7 @@
 
   <% if logged_in? && subscription = current_user.subscribed_to?(@project) %>
     <div class="btn-group pull-right-lg">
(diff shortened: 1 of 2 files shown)


andrew:
Thanks!

PR #1854, opened 2017-12-03
Be more permissive in package names in the API

katzj:
Some package names (notably java ones) have colons in their name.  Stick them in the set of allowed characters for the API similar to allowing @ in 0e7c59ecd3534badd56e14a55afca673baeeb7ff

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -49,13 +49,17 @@ Rails.application.routes.draw do
       get '/:host_type/:login', to: 'repository_users#show'
     end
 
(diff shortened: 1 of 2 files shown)


katzj:
I wonder if it's worth changing the constraint to be anything but / ?

andrew:
@katzj I think dot is the other one that will screw with the rails router.  Foobar.json for example

katzj:
Updated to be super permissive and updated the test to use a contrived and crazy package name with a lot of special characters

PR #1853, opened 2017-12-02
Add Racket

aelaa:
Feature asked by #1820 
Greeting from [24pullrequests.com](http://24pullrequests.com)

PR #1855, opened 2017-12-03
Add support for psc-package package manager

### LINKED ISSUE #1755: Add support for purescript package manager
Pretty early in development but adding it here so we can keep track: https://github.com/purescript/psc-package

Default repository of packages: https://github.com/purescript/package-sets

djpowers:
Resolves #1755

Cheers!

### DIFF
--- a/app/models/package_manager/psc_package.rb
+++ b/app/models/package_manager/psc_package.rb
@@ -1,0 +1,35 @@
+module PackageManager
+  class PscPackage < Base
+    HAS_VERSIONS = false
+    HAS_DEPENDENCIES = false
+    BIBLIOTHECARY_PLANNED = true
+    URL = 'https://github.com/purescript/psc-package'
+    COLOR = '#1D222D'
+
+    def self.project_names
+      get('https://raw.githubusercontent.com/purescript/package-sets/master/packages.json').keys
+    end
+
+    def self.project(name)
+      project = get('https://raw.githubusercontent.com/purescript/package-sets/master/packages.json')[name]
+      project['name'] = name
+
+      project
+    end
+
+    def self.mapping(project)
+      {
+        :name => project['name'],
+        :repository_url => project['repo']
+      }
+    end
+
+    def self.install_instructions(project, version = nil)
+      "psc-package install #{project.name}"
+    end
+
+    def self.formatted_name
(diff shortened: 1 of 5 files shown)


app/models/package_manager/psc_package.rb
```diff
@@ -0,0 +1,35 @@
...
+    URL = 'https://github.com/purescript/psc-package'
+    COLOR = '#1D222D'
+
+    def self.project_names
+      get('https://raw.githubusercontent.com/purescript/package-sets/master/packages.json').keys
+    end
+
+    def self.project(name)
+      project = get('https://raw.githubusercontent.com/purescript/package-sets/master/packages.json')[name]
+      project['name'] = name
```
djpowers: This is done so that we'll have access to the `name` field in `#mapping` and other methods, as the JSON returned otherwise does not include it.

andrew: 👍

andrew:
Brilliant, thanks so much @djpowers!

andrew:
I wonder if `PureScript` would be a better name for the type of packages, we have a few package managers that match the names of the language/runtime, and this pr that adds Racket support does it too: https://github.com/librariesio/libraries.io/pull/1853/files

Also, @djpowers, do you know if there's any relationship between package-sets and https://pursuit.purescript.org? Is it using the same dataset?

djpowers:
Thanks for taking a look!

I agree with the name change. I'd initially called it that, but wasn't exactly sure what the naming convention was. Just pushed up a fixup.

And I came across Pursuit while looking at package-sets. It wasn't immediately clear to me what the relationship between the two resources is, but there did seem to be a lot of overlap in the dataset at a quick glance.

app/models/package_manager/pure_script.rb
```diff
@@ -0,0 +1,35 @@
...
+        :repository_url => project['repo']
+      }
+    end
+
+    def self.install_instructions(project, version = nil)
+      "psc-package install #{project.name}"
+    end
+
+    def self.formatted_name
+      'PureScript'
```
andrew: We don't need this method if the name of the class is doesn't need tweaking

andrew:
🚢  https://libraries.io/purescript

Issue #1928, opened 2018-01-19
Upgrade license-compatibility to version 3.0.0


andrew:
Libraries.io has found that there is a newer version of license-compatibility that this project depends on.

More info: https://libraries.io/rubygems/license-compatibility/3.0.0

PR #1929, opened 2018-01-21
Upgrade license-compatibility to version 3.0.0

sahaynishant342:
Link to corresponding ticket: https://github.com/librariesio/libraries.io/issues/1928
In order to ensure that the project is using the above latest version of gem, we need to specify the same version in the Gemfile. With this, for both existing and new gem installations in the project, we will be installing version 3.0.0. Tested it though bundle install.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -44,7 +44,7 @@ gem 'foreman', require: false
 gem 'puma'
 gem 'rack-timeout'
 gem 'semantic_range'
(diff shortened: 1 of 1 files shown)


andrew:
Thanks for the pull request, the version only needed bumping in `Gemfile.lock` with the command `bundle update license-compatibility`, which I've done on master this morning so going to close this.

PR #1936, opened 2018-01-25
Serialize also dependents_count and dependent_repos_count

### LINKED ISSUE #1935: Expose dependents_count and dependent_repos_count in /api/:platform/:name
Hello,

first, let me thank you for wonderful service/software.

I'd like to suggest an enhancement of the API.
The `GET /api/:platform/:name` endpoint does not return all Project's fields that are stored in database.
For example `dependents_count` and `dependent_repos_count` are missing, even they are exposed via web (as `Dependent projects` and `Dependent repositories`).
Other way how to get the counts would be to count all entries in `/api/:platform/:name/dependents` and `/api/:platform/:name/dependent_repositories`, which as you can imagine would take eternity.

jpopelka:
Fixes #1935


andrew:
Thanks!

PR #1938, opened 2018-01-26
Handle cocoapods license fields sometimes being a hash rather than string

andrew:
Discovered whilst investigating https://github.com/librariesio/spdx/issues/42


andrew:
Deployed this now, resyncing all the incorrect cocoapods projects now

PR #1939, opened 2018-01-28
Update Homebrew to pull from formulae.brew.sh

### LINKED ISSUE #1756: Switch homebrew to pull from formulae.brew.sh
http://brewformulas.org has been timing out a lot and it looks like http://formulae.brew.sh/ is now an officially supported project, looks it has an API and and an rss feed: http://formulae.brew.sh/feed.atom

djpowers:
I didn't see a public API available, but based on the discussion [here](https://github.com/Homebrew/formulae.brew.sh/issues/45), using the output of the `brew info` command lets you retrieve essentially this same information. If this command isn't available on the production environment, hopefully there's a workaround available that would allow us to us it.

Resolves #1756


djpowers:
If the prod. server is running on Linux, [Linuxbrew](http://linuxbrew.sh/) might be a viable option.

andrew:
Thanks for tackling this, I don't think we can realistically shell out to the `brew` command on the servers though.

I spoke to @MikeMcQuaid at the weekend about getting a JSON equivalent api on http://formulae.brew.sh, sounds like that might be a possibility.

djpowers:
Thanks for looking into this. I'd be happy to revise this PR if/when it becomes possible to use a proper JSON API.

andrew:
Somehow managed to merge that into master a little early but it's going to be mostly 🚢 soon

djpowers:
Great, thanks for finishing this up! 😁

MikeMcQuaid:
> I spoke to @MikeMcQuaid at the weekend about getting a JSON equivalent api on http://formulae.brew.sh, sounds like that might be a possibility.

This is a thing now: http://formulae.brew.sh/api/formula.json and e.g. http://formulae.brew.sh/api/formula/a2ps.json. Those are kept up to date now. There will likely be a redirect in future to HTTPS but otherwise it should be a stable API.

andrew:
Updated the code here to use those two new endpoints, thanks @MikeMcQuaid!

MikeMcQuaid:
There's now even docs and some analytics endpoints (if that's useful for anyone): https://formulae.brew.sh/docs/api/

Issue #1916, opened 2018-01-09
SourceRank 2.0


andrew:
# SourceRank 2.0

Below are my thoughts on the next big set of changes to "SourceRank", which is the metric that Libraries.io calculates for each project to produce a number that can be used for sorting in lists and weighting in search results as well as encouraging good practises in open source projects to improve quality and discoverability.

## Goals:

-   Comparable number for packages within a given ecosystem
-   Easily understandable for a human to scan without explanation
-   Clear breakdown available of factors that went into the final number
-   Any negative factors that contribute to the score should be fixable by maintainer
-   Changes should be trackable over time
-   Less popular projects that are good quality and well maintained should still score well
-   Doing work to improve sourcerank for a project should encourage following best practises

## History:

SourceRank was inspired by google pagerank as a better alternative score to GitHub Stars.

The main element of the score is the number of open source software projects that depend upon a package.

If a lot of projects depend upon a package that implies a some other things about that package:

-   it's working software, people tend to remove dependencies that are broken
-   It's documented enough to be used by the people that have committed it

## Problems with 1.0:

Sourcerank 1.0 doesn't have a ceiling to the score, the project with the highest score is mocha with a sourcerank of [32](https://libraries.io/npm/mocha/sourcerank), when a user is shown an arbitrary number it's very difficult to know if that number is good or bad. Ideally the number should either be out of a total, i.e. 7/10, a percentage 82% or a score of some kinda like B+

Some of the elements of sourcerank cannot be fixed as they judge actions from years ago with the same level as recent actions, for example "Follows SemVer?" will punish a project for having an invalid semver number from many years ago, even if the project has followed semver perfectly for the past couple years. Recent behaviour should have more impact that past behaviour.

If a project solves a very specific niche problem or if a project tends to be used within closed source applications a lot more than open source projects (LDAP connectors, payment gateway clients etc) then usage data within open source will be small and sourcerank 1.0 would rank it lower.

Related to small usage factors, a project within a smaller ecosystem will currently get a low score even if it is the most used project within that whole ecosystem when compared to a much larger ecosystem, elm vs javascript for example.

Popularity should be based on the ecosystem which the package exists within rather than within the whole Libraries.io universe.

Quality or issues with dependencies of a project are not taken into account, if adding a package brings with it some bad quality dependencies then that should affect the score, in a similar way, the total number of direct and transitive dependencies should be taken into account.

[...]

PR #2056, opened 2018-03-27
Add new SourceRank calculator class

andrew:
⚠️ Work in progress ⚠️ 

First actual code towards SourceRank 2.0 for https://github.com/librariesio/libraries.io/issues/1916

Starting a separate class rather than a module that mix into `Project` as with SourceRank 1.0 to avoid any potential clashes, also makes it a little easier to test and possibly opens up a way for extracting/moving the SourceRank 1.0 implementation into it later on as well.

### DIFF
--- a/app/models/source_rank_calculator.rb
+++ b/app/models/source_rank_calculator.rb
@@ -1,0 +1,9 @@
+class SourceRankCalculator
+  def initialize(project)
+    @project = project
+  end
+
+  def overall_score
+    0
+  end
+end
--- a/spec/models/source_rank_calculator_spec.rb
+++ b/spec/models/source_rank_calculator_spec.rb
@@ -1,0 +1,12 @@
+require 'rails_helper'
+
+describe SourceRankCalculator do
+  let!(:project) { create(:project) }
+  let(:calculator) { SourceRankCalculator.new(project) }
+
+  describe "#overall_score" do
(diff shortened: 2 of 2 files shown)


andrew:
First pass at an implementation for the calculator is mostly complete, still needs some more specs for the recursive dependency data scores.

andrew:
I've now got this storing the value of the overall score in `sourcerank_2` and put together a rake tasks to generate scores for a test set of projects from rubygems.

One thing I've stopped it doing for now is recursively calculating scores for dependencies as it as the possibility to run wild, but that means to to begin with, some dependencies will have a score of zero because they've not been calculated yet, so the score of the overall project will need to be updated again after the dependency has been calculated.

The rake task has a naive approach to getting round that by first calculating the scores for all the projects with zero runtime dependencies, as they are at the edge of the graph and can't be affected by yet-to-be-calculated dependency scores. 

It then moves on to calculating scores for libraries with 1 dependency, then 2, then 3 etc but that can still run into the original issue as the 1 dependency still may not have a calculated score yet.

Two improvements off the top of my head:

- run the rake task multiple times until it stops updating projects
- keep track of projects that likely will need their scores updating by looking for zero score dependencies and then redoing those after a full run is complete

andrew:
We're moving away from the term `SourceRank`, so I've renamed and updated this pr to use the more generic "project score" language instead.

PR #2158, opened 2018-05-31
repo_search doesn't need facets on ES search

katzj:
Stop getting back faceted aggregations to speed up queries on pages
like /github so that we don't timeout. We're not using the aggregations
and they're expensive for ES to compute


katzj:
I think this should fix the most frequent timeout we're getting in appsignal right now (and thus that users are seeing). Each query in ES with aggregations takes like ~12 seconds. Without it takes < 2.

PR #2218, opened 2018-09-15
Fix project links in API doc page

marcandre:
I noticed also that the API page has inconsistencies, as noted in [#2126].

This PR fixes the inconsistencies related to the links.

No attempt (yet) has been made to insure that the example output actually matches the API's output.


katzj:
Thanks!

Issue #2201, opened 2018-07-21
Package not found suggests raising a bug


bruno-garcia:
I'm not sure this is in fact a bug. The UI does link to the bug tracker:

![image](https://user-images.githubusercontent.com/1633368/43040280-474c1ef0-8d40-11e8-969d-2bf41d2ab274.png)

I'm searching for [sentry-dotnet](https://github.com/getsentry/sentry-dotnet/) and indeed it's not found.
How does it work to get a repository and [nuget packages](https://www.nuget.org/profiles/sentry) indexed?

katzj:
The github repo is there now. nuget isn't showing up as it's not available from the nuget api right now -- fetching `https://api.nuget.org/v3/registration1/sentry/index.json` returns a 404

bruno-garcia:
It's available through `registration3`:
https://api.nuget.org/v3/registration3/sentry/index.json

katzj:
Hmm, will need to look a little closer to understand differences between the two registration endpoints

katzj:
And it now shows up (the package is named sentry -- https://libraries.io/nuget/sentry)

PR #2227, opened 2018-09-25
Switch to the (non-deprecated) registration3 endpoint for nuget

katzj:
nuget stopped updating registration1 a few months ago, we should
use registration3. It looks like everything we read out of the json
works just fine and it was already handling following the pages


katzj:
Fixes #2201

PR #2267, opened 2018-12-12
Mellisio/perf/add matview dependent repos

mellisio:


db/migrate/20181211181245_create_mat_view_for_dependent_repos.rb
```diff
@@ -0,0 +1,15 @@
+class CreateMatViewForDependentRepos < ActiveRecord::Migration[5.1]
+
+  def up
+    # This can take about 45 - 60 minutes
+    execute 'create materialized view project_dependent_repositories as
```
katzj: Not sure why travis isn't like this...

Also maybe do this as `WITH NO DATA` so that it doesn't fill the table during the deploy?

db/schema.rb
```diff
@@ -50,8 +50,8 @@
     t.string   "kind"
     t.boolean  "optional",     :default=>false
     t.string   "requirements"
-    t.datetime "created_at",   :null=>false
-    t.datetime "updated_at",   :null=>false
+    t.datetime "created_at"
```
katzj: Why are these definitions shifting?

lib/tasks/projects.rake
```diff
@@ -113,4 +113,11 @@ namespace :projects do
       end
     end
   end
+
+  desc 'Refresh dependent repos materialized view'
+  task refresh_dependent_repos_view: :environment do
+    # This task can take about 30 minutes,
+    exit if ENV['READ_ONLY'].present?
+    ActiveRecord::Base.connection.execute("refresh materialized view project_dependent_repositories")
```
katzj: this should probably be done with `CONCURRENTLY` so that reads can continue as the refresh happens

db/schema.rb
mellisio: I'm not sure... I just re-did the migration from a fresh master schema.db and it generated the same file...

mellisio: That's my usual procedure as well, however I think I'm running into this https://github.com/SchemaPlus/schema_plus_core/issues/22 locally, and I cannot reset the db.

PR #2275, opened 2018-12-20
Add materialized view with scenic gem so we can have it in schema.rb

mellisio:


db/migrate/20181213214340_create_project_dependent_repositories.rb
```diff
@@ -0,0 +1,10 @@
+class CreateProjectDependentRepositories < ActiveRecord::Migration[5.2]
+  def change
+    create_view :project_dependent_repositories, materialized: true
```
katzj: Can this create the view without data and then let the data get filled in on the first run of the refresh later? That lets us decouple deployment (and having *other* deployments occur) from adding the initial view

db/migrate/20181213214340_create_project_dependent_repositories.rb
```diff
@@ -0,0 +1,10 @@
...
+    add_index :project_dependent_repositories, [:project_id, :rank, :stargazers_count],
```
katzj: Otherwise these indexes are going to have to then also iterate over the table to be created rather than getting filled in as the data refreshed the first time (concurrently)

PR #2277, opened 2019-01-02
Internal API Key

mikeyoung85:
Adds an `is_internal` flag to API keys, defaulted to false. I've included a rake task to set a token to internal and added a method in `ApplicationController` to require an internal key or return a 403.


db/migrate/20181220164827_add_is_internal_to_api_keys.rb
```diff
@@ -0,0 +1,5 @@
+class AddIsInternalToApiKeys < ActiveRecord::Migration[5.1]
+  def change
+    add_column :api_keys, :is_internal, :boolean, default: false
```
katzj: This isn't a safe way to do the migration. Instead need to split it to be

```ruby
add_column :api_keys, :is_internal, :boolean
change_column_default :api_keys, :is_internal, false
```

mikeyoung85: I'm confused why that is. The `change_column_default` will not allow for rollback.

katzj: Adding the column with a default requires a table lock for the entire length of time to change every row while changing the default after does an update on a row-by-row basis.

The rollback is to drop the column

mikeyoung85: Ok I will rewrite it to use `change_column_default` and to support up/down methods. Does it make sense to add a null constraint?

katzj: Probably doesn't hurt

PR #2279, opened 2019-01-03
Avoid calculating the top ten, and dependent repos count on demand

mellisio:


app/models/project.rb
```diff
@@ -271,7 +271,15 @@ def homepage
   def set_dependents_count
     return if destroyed?
     new_dependents_count = dependents.joins(:version).pluck('DISTINCT versions.project_id').count
-    new_dependent_repos_count = dependent_repositories.open_source.count.length
+
+    # Most of the time we truncate this count if it is larger than 1,000 anyways, so only do slow version if less
```
katzj: Is the reason to do this just to avoid stale data? It's probably fine to just accept the eventual consistency and always go to the view

app/models/project.rb
```diff
@@ -415,6 +423,18 @@ def subscribed_repos(user)
     subscriptions.with_repository_subscription.where('repository_subscriptions.user_id = ?', user.id).map(&:repository).uniq
   end
 
+  def dependent_repos_view_query(limit, offset=0)
+    Repository.find_by_sql(["select * from repositories where id in (select repository_id from project_dependent_repositories where project_id = ? order by rank desc nulls last, stargazers_count desc limit ? offset ?);", id, limit, offset])
```
katzj: I find it a lot easier to read SQL when the reserved words are in caps. Also can you split this into multiple lines?

app/models/project.rb
```diff
@@ -415,6 +423,18 @@ def subscribed_repos(user)
     subscriptions.with_repository_subscription.where('repository_subscriptions.user_id = ?', user.id).map(&:repository).uniq
   end
 
+  def dependent_repos_view_query(limit, offset=0)
```
katzj: Do we need limit and offset to be args at this point?

mellisio: For this commit, strictly speaking, we do not. I was playing around with other sections on my local and for other parts will need it. I can take it out and at it as a part of another PR later if that's preferred.

PR #2282, opened 2019-01-08
Add Repository Maintenance Stat Worker

mikeyoung85:
Added a worker to run repository maintenance stats. There are 3 queues added to Sidekiq: low, medium, and high meant to prioritize repositories. The worker has a class method `queue` that takes in a priority as an argument and places the request on the correct priority queue for Sidekiq processing. There is a uniqueness lock applied across the 3 queues to prevent repositories being added a second time while waiting on processing.


app/workers/repository_maintenance_stat_worker.rb
```diff
@@ -0,0 +1,25 @@
+class RepositoryMaintenanceStatWorker
+    include Sidekiq::Worker
+    sidekiq_options queue: :repo_maintenance_stat, lock: :until_and_while_executing, unique_across_queues: true, retry: 3, on_conflict: :log
```
katzj: Probably worth a comment about why the various options are chosen here

app/workers/repository_maintenance_stat_worker.rb
```diff
@@ -0,0 +1,25 @@
...
+
+    def perform(repo_id)
+        GatherRepositoryMaintenanceStats.gather_stats(Repository.find_by_id(repo_id))
+    end
+
+    def self.queue(repo_id, priority: :medium)
```
katzj: Maybe `self.enqueue`? I think of `queue` as the thing and `enqueue` as putting an object into the queue

app/workers/repository_maintenance_stat_worker.rb
mikeyoung85: I added a comment to describe the purpose of the settings and what I expect the behavior to be. I can also describe each setting and its expected behavior to if that makes sense.

PR #2292, opened 2019-01-11
Rake Task to Process Repository Maintenance Stats

mikeyoung85:
Added rake tasks to update a project's maintenance stats and to fill in those stats for project's without any.

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -38,6 +38,9 @@ class Project < ApplicationRecord
   has_one :readme, through: :repository
   has_many :repository_maintenance_stats, through: :repository
 
+  scope :least_recently_updated_stats, -> { joins(:repository_maintenance_stats).group('projects.id').where.not(repository: nil).order('max(repository_maintenance_stats.updated_at) ASC') }
+  scope :no_existing_stats, -> { includes(:repository_maintenance_stats).where(repository_maintenance_stats: {id: nil}).where.not(repository: nil) }
+
   scope :platform, ->(platform) { where(platform: PackageManager::Base.format_name(platform)) }
   scope :lower_platform, ->(platform) { where('lower(projects.platform) = ?', platform.try(:downcase)) }
   scope :lower_name, ->(name) { where('lower(projects.name) = ?', name.try(:downcase)) }
@@ -537,6 +540,10 @@ class Project < ApplicationRecord
     raise ActiveRecord::RecordNotFound if version.nil?
 
     version
+  end
+
+  def update_maintenance_stats_async(priority: :medium)
+    RepositoryMaintenanceStatWorker.enqueue(repository.id, priority: priority) unless repository.nil?
   end
 
 end
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -38,7 +38,7 @@ class Project < ApplicationRecord
   has_one :readme, through: :repository
   has_many :repository_maintenance_stats, through: :repository
 
-  scope :least_recently_updated_stats, -> { joins(:repository_maintenance_stats).group('projects.id').where.not(repository: nil).order('max(repository_maintenance_stats.updated_at) ASC') }
+  scope :least_recently_updated_stats, -> { joins(:repository_maintenance_stats).group('projects.id').where.not(repository: nil).order(Arel.sql('max(repository_maintenance_stats.updated_at) ASC')) }
   scope :no_existing_stats, -> { includes(:repository_maintenance_stats).where(repository_maintenance_stats: {id: nil}).where.not(repository: nil) }
 
   scope :platform, ->(platform) { where(platform: PackageManager::Base.format_name(platform)) }
@@ -274,7 +274,7 @@ class Project < ApplicationRecord
 
   def set_dependents_count
     return if destroyed?
-    new_dependents_count = dependents.joins(:version).pluck('DISTINCT versions.project_id').count
+    new_dependents_count = dependents.joins(:version).pluck(Arel.sql('DISTINCT versions.project_id')).count
     new_dependent_repos_count = dependent_repos_fast_count
 
     updates = {}
--- a/lib/tasks/projects.rake
+++ b/lib/tasks/projects.rake
@@ -119,4 +119,18 @@ namespace :projects do
     exit if ENV['READ_ONLY'].present?
     ProjectDependentRepository.refresh
   end
+
+  supported_platforms = ['Maven', 'npm', 'Bower', 'PyPI', 'Rubygems', 'Packagist']
+
+  desc 'Create maintenance stats for projects'
+  task create_maintenance_stats: :environment do
+    exit if ENV['READ_ONLY'].present?
+    Project.no_existing_stats.where(platform: supported_platforms).limit(500).each(&:update_maintenance_stats_async)
+  end
+
+  desc 'Update maintenance stats for projects'
+  task update_maintenance_stats: :environment do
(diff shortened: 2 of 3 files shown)


app/models/project.rb
```diff
@@ -38,6 +38,9 @@ class Project < ApplicationRecord
   has_one :readme, through: :repository
   has_many :repository_maintenance_stats, through: :repository
 
+  scope :least_recently_updated_stats, -> { joins(:repository_maintenance_stats).group('projects.id').where.not(repository: nil).order(Arel.sql('max(repository_maintenance_stats.updated_at) ASC')) }
```
mikeyoung85: This query should find project's with stats, find the latest updated date for that project's stats, and then return them in ascending order. The idea is to find projects that have stats already and find the ones with the oldest stats so that they can be refreshed.

`Arel.sql()` was added to get rid of some deprecation warnings that are shown from rails.

app/models/project.rb
```diff
@@ -38,6 +38,9 @@ class Project < ApplicationRecord
...
+  scope :no_existing_stats, -> { includes(:repository_maintenance_stats).where(repository_maintenance_stats: {id: nil}).where.not(repository: nil) }
```
mikeyoung85: This query is meant to find projects that have a repository with no stats at all.

lib/tasks/projects.rake
```diff
@@ -119,4 +119,18 @@ namespace :projects do
...
+  desc 'Create maintenance stats for projects'
+  task create_maintenance_stats: :environment do
+    exit if ENV['READ_ONLY'].present?
+    Project.no_existing_stats.where(platform: supported_platforms).limit(500).each(&:update_maintenance_stats_async)
+  end
+
+  desc 'Update maintenance stats for projects'
+  task update_maintenance_stats: :environment do
+    exit if ENV['READ_ONLY'].present?
+    Project.least_recently_updated_stats.where(platform: supported_platforms).limit(500).each{|project| project.update_maintenance_stats_async(priority: :low)}
```
mikeyoung85: This enqueues the update worker in low priority.

mellisio: Any reason to not use find_each and just churn through it?

mikeyoung85: That is one option. The way it is currently set up you would not be able to add the same project to a higher priority queue if it is already enqueued at a lower priority. So I think we would want to change that. I figured this rake task would be called from a cron job, potentially every hour or so. We could instead put all projects on the queue and run that every few weeks instead.

app/models/project.rb
mikeyoung85: Explain:

```
 QUERY PLAN
------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Sort  (cost=1886.36..1889.40 rows=1214 width=417)
   Sort Key: (max(repository_maintenance_stats.updated_at))
   ->  HashAggregate  (cost=1812.03..1824.17 rows=1214 width=417)
         Group Key: projects.id
         ->  Merge Join  (cost=207.52..1805.96 rows=1214 width=417)
               Merge Cond: (repositories.id = projects.repository_id)
               ->  Merge Join  (cost=0.57..2782.16 rows=27024 width=20)
                     Merge Cond: (repositories.id = repository_maintenance_stats.repository_id)
                     ->  Index Only Scan using repositories_pkey on repositories  (cost=0.29..1483.00 rows=25298 width=4)
                     ->  Index Scan using index_repository_maintenance_stats_on_repository_id on repository_maintenance_stats  (cost=0.29..2195.42 rows=27024 width=16)
               ->  Sort  (cost=206.94..210.69 rows=1500 width=409)
                     Sort Key: projects.repository_id
                     ->  Seq Scan on projects  (cost=0.00..127.81 rows=1500 width=409)
                           Filter: (repository_id IS NOT NULL)

```

app/models/project.rb
mikeyoung85: SQL

```
SELECT project.*, repository_maintenance_stats.* 
FROM projects 
LEFT OUTER JOIN repositories ON repositories.id = projects.repository_id 
LEFT OUTER JOIN repository_maintenance_stats ON repository_maintenance_stats.repository_id = repositories.id 
WHERE repository_maintenance_stats.id IS NULL AND projects.repository_id IS NOT NULL
```

It is not actually selecting project.* and repository_maintenance_stats.* but is selecting each individual column as an alias, but I think it is effectively every column.

mikeyoung85: Explain plan:

```
QUERY PLAN
--------------------------------------------------------------------------------------------------------------------
 Hash Right Join  (cost=1101.48..1829.08 rows=1 width=465)
   Hash Cond: (repository_maintenance_stats.repository_id = repositories.id)
   Filter: (repository_maintenance_stats.id IS NULL)
   ->  Seq Scan on repository_maintenance_stats  (cost=0.00..610.24 rows=27024 width=56)
   ->  Hash  (cost=1082.73..1082.73 rows=1500 width=413)
         ->  Merge Right Join  (cost=207.23..1082.73 rows=1500 width=413)
               Merge Cond: (repositories.id = projects.repository_id)
               ->  Index Only Scan using repositories_pkey on repositories  (cost=0.29..1483.00 rows=25298 width=4)
               ->  Sort  (cost=206.94..210.69 rows=1500 width=409)
                     Sort Key: projects.repository_id
                     ->  Seq Scan on projects  (cost=0.00..127.81 rows=1500 width=409)
                           Filter: (repository_id IS NOT NULL)
```

app/models/project.rb
mikeyoung85: We should probably have the `repository_maintenance_stats.updated_at` indexed. I can try that locally to see if it affects that HashAggregate cost.

katzj: The updated_at worries me a little less than the sequential scan on projects to get ones without repository_id.

app/models/project.rb
katzj: Similarly the repository_id is not null bit looks like it could get large

app/models/project.rb
mikeyoung85: I think I can take out the part where it is looking for a non null value for `repository_id`. I thought that would help limit the extent of the query but may not be worth it given this plan result.

mikeyoung85: This is without the `nil` check on `repository_id`:
```
QUERY PLAN
------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Sort  (cost=1950.91..1954.91 rows=1602 width=417)
   Sort Key: (max(repository_maintenance_stats.updated_at))
   ->  HashAggregate  (cost=1849.62..1865.64 rows=1602 width=417)
         Group Key: projects.id
         ->  Merge Join  (cost=236.86..1841.61 rows=1602 width=417)
               Merge Cond: (repositories.id = projects.repository_id)
               ->  Merge Join  (cost=0.57..2782.16 rows=27024 width=20)
                     Merge Cond: (repositories.id = repository_maintenance_stats.repository_id)
                     ->  Index Only Scan using repositories_pkey on repositories  (cost=0.29..1483.00 rows=25298 width=4)
                     ->  Index Scan using index_repository_maintenance_stats_on_repository_id on repository_maintenance_stats  (cost=0.29..2195.42 rows=27024 width=16)
               ->  Sort  (cost=236.29..241.24 rows=1981 width=409)
                     Sort Key: projects.repository_id
                     ->  Seq Scan on projects  (cost=0.00..127.81 rows=1981 width=409)
```

It looks like there is still a sequential scan of the projects table.

Issue #2147, opened 2018-05-16
No results for search 'thymeleaf-extras-springsecurity4'


fapdash:
Maven: https://search.maven.org/#search%7Cga%7C1%7Ca%3A%22thymeleaf-extras-springsecurity4%22

movermeyer:
FWIW: it is available on libraries.io: https://libraries.io/maven/org.thymeleaf.extras:thymeleaf-extras-springsecurity4

Not sure why the search isn't finding it.

katzj:
This will get fixed as the current reindex of maven packages finishes

Issue #2148, opened 2018-05-16
No results for search 'spring-boot-email-core'


fapdash:
Maven: https://search.maven.org/#search%7Cga%7C1%7Cspring-boot-email-core

movermeyer:
FWIW: it is available on libraries.io: https://libraries.io/maven/it.ozimov:spring-boot-email-core.

I'm not sure why the search doesn't find it.

katzj:
This will get fixed as the current reindex of maven packages finishes

PR #2294, opened 2019-01-16
struts2 for example is org.apache.struts:struts2-core, and previously…

mellisio:


app/models/concerns/project_search.rb
```diff
@@ -51,6 +52,14 @@ def exact_name
       name
     end
 
+    def maven_name
```
katzj: I wonder if there's a more generic thing to think about here given things like the `@foo/bar` style npm packages, etc.

mellisio: Those ones work, I've only noticed a problem with Java so far. https://libraries.io/search?q=cli-plugin-eslint That being said, not an expert with elasticsearch so I don't know if there's a better way.

katzj: This is the right approach... I just think that we want to do it a little more generally. And actually thinking about it, we probably want both groupId and artifactId to get indexed nicely.

I wonder if the thing to do is to have a field that's `extra_names` (or something -- I don't like that name but my ☕️ hasn't kicked in yet) which is an array of additional searchable words for a package. And then on a per ecosystem basis we return what those would be

katzj:
This will fix #2147 and #2148

PR #2297, opened 2019-01-28
Add model and endpoint to prioritize maintenance stats for projects

mellisio:


config/routes.rb
```diff
@@ -58,6 +58,8 @@
     PROJECT_CONSTRAINT = /[^\/]+/
     VERSION_CONSTRAINT = /[\w\.\-]+/
 
+    get '/priority/:platform/:name/:priority', to: 'project_update_priority#update', constraints: { :platform => PLATFORM_CONSTRAINT, :name => PROJECT_CONSTRAINT }
```
mikeyoung85: Should this be a `post` or `put` instead of `get`?

mikeyoung85: Also I think we should consider a batch input for this instead of calling these one by one.

db/migrate/20190128142743_create_project_update_priorities.rb
```diff
@@ -0,0 +1,9 @@
+class CreateProjectUpdatePriorities < ActiveRecord::Migration[5.2]
+  def change
+    create_table :project_update_priorities do |t|
+      t.references :project, unique: true
+      t.integer :priority, default: 0
```
mikeyoung85: Should this map to a string instead? Does the integer go off the cardinality of the enum in the class? I think it is a little more work to make it map to a string.

db/migrate/20190128142743_create_project_update_priorities.rb
```diff
@@ -0,0 +1,9 @@
...
+      t.timestamps
+    end
```
mikeyoung85: Can we add an index to make it unique on project id so we don't have a project in multiple priorities? If not, we should make the project + priority a unique combo.

lib/tasks/projects.rake
```diff
@@ -133,6 +133,10 @@ namespace :projects do
   task :update_maintenance_stats, [:number_to_sync] => :environment do |_task, args|
     exit if ENV['READ_ONLY'].present?
     number_to_sync = args.number_to_sync || 2000
-    Project.least_recently_updated_stats.where(platform: supported_platforms).limit(number_to_sync).each{|project| project.update_maintenance_stats_async(priority: :low)}
+    Project.low_priority_updates.where(platform: supported_platforms).limit(number_to_sync).each{|project| project.update_maintenance_stats_async(priority: :low)}
+    Project.medium_priority_updates.where(platform: supported_platforms).limit(number_to_sync).each{|project| project.update_maintenance_stats_async(priority: :medium)}
+    Project.high_priority_updates.where(platform: supported_platforms).limit(number_to_sync).each{|project| project.update_maintenance_stats_async(priority: :high)}
```
mikeyoung85: Totally optional but I think it would be nice to be able to specify the number per priority. I don't think it is going to be an issue any time soon though.

spec/requests/project_update_priorities_spec.rb
```diff
@@ -0,0 +1,35 @@
+require "rails_helper"
+
+describe "HooksController" do
```
mikeyoung85: Can you add some tests to verify the new SQL queries in `Project`.

app/models/project_update_priority.rb
```diff
@@ -0,0 +1,16 @@
...
+
+  def self.low_priority_date
+    1.month.ago
+  end
+
+  def self.medium_priority_date
+    2.weeks.ago
+  end
+
+  def self.high_priority_date
```
mikeyoung85: I think we should be careful utilizing these dates per priority. We may end up in a situation where we can't keep up with the amount of projects in a certain priority if we dump a lot in one priority because the time limit we are checking for is too small.

PR #2298, opened 2019-01-29
AuthTokens look to have been created manually in the past. Now update…

mellisio:


mikeyoung85:
Any idea how many tokens this would clear out?

mellisio:
No idea.

app/models/concerns/github_identity.rb
```diff
@@ -34,9 +34,21 @@ def update_repo_permissions_async
     SyncPermissionsWorker.perform_async(self.id)
   end
 
+  def update_auth_token
+    at = AuthToken.find_or_create_by(login: nickname)
+    at.token = token
+    if at.still_authorized?
```
mikeyoung85: I think my only concern, besides not really knowing what other code would be affected by this, is `still_authorized?` on `AuthToken` could return false if the token has hit its rate limit at the time this method is run and would be deleted when running the rake task. I don't know if that is really a valid concern since that is a very low chance of happening.

lib/tasks/users.rake
```diff
@@ -4,4 +4,15 @@ namespace :users do
     exit if ENV['READ_ONLY'].present?
     User.where(currently_syncing: false).optin.order('last_synced_at ASC').where('last_synced_at < ?', 1.week.ago).limit(100).each(&:update_repo_permissions_async)
   end
+
+  task fix_auth_tokens: :environment do
+    exit if ENV['READ_ONLY'].present?
+    AuthToken.authorized.find_each do |token|
+      if token.still_authorized?
```
mikeyoung85: I think my only concern, besides not really knowing what other code would be affected by this, is still_authorized? on AuthToken could return false if the token has hit its rate limit at the time this method is run and would be deleted when running the rake task. I don't know if that is really a valid concern since that is a very low chance of happening.

mellisio: rate limit doesn't return true or false it returns a struct. `#<struct Octokit::RateLimit limit=5000, remaining=4966, resets_at=2019-02-01 11:05:19 -0500, resets_in=1720>`. The remaining would just be 0. The check just makes sure you don't get an auth failure. Valid tokens still get added on user permission syncs from the accounts that we still have access to.

mellisio:
@havocp All of the tokens are also in the identities table. Since we're just using github tokens for the api calls.

PR #2299, opened 2019-01-30
Add repository contributor stats

mellisio:


app/models/gather_repository_maintenance_stats.rb
```diff
@@ -45,6 +45,13 @@ def self.gather_stats(repository)
             Rails.logger.warn(e.message)
         end
 
+        begin
+          result = MaintenanceStats::Queries::RepositoryContributorStatsQuery.new(v3_client).query(params: {full_name: repository.full_name})
```
mikeyoung85: Can you just get rid of `RepositoryContributorsQuery` in this PR too. This seems like a much better API endpoint.

PR #2301, opened 2019-01-31
Add a begin method that checks for empty maintenance stats to avoid D…

mellisio:


config/routes.rb
```diff
@@ -59,6 +59,7 @@
     VERSION_CONSTRAINT = /[\w\.\-]+/
 
     get '/maintenance/stats/enqueue/:platform/:name', to: 'maintenance_stats#enqueue'
+    get '/maintenance/stats/begin/:platform/:name', to: 'maintenance_stats#begin_watching'
```
mikeyoung85: We should add the platform and name constraints here that the other routes are using.

mellisio: Can't believe I forgot! The reason why I put them below the constraints was to use them, just forgot to put them on...

app/controllers/api/maintenance_stats_controller.rb
```diff
@@ -3,6 +3,11 @@ class Api::MaintenanceStatsController < Api::ApplicationController
 
   def enqueue
     @project.update_maintenance_stats_async(priority: :high)
-    head :ok
+    head :accepted
+  end
+
+  def begin_watching
+    @project.update_maintenance_stats_async(priority: high) if @project.repository_maintenance_stats.length == 0
```
mikeyoung85: Can we use `empty?` here instead of `length == 0`?

mikeyoung85: Or maybe `exists?`. I think that is supposed to be the most efficient version of this.

PR #2307, opened 2019-02-12
Update project search to allow extending the searchable items when ne…

mellisio:


app/models/concerns/project_search.rb
```diff
@@ -51,6 +52,14 @@ def exact_name
       name
     end
 
+    def extra_searchable_names
+      if platform == "Maven"
+        name.split(":")
+      else
+        [name]
```
katzj: For other platforms, can we get by with just an empty list for now?

app/models/concerns/project_search.rb
```diff
@@ -51,6 +52,14 @@ def exact_name
       name
     end
 
+    def extra_searchable_names
+      if platform == "Maven"
+        name.split(":")
```
katzj: So this will let searches return both group id and artifact id? That sounds reasonable I think

mellisio: yes.

katzj:
It looks like we can add the field to the existing index with something like
```bash
curl -X PUT http://localhost:9200/projects/_mapping/project -H "Content-Type: application/json" -d'{"properties": {"extra_searchable_names": {"type": "string"}}}'
```

Then it'll get updated as projects update. We can force a resync of the maven stuff to get them which is the primary set we know about for now.

Issue #2245, opened 2018-11-06
Translate scm:git:git@github repo urls


katzj:
Some maven projects have a repository url in their pom.xml that looks like `scm:git:git@github.com:sbt/junit-interface.git`. We should translate those to be `https://github.com` urls

AgonIdrizi:
Hey,
Can i contribute, to correct those urls.
Thnx

katzj:
Will get fixed up as we update maven packages

PR #2322, opened 2019-03-11
Reformat Project repository_url On Save

mikeyoung85:
This might help with a few repositories that have an old poorly formatted repository url and do not find a new one to reformat on sync.


katzj:
Fixes #2245

PR #2328, opened 2019-03-21
Move deployment to be totally kubernetes based

katzj:
Remove capistrano and just use k8s for database migrations. Deploys now happen by running `./bin/deploy.sh`


Gemfile
```diff
@@ -92,14 +92,6 @@ group :development do
   gem 'spring'
   gem 'sql_queries_count'
   gem 'brakeman', :require => false
-  gem 'capistrano'
```
jsonperl: 👋 capistrano

PR #2334, opened 2019-03-27
Define ProjectSearch#marshal_load to match marshal_dump

kbarrette:
Looks like this may have been broken in #2269

You must define `marshal_load` if you define `marshal_dump` (https://ruby-doc.org/core-2.6/Marshal.html)

### DIFF
--- a/spec/models/concerns/project_search_spec.rb
+++ b/spec/models/concerns/project_search_spec.rb
@@ -1,0 +1,12 @@
+require 'rails_helper'
+
+RSpec.describe ProjectSearch do
+  subject { build(:project) }
+
+  it "Marshals properly" do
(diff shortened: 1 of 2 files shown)


spec/models/concerns/project_search_spec.rb
```diff
@@ -0,0 +1,12 @@
+require 'rails_helper'
+
+RSpec.describe ProjectSearch do
+  subject { build(:project) }
+
+  it "Marshals properly" do
+    original = subject.to_json
+    result = Marshal.load(Marshal.dump(subject)).to_json
```
kbarrette: Using JSON here since the objects are not directly comparable (they are clones)

PR #2363, opened 2019-05-29
Maven: Look up Parent POM for Information

mikeyoung85:
This will look into the POM xml for an artifact and if there is a parent specified it will also look up the parent POM file recursively. Values are overwritten by the child POM file and missing values are carried down from the parent. This should hopefully help with some of the Maven data being missing or incorrect.

### DIFF
--- a/app/models/package_manager/maven.rb
+++ b/app/models/package_manager/maven.rb
@@ -83,13 +83,26 @@ module PackageManager
       else
         xml = version_xml
       end
-      {
+
+      keys = %i[description homepage repository_url licenses]
+      parent = Hash[keys.product([nil])]
+      if xml.locate('parent').first
+        group_id = xml.locate('parent/groupId').first.try(:nodes).try(:first)
+        artifact_id = xml.locate('parent/artifactId').first.try(:nodes).try(:first)
+        unless [group_id, artifact_id].any?(nil)
+          parent = mapping(project([group_id, artifact_id].join(":")))
+        end
+      end
+
+      # merge with parent data if available and take child values on overlap
+      child = {
         description: xml.locate('description').first.try(:nodes).try(:first),
         homepage: xml.locate('url').first.try(:nodes).try(:first),
         repository_url: repo_fallback(xml.locate('scm/url').first.try(:nodes).try(:first),
                                       xml.locate('url').first.try(:nodes).try(:first)),
         licenses: xml.locate('licenses/license/name').map{|l| l.nodes}.flatten.join(",")
-      }
+      }.reject!{|k,v| v.nil?}
+      parent.merge(child)
     end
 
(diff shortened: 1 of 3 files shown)


app/models/package_manager/maven.rb
```diff
@@ -83,13 +83,26 @@ def self.mapping_from_pom_xml(version_xml)
...
         xml = version_xml
       end
-      {
+
+      keys = %i[description homepage repository_url licenses]
+      parent = Hash[keys.product([nil])]
+      if xml.locate('parent').first
+        group_id = xml.locate('parent/groupId').first.try(:nodes).try(:first)
+        artifact_id = xml.locate('parent/artifactId').first.try(:nodes).try(:first)
+        unless [group_id, artifact_id].any?(nil)
```
kbarrette: `if group_id && artifact_id` would suffice, I think?

app/models/package_manager/maven.rb
```diff
@@ -83,13 +83,26 @@ def self.mapping_from_pom_xml(version_xml)
       else
         xml = version_xml
       end
-      {
+
+      keys = %i[description homepage repository_url licenses]
+      parent = Hash[keys.product([nil])]
+      if xml.locate('parent').first
+        group_id = xml.locate('parent/groupId').first.try(:nodes).try(:first)
```
kbarrette: Could use `&.` over `try`

app/models/package_manager/maven.rb
```diff
@@ -83,13 +83,26 @@ def self.mapping_from_pom_xml(version_xml)
...
-        repository_url: repo_fallback(xml.locate('scm/url').first.try(:nodes).try(:first),
-                                      xml.locate('url').first.try(:nodes).try(:first)),
+
+      keys = %i[description homepage repository_url licenses]
+      parent = Hash[keys.product([nil])]
+      if xml.locate('parent').first
+        group_id = xml.locate('parent/groupId').first&.nodes&.first
+        artifact_id = xml.locate('parent/artifactId').first&.nodes&.first
+        if group_id && artifact_id
+          parent = mapping(project([group_id, artifact_id].join(":")))
```
katzj: I wonder if we want to have a maxdepth to avoid someone doing an incredibly deep pom nesting and causing us probelms

mikeyoung85: I don't think that would be an issue, but it is something to protect against. What would a reasonable depth limit be? I've never seen a package with more than 2-3 parent projects. Maybe setting the max depth to something like 5?

katzj: 5 sounds totally good to me

mikeyoung85:
Does it make sense to roll up dependencies from parent poms as well? I think that is a separate set of logic.

PR #2367, opened 2019-05-31
Fix two bugs with handling spdx license expressions

katzj:


spec/models/project_spec.rb
```diff
@@ -51,6 +51,12 @@
       expect(project.normalized_licenses).to eq(["MIT", "ISC"])
     end
 
+    it 'handles (OR) separated licenses', focus: true do
```
kbarrette: Probably want to nuke the `focus: true`

PR #2369, opened 2019-05-31
Improve java license handling

kbarrette:


app/models/package_manager/maven.rb
```diff
@@ -134,5 +138,34 @@ def self.project_name(name)
...
+    end
+
+    def self.licenses(xml)
+      xml_licenses = xml
+        .locate('licenses/license/name')
+        .flat_map(&:nodes)
+      return xml_licenses if xml_licenses.any?
+
+      LICENSE_STRINGS
+        .select { |string, _| xml.nodes.first&.value&.include?(string) }
```
kbarrette: This will only work if the license comment is the first element - think it should be more flexible?

app/models/package_manager/maven.rb
```diff
@@ -7,6 +7,9 @@ class Maven < Base
     URL = "http://maven.org"
     BASE_URL = "https://maven-repository.com"
     COLOR = '#b07219'
+    LICENSE_STRINGS = {
+      "http://www.apache.org/licenses/LICENSE-2.0" => "Apache-2.0"
```
mikeyoung85: Can you add the Eclipse license in here too? "http://www.eclipse.org/legal/epl-v10.html"

app/models/package_manager/maven.rb
```diff
@@ -134,5 +138,34 @@ def self.project_name(name)
...
     end
+
+    def self.get_pom(group_id, artifact_id, version, seen=[])
+      seen << [group_id, artifact_id, version]
+
+      xml = get_xml(
+        "http://repo1.maven.org/maven2/#{group_id}/#{artifact_id}/#{version}/#{artifact_id}-#{version}.pom"
+      )
+
+      next_group_id = xml.locate('distributionManagement/relocation/groupId').first&.nodes&.first || group_id
```
mikeyoung85: Spicy

app/models/package_manager/maven.rb
```diff
@@ -134,5 +138,34 @@ def self.project_name(name)
...
+      next_artifact_id = xml.locate('distributionManagement/relocation/artifactId').first&.nodes&.first || artifact_id
+      next_version = xml.locate('distributionManagement/relocation/version').first&.nodes&.first || version
+
+      if !seen.include?([next_group_id, next_artifact_id, next_version])
+        get_pom(next_group_id, next_artifact_id, next_version, seen)
```
mikeyoung85: What happens if we aren't able to locate this pom? I think the way this is now, it would error out completely. Can we default to return this pom if there is an error following the relocation properties?

app/models/package_manager/maven.rb
mikeyoung85: We could load all comments here: `xml.locate('*/^Comment')` I think. Or maybe just top level comments and iterate through each.

app/models/package_manager/maven.rb
kbarrette: I'd actually like to simplify this, if possible - thoughts?

kbarrette:
@mikeyoung85 fixed as suggested

app/models/package_manager/maven.rb
```diff
@@ -79,6 +84,7 @@ def self.mapping_from_pom_xml(version_xml, depth = 0)
       else
         xml = version_xml
       end
+<<<<<<< HEAD
```
mikeyoung85: Looks like a merge issue in here.

app/models/package_manager/maven.rb
```diff
@@ -148,5 +163,40 @@ def self.project_name(name)
...
+      end
+    end
+
+    def self.licenses(xml)
+      xml_licenses = xml
+        .locate('licenses/license/name')
+        .flat_map(&:nodes)
+      return xml_licenses if xml_licenses.any?
+
+      comments = xml.locate('*/^Comment')
```
mikeyoung85: Does this actually work? I didn't actually get a chance to test it in my recommendation 😄

kbarrette: The test passes!

app/models/package_manager/maven.rb
kbarrette: Whoops.

PR #2374, opened 2019-06-03
When there are no versions available from maven-repository.com, fall back to DB

kbarrette:


app/models/package_manager/maven.rb
```diff
@@ -193,5 +190,22 @@ def self.licenses(xml)
...
+
+    def self.latest_version(project)
+      if project[:versions].present?
+        project[:versions]
+          .max_by { |version| version[:published_at] }
+          .dig(:number)
+      else
+        # TODO this is in place to handle packages that are no longer on maven-repository.com
+        # this could be removed if we switched to a package data provider that supplied full information
+        Project
```
mikeyoung85: I think this should help a bit. We will run into an issue if the pom declares a parent that is also not found. I don't know how likely we have the parent poms as projects. I think you have split out the pom gathering methods. So I think we could switch:
`parent = mapping(project([group_id, artifact_id].join(":")), depth += 1)` on line 95 with something like `parent = get_pom(project[:groupPath], project[:artifactId], latest_version(project))` instead of calling `mapping`.

PR #2375, opened 2019-06-04
Properly construct urls for relocated maven packages

kbarrette:


kbarrette:
@katzj after some discussion with @mikeyoung85, we found a few more areas with similar problems - updated with a slightly larger change set that should catch them all

PR #2376, opened 2019-06-05
More maven bugfixes

kbarrette:


app/models/package_manager/maven.rb
```diff
@@ -59,13 +59,19 @@ def self.recent_names
...
         name: name,
-        path: sections.join('/'),
+        path: path,
         groupId: sections[0],
         artifactId: sections[1],
-        versions: versions(h),
+        versions: versions,
       }
+    rescue
+      {}
```
mikeyoung85: I think passing back an empty Hash will cause an error further downstream. I don't think the rescue will help.

app/models/package_manager/maven.rb
```diff
@@ -59,13 +59,19 @@ def self.recent_names
 
     def self.project(name)
       sections = name.split(':')
+      path = sections.join('/')
+      versions = versions({ path: path })
+      return {} unless versions.present?
```
mikeyoung85: Are we nearly back to the original version of this method?

app/models/package_manager/maven.rb
```diff
@@ -106,7 +112,7 @@ def self.mapping_from_pom_xml(version_xml, depth = 0)
           xml.locate('scm/url/?[0]').first,
           xml.locate('url/?[0]').first
         ),
-        licenses: licenses(xml).join(","),
+        licenses: licenses(version_xml).join(","),
```
mikeyoung85: Isn't `xml` the "canonical" xml in this method?

app/models/package_manager/maven.rb
kbarrette: `Base` checks `project.present?` - this is the same pattern used in other package manager classes to signal "do nothing"

app/models/package_manager/maven.rb
kbarrette: Nope - we call `version_xml.project` which doesn't include the top-level comment (line 84)

app/models/package_manager/maven.rb
kbarrette: 🤷‍♂

app/models/package_manager/maven.rb
mikeyoung85: Could we pass back the Hash with an empty versions if the versions call fails?

PR #2380, opened 2019-06-10
Expand Maven Properties

mikeyoung85:
Use Bibliothecary to expand properties in POM files. Gather properties from parent projects and send those in alongside child pom files.

### DIFF
--- a/app/models/package_manager/maven.rb
+++ b/app/models/package_manager/maven.rb
@@ -126,13 +126,9 @@ module PackageManager
     end
 
     def self.extract_pom_properties(xml)
-      props = {}
-      if xml.respond_to? 'properties'
-        xml.properties.each do |prop_node|
-          props[prop_node.value] = prop_node.nodes.first
-        end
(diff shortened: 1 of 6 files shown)


mikeyoung85:
Before: spring_before

After: spring_with_props

app/models/package_manager/maven.rb
```diff
@@ -106,21 +107,37 @@ def self.mapping_from_pom_xml(version_xml, depth = 0)
...
     end
 
+    def self.extract_pom_value(xml, location, parent_properties = {})
+      # Bibliothecary will help handle property expansion within the xml
+      Bibliothecary::Parsers::Maven.extract_pom_info(xml, location, parent_properties)
+    end
+
+    def self.extract_pom_properties(xml)
+      props = {}
+      if xml.respond_to? 'properties'
```
kbarrette: Can we just do `xml.locate("properties").each_with_object...`?

PR #2382, opened 2019-06-12
Change Project Name Constraint to Fix Project Rendering Issue

mikeyoung85:
Change constraint to allow for packages with `/`.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -58,7 +58,7 @@ Rails.application.routes.draw do
     PROJECT_CONSTRAINT = /[^\/]+/
     VERSION_CONSTRAINT = /[\w\.\-]+/
 
-    put '/maintenance/stats/enqueue/:platform/:name', as: :maintenance_stat_enqueue, to: 'maintenance_stats#enqueue', constraints: { :platform => PLATFORM_CONSTRAINT, :name => /.*/ }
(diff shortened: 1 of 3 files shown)


config/routes.rb
```diff
@@ -58,7 +58,7 @@
     PROJECT_CONSTRAINT = /[^\/]+/
     VERSION_CONSTRAINT = /[\w\.\-]+/
 
-    put '/maintenance/stats/enqueue/:platform/:name', as: :maintenance_stat_enqueue, to: 'maintenance_stats#enqueue', constraints: { :platform => PLATFORM_CONSTRAINT, :name => PROJECT_CONSTRAINT }
+    put '/maintenance/stats/enqueue/:platform/:name', as: :maintenance_stat_enqueue, to: 'maintenance_stats#enqueue', constraints: { :platform => PLATFORM_CONSTRAINT, :name => /.*/ }
     get '/maintenance/stats/begin/:platform/:name', to: 'maintenance_stats#begin_watching', constraints: { :platform => PLATFORM_CONSTRAINT, :name => PROJECT_CONSTRAINT }
```
mikeyoung85: I don't know if these other routes should have the same constraint. There doesn't seem to be a clear pattern in the routes to me.

katzj: Should we instead be url encoding the name when creating this link?

mikeyoung85: I added encoding for that param and decoding in the `application_controller`.  I don't know how I feel about putting the decoding in such a fundamental method. The other option is to create something similar to what is in place for Project sync instead of embedding an API call in the frontend.

katzj: The little route that doesn't require the encoding and isn't an api route sounds even better :)

mikeyoung85:
Going to go a different route

PR #2384, opened 2019-06-12
Handle Hyphens and Underscores for Pypi Projects

mikeyoung85:
Still need to do some more testing locally to make sure nothing is screwed up with this, but wanted opinions. This creates a method in package managers to have a suggested list of names to use to find projects. For all the cases except Pypi, this is just the project name. For Pypi this includes replacing hyphens with underscores and vice versa. This is to help mimic the same mechanic that Pypi supports and would help find projects when reading Python dependency manifests.

### DIFF
--- a/app/controllers/api/status_controller.rb
+++ b/app/controllers/api/status_controller.rb
@@ -16,8 +16,8 @@ class Api::StatusController < Api::ApplicationController
   private
 
   def find_projects(projects, platform)
-    project_find_names = projects.map{|project| project_names(project, platform)}
-    Project.platform(platform).where(name: project_find_names).includes(:repository, :versions, :repository_maintenance_stats)
+    project_find_names = projects.flat_map{|project| project_names(project, platform)}.map(&:downcase)
+    Project.platform(platform).where('lower(platform)=? AND lower(name) in (?)', platform.downcase, project_find_names).includes(:repository, :versions, :repository_maintenance_stats)
   end
 
   def project_names(project, platform)
--- a/app/controllers/api/status_controller.rb
+++ b/app/controllers/api/status_controller.rb
@@ -6,7 +6,7 @@ class Api::StatusController < Api::ApplicationController
       # Try to get all the projects passed in.
       @projects = params[:projects]
         .group_by { |project| project[:platform] }
-        .flat_map { |platform, projects| find_projects(projects, platform) }
+        .flat_map(&method(:find_projects))
         .compact
     else
       @projects = []
@@ -16,7 +16,7 @@ class Api::StatusController < Api::ApplicationController
 
   private
 
-  def find_projects(projects, platform)
+  def find_projects((platform, projects))
     projects.each_slice(1000).flat_map do |slice|
       project_find_names = slice.flat_map { |project| project_names(project, platform) }.map(&:downcase)
       Project.platform(platform).where('lower(platform)=? AND lower(name) in (?)', platform.downcase, project_find_names).includes(:repository, :versions, :repository_maintenance_stats)
--- a/app/controllers/api/status_controller.rb
+++ b/app/controllers/api/status_controller.rb
@@ -24,10 +24,8 @@ class Api::StatusController < Api::ApplicationController
   end
 
   def project_names(project, platform)
-    begin
       "PackageManager::#{platform.capitalize}".constantize.project_find_names(project[:name])
-    rescue => exception
-      PackageManager::Base.project_find_names(project[:name])
-    end
+  rescue NameError
+    PackageManager::Base.project_find_names(project[:name])
   end
 end
--- a/app/controllers/api/status_controller.rb
+++ b/app/controllers/api/status_controller.rb
@@ -4,9 +4,10 @@ class Api::StatusController < Api::ApplicationController
   def check
     if params[:projects].any?
       # Try to get all the projects passed in.
-      @projects = params[:projects].group_by{|project| project[:platform] }.map do |platform, projects|
-        projects.each_slice(1000).map{|slice| find_projects(slice, platform)}
-      end.flatten.compact
+      @projects = params[:projects]
+        .group_by { |project| project[:platform] }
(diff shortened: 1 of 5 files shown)


app/controllers/api/status_controller.rb
```diff
@@ -5,32 +5,26 @@ def check
...
       @projects = []
     end
     render json: @projects, each_serializer: ProjectStatusSerializer, show_score: params[:score], show_stats: internal_api_key?
   end
+
+  private
+
+  def find_projects(projects, platform)
+    project_find_names = projects.flat_map{|project| project_names(project, platform)}.map(&:downcase)
+    Project.platform(platform).where('lower(platform)=? AND lower(name) in (?)', platform.downcase, project_find_names).includes(:repository, :versions, :repository_maintenance_stats)
```
mikeyoung85: This only does one query with lower cased project names instead of using those as the fallback query.

app/controllers/api/status_controller.rb
```diff
@@ -5,32 +5,26 @@ def check
     if params[:projects].any?
       # Try to get all the projects passed in.
       @projects = params[:projects].group_by{|project| project[:platform] }.map do |platform, projects|
-        projects.each_slice(1000).map do |slice|
-          Project.platform(platform).where(name: slice.map{|project| project[:name] }.uniq).includes(:repository, :versions, :repository_maintenance_stats)
-        end
+        projects.each_slice(1000).map{|slice| find_projects(slice, platform)}
```
kbarrette: Love breaking this out into another method!

Style-wise, I'd consider re-writing it as something like:
```ruby
@projects = params[:projects]
  .group_by { |project| project[:platform] }
  .flat_map(&method(:find_projects)) # moving the slicing into find_projects
  .compact
```

app/controllers/api/status_controller.rb
```diff
@@ -5,32 +5,26 @@ def check
...
     else
       @projects = []
     end
     render json: @projects, each_serializer: ProjectStatusSerializer, show_score: params[:score], show_stats: internal_api_key?
   end
+
+  private
+
+  def find_projects(projects, platform)
+    project_find_names = projects.flat_map{|project| project_names(project, platform)}.map(&:downcase)
```
kbarrette: Style would generally be to have spaces inside `{ }` and between `flat_map` and `{`, although this file looks to be doing no spaces?

app/controllers/api/status_controller.rb
```diff
@@ -5,32 +5,26 @@ def check
...
+    Project.platform(platform).where('lower(platform)=? AND lower(name) in (?)', platform.downcase, project_find_names).includes(:repository, :versions, :repository_maintenance_stats)
+  end
+
+  def project_names(project, platform)
+    begin
+      "PackageManager::#{platform.capitalize}".constantize.project_find_names(project[:name])
+    rescue => exception
```
kbarrette: You can just put the `rescue` at the method level:

```ruby
def project_names(project, platform)
  "PackageManager::#{platform.capitalize}".constantize.project_find_names(project[:name])
rescue
  PackageManager::Base.project_find_names(project[:name])
end
```

kbarrette: On second thought, is this needed? All the `PackageManager`s subclass `Base`, right? So they should have the parent method?

app/models/project.rb
```diff
@@ -516,7 +516,13 @@ def download_registry_users
   def self.find_with_includes!(platform, name, includes)
     # this clunkiness is because .includes() doesn't allow zero args and we want
     # to allow that.
-    query = self.visible.platform(platform).where(name: name)
+    project_find_names = begin
+       "PackageManager::#{platform.capitalize}".constantize.project_find_names(name)
+    rescue => exception
```
kbarrette: Same question as above

spec/models/package_manager/pypi_spec.rb
```diff
@@ -45,4 +45,18 @@
       expect(described_class.mapping(bandit)[:licenses]).to eq("Apache Software License")
     end
   end
+
+  describe 'project_find_names' do
+    it 'suggests underscore version of name' do
+      suggested_find_names = described_class.project_find_names('test-hyphen')
+      expect(suggested_find_names.include?('test_hyphen')).to be true
```
kbarrette: You can `expect(...).to include('test_hyphen', 'test-hyphen')`

app/controllers/api/status_controller.rb
mikeyoung85: This is the to cover off if there is an issue getting the PackageManager class. Currently if you pass in a bad platform, the API just returns empty results. This would attempt to mimic that instead of returning a 500.

kbarrette: Got it - can we just rescue the `NameError`, then?

app/models/project.rb
mikeyoung85: Here the rescue is meant to raise a `RecordNotFound` error in place of an error of not being able to find the `PackageManager` class.

app/controllers/api/status_controller.rb
```diff
@@ -4,33 +4,28 @@ class Api::StatusController < Api::ApplicationController
...
-            # lowercase compare the downcased_names... use lower(platform) to improve index usage
-            Project.where('lower(platform)=? AND lower(name) in (?)', platform.downcase, downcased_names).includes(:repository, :versions, :repository_maintenance_stats)
-          end
-        end.flatten.compact
-        # Concatanate any new items found.
-        @projects.concat(lowercase_project_named)
-      end
+      @projects = params[:projects]
+        .group_by { |project| project[:platform] }
+        .flat_map { |platform, projects| find_projects(projects, platform) }
```
kbarrette: If you wanted to, you could destructure the args in the method def, and do `.flat_map(&method(:find_projects))`

mikeyoung85: Ok I'll try that. I had issues with it not passing the args in when I first tried it.

PR #2386, opened 2019-06-14
Basic support for metadata redirects

kbarrette:
There's a balance to be struck here - at what point do we just start using `go get`?

This only handles metadata redirects to github.com urls - we may want to expand that, but I wanted some discussion.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -75,6 +75,24 @@ class ApplicationController < ActionController::Base
   def find_project
     @project = Project.find_with_includes!(params[:platform], params[:name], [:repository, :versions])
     @color = @project.color
+  rescue ActiveRecord::RecordNotFound
+    raise if params[:name].blank? || params[:platform]&.downcase != "go"
+
+    resolved_name = PackageManager::Go.resolved_name(params[:name])
+    if resolved_name != params[:name] && Project.known?(params[:platform], resolved_name)
+      redirect_to(
+        # Unescape since url_for automatically escapes our already-escaped resolved_name
+        URI.unescape(
+          url_for(
+            params
+              .to_unsafe_h
+              .merge({ "name" => CGI.escape(resolved_name) })
+          )
+        )
+      )
+    else
+      raise
+    end
   end
 
   def find_project_lite
--- a/app/models/package_manager/go.rb
+++ b/app/models/package_manager/go.rb
@@ -5,6 +5,20 @@ module PackageManager
     BIBLIOTHECARY_SUPPORT = true
     URL = 'http://go-search.org/'
     COLOR = '#375eab'
+    KNOWN_HOSTS = [
+      'bitbucket.org',
+      'github.com',
+      'launchpad.net',
+      'hub.jazz.net',
+    ]
+    KNOWN_VCS = [
+      '.bzr',
+      '.fossil',
+      '.git',
+      '.hg',
+      '.svn',
+    ]
+
 
     def self.package_link(project, version = nil)
       "http://go-search.org/view?id=#{project.name}"
@@ -33,6 +47,23 @@ module PackageManager
         homepage: project['ProjectURL'],
         repository_url: "https://#{project['Package']}"
       }
+    end
+
+    # https://golang.org/cmd/go/#hdr-Import_path_syntax
+    def self.resolved_name(name)
+      return name if name.start_with?(*KNOWN_HOSTS)
+      return name if KNOWN_VCS.any?(&name.method(:include?))
+
+      go_import = get_html('https://' + name + '?go-get=1')
+        .xpath('//meta[@name="go-import"]')
+        .first
+        &.attribute("content")
+        &.value
+        &.split(" ")
+        &.last
+        &.sub(/https?:\/\//, "")
+
+      go_import&.start_with?(*KNOWN_HOSTS) ? go_import : name
     end
   end
 end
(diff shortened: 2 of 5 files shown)


kbarrette:
@mikeyoung85 @jsonperl if you have a moment, please check this out

app/models/package_manager/go.rb
```diff
@@ -34,5 +48,18 @@ def self.mapping(project)
...
+    def self.resolved_name(name)
+      return name if name.start_with?(*KNOWN_HOSTS)
+      return name if KNOWN_VCS.any?(&name.method(:include?))
+
+      get_html('https://' + name + '/?go-get=1')
+        .xpath('//meta[@name="go-import"]')
+        .first
+        &.attribute("content")
+        &.value
+        &.[](/(github\.com.*$)/, 1) || name
```
jsonperl: That's fancy regexin!

app/models/package_manager/go.rb
```diff
@@ -34,5 +48,18 @@ def self.mapping(project)
...
       }
     end
+
+    # See `go help importpath`
+    def self.resolved_name(name)
+      return name if name.start_with?(*KNOWN_HOSTS)
+      return name if KNOWN_VCS.any?(&name.method(:include?))
+
+      get_html('https://' + name + '/?go-get=1')
+        .xpath('//meta[@name="go-import"]')
```
jsonperl: I had ZERO idea that these meta tags were a thing.

jsonperl:
Code looks good to me. Do you know of any specific failing cases that this should fix? Also wondering why we're punting on anything other than github.com?

app/models/package_manager/go.rb
```diff
@@ -34,5 +48,18 @@ def self.mapping(project)
         repository_url: "https://#{project['Package']}"
       }
     end
+
+    # See `go help importpath`
```
katzj: Can you link to the docs -- https://golang.org/cmd/go/#hdr-Import_path_syntax

kbarrette:
@mikeyoung85 @jsonperl @katzj 

Updated to handle redirects. This feels slightly risky - I'm open to suggestions for other methods.

Talking points include: Should we be caching this stuff? For how long? Should we just be looking at `repository_url` and calling that good enough? Does this open up some redirect loop problems?

pom.xml
```diff
@@ -0,0 +1,232 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
mikeyoung85: What was this added for?

kbarrette: Whoops, test file - removed it.

app/controllers/application_controller.rb
```diff
@@ -75,6 +75,24 @@ def find_platform(param = :id)
   def find_project
     @project = Project.find_with_includes!(params[:platform], params[:name], [:repository, :versions])
     @color = @project.color
+  rescue ActiveRecord::RecordNotFound
+    raise if params[:name].blank? || params[:platform].downcase != "go"
```
katzj: There's also `app/controllers/api/application_controller.rb:find_proect`. I wonder if there's a way we could have this handled in `Project.find_with_includes`? It looks like that already does some normalization. And we could then do in a way to allow resolving names for other package managers in the future

app/controllers/application_controller.rb
```diff
@@ -75,6 +75,24 @@ def find_platform(param = :id)
...
+
+    resolved_name = PackageManager::Go.resolved_name(params[:name])
+    if resolved_name != params[:name] && Project.known?(params[:platform], resolved_name)
+      redirect_to(
```
katzj: Should we redirect or just use it as `project` at this point?

app/controllers/application_controller.rb
kbarrette: Could definitely move it there

app/controllers/application_controller.rb
kbarrette: I was thinking redirect since it felt strange to serve data that would say "project name: foo" when the url said "project name: bar" - open to discussion on that though.

app/controllers/application_controller.rb
kbarrette: I think this is the only `find_project` - `Api::ApplicationController` doesn't seem to have one.

app/models/package_manager/go.rb
```diff
@@ -34,5 +48,22 @@ def self.mapping(project)
         repository_url: "https://#{project['Package']}"
       }
     end
+
+    # https://golang.org/cmd/go/#hdr-Import_path_syntax
+    def self.resolved_name(name)
+      return name if name.start_with?(*KNOWN_HOSTS)
+      return name if KNOWN_VCS.any?(&name.method(:include?))
+
+      go_import = get_html('https://' + name + '?go-get=1')
```
mikeyoung85: Could we add a new column to project to store resolved names or something like that? I'm conflicted about what to suggest on this PR. One of the worries I have is how long that redirect could take since we are making a call out to the internet here.

kbarrette: We could definitely add that column. Would we then have to worry about flushing it if a redirecting name goes away?

mikeyoung85: Eh, that's a good point. I'm not too familiar with Go and how often this stuff would occur or even how this naming thing is used in the community. It might be something we could handle in the `PackageManager::Go.update` method?

PR #2391, opened 2019-06-24
Enable Maintenance Stats for Bitbucket Projects

mikeyoung85:
Enables maintenance stats for projects with Bitbucket repositories. Gathers up stats for commits, issues, and pull requests.


app/models/gather_repository_maintenance_stats.rb
```diff
@@ -1,12 +1,33 @@
 class GatherRepositoryMaintenanceStats
+    def self.gather_bitbucket_stats(repository)
```
katzj: Would some of this make more sense in the repository hosts instead of having `gather_bitbucket` and `gather_github`?

app/models/repository_host/bitbucket.rb
```diff
@@ -70,6 +70,10 @@ def download_pull_requests(token = nil)
       nil
     end
 
+    def retrieve_commits(token = nil)
+      api_client(token).repos.commits.list(repository.owner_name, repository.project_name)['values']
```
katzj: How doe sthis work with rate limiting?

app/models/repository_issue/bitbucket.rb
```diff
@@ -40,7 +40,7 @@ def self.create_from_hash(name_with_owner, issue_hash, token = nil)
 
       i.repository_id = repository.id
       i.labels = [issue_hash.fetch(:metadata, {})[:kind]]
-      i.pull_request = issue_hash[:type] == 'pull_request'
+      i.pull_request = issue_hash[:type] == 'pullrequest'
```
katzj: Is this a bug that's existed for a whiel?

app/models/gather_repository_maintenance_stats.rb
mikeyoung85: I think that would make more sense now that I've gone through `repository_host` with this work. Then the API to get stats would just be `repository.gather_stats`. I'll shift this stuff over today.

app/models/repository_issue/bitbucket.rb
mikeyoung85: It's something I noticed while working through this. I'm not sure when that value changed, but it was what I was seeing in the live API responses. We also have a mix of V1 and V2 API calls to Bitbucket and it could differ between the two?

app/models/repository_host/bitbucket.rb
mikeyoung85: This is the only documentation I can find on rate limits: https://confluence.atlassian.com/bitbucket/rate-limits-668173227.html?_ga=2.192562154.1743157817.1561468931-1728484474.1527255352. Looks to be 1000 requests/hour on the repositories and commits. It looks like the rate is tied to the IP address if unauthenticated and the user if it is authenticated.

app/models/gather_repository_maintenance_stats.rb
```diff
@@ -1,12 +1,33 @@
...
+        metrics << MaintenanceStats::Stats::Bitbucket::IssueRates.new(repository.issues).get_stats
+
+        add_metrics_to_repo(repository, metrics)
+        metrics
+    end
+
     def self.gather_stats(repository)
         # only support Github repos for now
         # check to make sure the Project URLs are also pointing to a Github repository
         unless stats_enabled?(repository)
```
kbarrette: I think this might be easier to read as something like:
```ruby
if github?(repository)
  gather_github_stats(repository)
elsif bitbucket?(repository)
  gather_bitbucket_stats(repository)
else
  repository.repository_maintenance_stats.destroy_all
  nil
end
```

app/models/maintenance_stats/stats/bitbucket/commit_count_stats.rb
```diff
@@ -0,0 +1,21 @@
+module MaintenanceStats
+    module Stats
```
kbarrette: Too much indentation!

app/models/maintenance_stats/stats/bitbucket/commit_count_stats.rb
```diff
@@ -0,0 +1,21 @@
...
+        module Bitbucket
+            class CommitsStat < MaintenanceStats::Stats::BaseStat
+                def count_up_commits(dataset, since)
+                    return nil if dataset.nil? || dataset.empty?
```
kbarrette: `return unless dataset.present?`

app/models/maintenance_stats/stats/bitbucket/commit_count_stats.rb
```diff
@@ -0,0 +1,21 @@
...
+                    dataset.select {|commit| Date.parse(commit["date"]) >= since}.size
+                end
+
+                def get_stats
+                    {
+                        "last_week_commits": count_up_commits(@results, 1.week.ago),
```
kbarrette: Shouldn't need the quotes

app/models/maintenance_stats/stats/bitbucket/commit_count_stats.rb
```diff
@@ -0,0 +1,21 @@
...
+                        "last_month_commits": count_up_commits(@results, 1.month.ago),
```
kbarrette: Do we want to reference these all from the same time? It's possible to roll over a day or whatever and get unexpected results

app/models/maintenance_stats/stats/bitbucket/issue_stats.rb
```diff
@@ -0,0 +1,81 @@
...
+                    else
+                      issues.where(state: state).size
+                    end
+                  end
+            
+                  def closed_issues_count
+                    return @closed_issues_count unless @closed_issues_count.nil?
+            
+                    @closed_issues_count = count_issues(issues, state: "closed")
+                    @closed_issues_count
```
kbarrette: No need for this line - the value of the assignment is the assigned value

The usual pattern would be something like:
```ruby
@closed_issues_count ||= count_issues(issues, state: "closed")

File: app/models/maintenance_stats/stats/bitbucket/issue_stats.rb [hunk 10]
```diff
@@ -0,0 +1,81 @@
...
+                  def open_issues_count
+                    return @open_issues_count unless @open_issues_count.nil?
+            
+                    @open_issues_count = count_issues(issues, state: "open")
+                    @open_issues_count
+                  end
+            
+                  def issue_closure_rate
+                    return 1.0 if count_issues(issues) == 0
+                    closed_issues_count.to_f / count_issues(issues).to_f
```
kbarrette: Should we be using `BigDecimal` here for more precision, or is floating point OK?

mikeyoung85: I think floating should be fine, precision isn't a big factor here.

mikeyoung85:
Updated the new code with the suggested changes. I also moved the stats gathering under the correct `repository_host` class and now gathering is called from the repository directly, which calls the correct set of queries and stats. I also moved the Github stats under their own module to keep the two sets distinct from each other. I would like to also move the stats that way, but that will require some effort to redo the VCR mocks since those have class names embedded in the calls.

app/models/repository_host/github.rb
```diff
@@ -221,10 +225,84 @@ def download_tag(token, tag, existing_tag_names)
...
     def api_client(token = nil)
       AuthToken.fallback_client(token)
     end
+
+    def check_for_v4_error_response(response)
+      # errors can be stored in the response from Github or can be stored in the response object from HTTP errors
+      response.errors.each do |message|
+          Rails.logger.warn(message)
+      end
+      response.data.errors.each do |message|
```
kbarrette: You can one-line this like
```ruby
response.data.errors.each(&Rails.logger.method(:warn)) if response.data.errors.present?
```

app/models/maintenance_stats/stats/github/commit_count_stats.rb
```diff
@@ -0,0 +1,61 @@
...
+  module Stats
+    module Github
+      class BaseCommitCount < BaseStat
+        def pull_out_commit_count(dataset)
+            return nil if dataset.data.repository.nil? || dataset.data.repository.default_branch_ref.nil?
+            dataset.data.repository.default_branch_ref.target.history.total_count
+        end
+      end
+
+      class LastWeekCommitsStat < BaseCommitCount
```
kbarrette: Out of curiosity, why classes with a single method in them rather than a bunch of methods?

mikeyoung85: I think it was initially in the way I set up the method that took queries and assigned them to stat classes. That was changed before merge to be a little more simple. I would like to run back through the Github pieces tomorrow and get those queries and stat classes cleaned up.

PR #2399, opened 2019-07-08
Support Go redirects in bulk API

kbarrette:
Two changes here:
1. Serve project data with the name that was used to request it
2. Handle Go redirects in the bulk status API

Think I should centralize some of this Go handling somehow?

### DIFF
--- a/app/controllers/api/status_controller.rb
+++ b/app/controllers/api/status_controller.rb
@@ -2,30 +2,33 @@ class Api::StatusController < Api::ApplicationController
   before_action :require_api_key
 
   def check
-    if params[:projects].any?
-      # Try to get all the projects passed in.
-      @projects = params[:projects]
-        .group_by { |project| project[:platform] }
-        .flat_map(&method(:find_projects))
-        .compact
-    else
-      @projects = []
-    end
-    render json: @projects, each_serializer: ProjectStatusSerializer, show_score: params[:score], show_stats: internal_api_key?
+    render(
+      json: projects.values,
+      each_serializer: ProjectStatusSerializer,
+      show_score: params[:score],
+      show_stats: internal_api_key?,
+      project_names: project_names
+    )
   end
 
   private
 
-  def find_projects((platform, projects))
-    projects.each_slice(1000).flat_map do |slice|
-      project_find_names = slice.flat_map { |project| project_names(project, platform) }.map(&:downcase)
-      Project.visible.platform(platform).where('lower(platform)=? AND lower(name) in (?)', platform.downcase, project_find_names).includes(:repository, :versions, :repository_maintenance_stats)
-    end
+  def projects
+    @projects ||= params[:projects]
+      .group_by { |project| project[:platform] }
+      .map(&method(:platform_projects))
+      .reduce({}, :merge)
   end
 
-  def project_names(project, platform)
-    platform_class = PackageManager::Base.find(platform)
-    return PackageManager::Base.project_find_names(project[:name]) if platform_class.nil?
-    platform_class.project_find_names(project[:name])
+  def platform_projects((platform, group))
+    ProjectStatusQuery
+      .new(platform, group.map { |p| p[:name] })
+      .projects_by_name
+  end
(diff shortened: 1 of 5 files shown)


app/controllers/api/status_controller.rb
```diff
@@ -11,21 +11,55 @@ def check
...
+        .includes(:repository, :versions, :repository_maintenance_stats)
     end
+
+    if platform.downcase == "go"
+      missing = params[:projects]
+        .select { |project| found.none? { |p| p.name.downcase == project[:name].downcase } }
+
+      missing.each do |project|
+        resolved_name = PackageManager::Go.resolved_name(project[:name])
+        resolved_project = Project.find_by(platform: platform, name: resolved_name)
```
katzj: Is it worth making it so this is a single query? Would need to do the name in list as above and keep a map of resolved to original names but I think the reduction in database queries is worth the complexity.

app/controllers/api/status_controller.rb
```diff
@@ -11,21 +11,55 @@ def check
...
+        .where('lower(platform)=? AND lower(name) in (?)', platform.downcase, project_find_names)
+        .includes(:repository, :versions, :repository_maintenance_stats)
     end
+
+    if platform.downcase == "go"
+      missing = params[:projects]
+        .select { |project| found.none? { |p| p.name.downcase == project[:name].downcase } }
+
+      missing.each do |project|
+        resolved_name = PackageManager::Go.resolved_name(project[:name])
```
mikeyoung85: Is there any way to batch up these requests to resolve project names?

kbarrette: Not that I can think of - we have to fetch each URL

app/controllers/api/status_controller.rb
kbarrette: Can do

kbarrette:
@katzj @mikeyoung85 pulled out a query object to make this a bit cleaner and reduce the number of queries it makes

kbarrette:
Also note that I nuked the `.platform` scope on the `Project` query, since it seemed redundant since we're specifying platform in the `where` clause anyway.

app/controllers/api/status_controller.rb
```diff
@@ -2,30 +2,41 @@ class Api::StatusController < Api::ApplicationController
...
+      each_serializer: ProjectStatusSerializer,
+      show_score: params[:score],
+      show_stats: internal_api_key?,
+      project_names: project_names
+    )
   end
 
   private
 
+<<<<<<< HEAD
```
mikeyoung85: Some merge issues here to cleanup.

app/queries/project_status_query.rb
```diff
@@ -0,0 +1,51 @@
...
+  def missing_projects
+    return {} unless @platform.downcase == "go"
+
+    fetch_projects(
+      @requested_project_names.reject(&projects.method(:key?)),
+      &PackageManager::Go.method(:resolved_name)
+    )
+  end
+
+  def fetch_projects(requested_project_names)
```
mikeyoung85: Should the block be declared in the method signature? I was confused what was going on with the Array yield portion and why it was needed until I saw the calls to this method.

app/controllers/api/status_controller.rb
kbarrette: derp.

app/queries/project_status_query.rb
kbarrette: The block argument is implicit and idiomatic. It IS possible to specify an argument and `.call` it, but historically that is quite a bit slower (not sure if that has changed in newer Ruby?)

I'd love to make this whole `resolved_project_names` thing easier to read (I've already rebuilt it a few times) - lemme try one more!

kbarrette: @mikeyoung85 maybe more readable now?

mikeyoung85: Seems better to me. That slight separation makes it easier to spot the block.

PR #2400, opened 2019-07-09
Cleanup GitHub Stats

mikeyoung85:
I wanted to get the Github stats fixed up a bit. 

In this PR:
* moved query and stat classes into `Github` module
* removed some unused or duplicate stats
* used `repository_host` `download_issues` and `download_pull_requests` and use the database to gather those stats
* redid specs with a smaller repository to cut down on amount of data in VCR cassettes


app/models/repository_host/github.rb
```diff
@@ -231,56 +231,42 @@ def gather_maintenance_stats
         return []
       end
 
-      client = AuthToken.v4_client
+      exists = !Github.fetch_repo(repository.full_name).nil?
```
mikeyoung85: The most important changes are made in this method.

app/models/repository_host/github.rb
```diff
@@ -231,56 +231,42 @@ def gather_maintenance_stats
...
-      end
-
-      result = MaintenanceStats::Queries::RepoReleasesQuery.new(client).query( params: {owner: repository.owner_name, repo_name: repository.project_name, end_date: now - 1.year} )
+      result = MaintenanceStats::Queries::Github::RepoReleasesQuery.new(v4_client).query( params: {owner: repository.owner_name, repo_name: repository.project_name, end_date: now - 1.year} )
       unless check_for_v4_error_response(result)
         metrics << MaintenanceStats::Stats::Github::ReleaseStats.new(result).get_stats
       end
 
-      result = MaintenanceStats::Queries::CommitCountQuery.new(client).query(params: {owner: repository.owner_name, repo_name: repository.project_name, start_date: (now - 1.week).iso8601} )
+      result = MaintenanceStats::Queries::Github::CommitCountQuery.new(v4_client).query(params: {owner: repository.owner_name, repo_name: repository.project_name, start_date: (now - 1.week).iso8601} )
```
mikeyoung85: I would also like to get this down to one GraphQL query, but have not had a chance to try and figure it out yet.

app/models/maintenance_stats/queries/github/commit_count_query.rb
```diff
@@ -0,0 +1,52 @@
+module MaintenanceStats
+  module Queries
+    module Github
+      class CommitCountQuery < BaseQuery
+        CommitCountsQuery = Rails.application.config.graphql.client.parse <<-'GRAPHQL'
```
kbarrette: This looks like a class - if it's not, it should be all caps, snake case

kbarrette: Also, no need to quote the heredoc name

app/models/maintenance_stats/queries/github/commit_count_query.rb
```diff
@@ -0,0 +1,52 @@
...
+                    }
+                  }
+                }
+                name
+              }
+            }
+          }
+        GRAPHQL
+
+        @@valid_params = [:owner, :name, :start_date]
```
kbarrette: You probably don't want class variables here. Looks like these can be constants anyway?

app/models/maintenance_stats/queries/github/releases_query.rb
```diff
@@ -0,0 +1,81 @@
...
+            end
+          end
+
+          @releases
+        end
+
+        private
+
+        def check_for_errors(result)
+          return result.data.nil? || result.errors.any? || result.data.errors.any?
```
kbarrette: No need for `return`

app/models/maintenance_stats/queries/github/repository_contributor_stats_query.rb
```diff
@@ -0,0 +1,25 @@
...
+        def self.client_type
+          :v3
+        end
+        def query(params: {}, count: 0)
+          raise Octokit::Error.new "Didn't get a response after #{@@retry_amount} attempts" if params[:count].present? && params[:count] >= @@retry_amount
+          validate_params(params)
+
+          resp = @client.contributor_stats(params[:full_name])
+          if @client.last_response.status == 202
+            sleep 1
```
kbarrette: 4 seconds of total possible sleep time sounds like a lot, although I'm not sure what this is waiting on?

app/models/maintenance_stats/stats/github/db_issues_stats.rb
```diff
@@ -1,59 +1,59 @@
...
-          one_year_open_issues: open_issues_count,
-          one_year_closed_issues: closed_issues_count,
-          one_year_total_issues: issues.length,
-          one_year_issue_closure_rate: issue_closure_rate,
-          one_year_open_pull_requests: open_pull_request_count,
-          one_year_closed_pull_requests: closed_pull_request_count,
-          one_year_total_pull_requests: pull_requests.length,
-          one_year_pull_request_closure_rate: pull_request_closure_rate,
-          issues_stats_truncated: @results[:truncated]
+            one_year_open_issues: open_issues_count,
```
kbarrette: 🎉

app/models/repository_host/github.rb
kbarrette: `.present?` might be more meaningful here

spec/models/maintenance_stats/github/release_stats_spec.rb
```diff
@@ -31,18 +31,18 @@
...
-        expect(results[:last_release_date]).to eql "2019-01-18T21:32:44Z"
+        expect(results[:last_release_date]).to eql "2019-01-05T15:47:46Z"
     end
   end
   
   context "repository with no commits" do
     let(:repository) { create(:repository, full_name: 'buddhamagnet/heidigoodchild') }
     let(:query_results) do
-        VCR.use_cassette('github/empty_repository', :match_requests_on => [:body]) do
+        VCR.use_cassette('github/empty_repository', :match_requests_on => [:method, :uri, :body, :query]) do
```
kbarrette: Is there anything special that needs to be done to re-record these cassettes? If so, now might be good time to document it for posterity.

mikeyoung85: Where would be a good spot for information like that?

app/models/maintenance_stats/queries/github/repository_contributor_stats_query.rb
mikeyoung85: This request to GitHub will often return a `202` instead of the actual data to let GitHub process the information needed for the request. This is there to retry the call after a short wait, not ideal. It would be better if this was handled in the Octokit client.

kbarrette: Maybe use a simple increasing backoff?

spec/models/maintenance_stats/github/release_stats_spec.rb
kbarrette: In the past I've just plopped it in comments near the VCR calls

app/models/maintenance_stats/queries/github/releases_query.rb
```diff
@@ -0,0 +1,81 @@
...
+            page_info = { has_next_page: has_next_page, cursor: cursor}
+
+            if @releases.nil?
+              @releases = result
+              next
+            end
+            
+            result.data.repository&.releases.nodes.each do |release|
+              publish_date = DateTime.parse(release.published_at)
+              if publish_date > end_date
```
kbarrette: Is this backwards? I feel like it should be `publish_date < end_date`

Also, should it be `<=` so `end_date` is inclusive?

app/models/maintenance_stats/queries/github/releases_query.rb
```diff
@@ -0,0 +1,81 @@
...
+                @releases.data.repository.releases.nodes << release
+              else
+                has_next_page = false
```
kbarrette: Should this be modifying `page_info`?

app/models/maintenance_stats/queries/github/releases_query.rb
```diff
@@ -0,0 +1,81 @@
...
+
+        def query(params: {})
+          validate_params(params)
+
+          end_date = params[:end_date]
+          params = params.slice!(:end_date)
+
+          @releases = nil
+
+          page_info = { has_next_page: true, cursor: nil}
```
kbarrette: I'm wondering if this should just be two variables: `has_next_page` and `cursor` - looks like `has_next_page` is used only to control the loop, and `cursor` is used just in the query params?

app/models/maintenance_stats/queries/github/releases_query.rb
```diff
@@ -0,0 +1,81 @@
...
+          :v4
+        end
+
+        def query(params: {})
+          validate_params(params)
+
+          end_date = params[:end_date]
+          params = params.slice!(:end_date)
+
+          @releases = nil
```
kbarrette: Looks like this should just be a local var?

app/models/maintenance_stats/queries/github/releases_query.rb
```diff
@@ -0,0 +1,81 @@
...
+            has_next_page = result.data.repository.releases.page_info.has_next_page
+            cursor = result.data.repository.releases.page_info.end_cursor
+            page_info = { has_next_page: has_next_page, cursor: cursor}
+
+            if @releases.nil?
+              @releases = result
+              next
+            end
+            
+            result.data.repository&.releases.nodes.each do |release|
```
kbarrette: `&.releases` could be nil, causing `.nodes` to bomb

app/models/maintenance_stats/queries/github/releases_query.rb
```diff
@@ -0,0 +1,81 @@
...
+
+          @releases = nil
+
+          page_info = { has_next_page: true, cursor: nil}
+          while(page_info[:has_next_page])
+            params[:cursor] = page_info[:cursor] if page_info[:cursor]
+            result = @client.query(RELEASES_QUERY, variables: params)
+
+            if check_for_errors(result)
+              @releases = result if @releases.nil?
```
kbarrette: Is `return releases || result if check_for_errors(result)` too much?

app/models/maintenance_stats/queries/github/releases_query.rb
mikeyoung85: I'll rename these variables. The releases are going from newest to oldest and `end_date` is the earliest date we would care about. Like get all releases from the last year, `end_date` would be one year ago. So if the current release's date is after `end_date` then add it to the list and keep going to the next release.

app/models/maintenance_stats/queries/github/releases_query.rb
mikeyoung85: `page_info` is a local variable in this loop and reinitialized after each api request.

app/models/maintenance_stats/queries/github/releases_query.rb
mikeyoung85: Should `repository`, `releases`, and `nodes` all have nil protection?

app/models/maintenance_stats/queries/github/releases_query.rb
mikeyoung85: I think that's essentially the same statement.

app/models/maintenance_stats/queries/github/releases_query.rb
mikeyoung85: I'll separate them.

app/models/maintenance_stats/queries/github/releases_query.rb
kbarrette: Ooh, makes sense. The name was confusing, for sure.

app/models/maintenance_stats/queries/github/releases_query.rb
kbarrette: Yeah, since `&.` returns `nil` if the object you're calling on is `nil`, anything after it might be called on `nil`

app/models/maintenance_stats/queries/github/releases_query.rb
kbarrette: Yeah, I think the assignment of the instance variable `@releases` was making me question it

app/models/maintenance_stats/stats/github/release_stats.rb
```diff
@@ -8,11 +8,13 @@ def initialize(dataset)
         end
 
         def get_stats
-          last_week_releases = @results.data.repository.releases.nodes.select {|release| release.published_at > @now - 7}.count
-          last_month_releases = @results.data.repository.releases.nodes.select {|release| release.published_at > @now - 30}.count
-          last_two_month_releases = @results.data.repository.releases.nodes.select {|release| release.published_at > @now - 60}.count
-          last_year_releases = @results.data.repository.releases.nodes.select {|release| release.published_at > @now - 365}.count
-          {
+          return {} if @results.nil?
```
mikeyoung85: This is a slight change. Before we were sending back `0` releases, but our original stance was to return nothing for a stat if we didn't have the data for it. I think this is more correct.

app/models/maintenance_stats/stats/github/release_stats.rb
```diff
@@ -24,7 +26,7 @@ def get_stats
         private
 
         def last_release_date
-          @results.data.repository.releases.nodes.first&.published_at 
+          @results&.map { |node| DateTime.parse(node.published_at) }.first&.strftime('%FT%TZ')
```
mikeyoung85: The `strftime` is to keep the publish date in the same format as what it was previously.

app/models/maintenance_stats/queries/github/releases_query.rb
```diff
@@ -0,0 +1,77 @@
...
+        REQUIRED_PARAMS = [:owner, :repo_name]
+
+        def self.client_type
+          :v4
+        end
+
+        def query(params: {})
+          validate_params(params)
+
+          end_date = params[:end_date]
```
kbarrette: `end_date = params.delete(:end_date)`

app/models/maintenance_stats/queries/github/releases_query.rb
```diff
@@ -0,0 +1,77 @@
...
+            result = @client.query(RELEASES_QUERY, variables: params)
+
+            # send back the releases we have so far if we encounter an error
+            return releases if check_for_errors(result)
+
+            has_next_page = result.data.repository.releases.page_info.has_next_page
+            cursor = result.data.repository.releases.page_info.end_cursor
+
+            # initalize releases if we have not started gathering data
+            releases ||= []
```
kbarrette: On line 38 `releases = []` would make this unnecessary, I think? Unless it's important to return nil?

app/models/maintenance_stats/queries/github/repository_contributor_stats_query.rb
```diff
@@ -0,0 +1,27 @@
...
+          :v3
+        end
+
+        def query(params: {}, count: 0)
+          raise Octokit::Error.new "Didn't get a response after #{RETRY_AMOUNT} attempts" if params[:count].present? && params[:count] >= RETRY_AMOUNT
+          validate_params(params)
+
+          resp = @client.contributor_stats(params[:full_name])
+          if @client.last_response.status == 202
+            sleep (count + 1) * 0.5
```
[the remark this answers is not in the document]
kbarrette: Thoughts on ditching the `+ 1` to retry after 500ms the first time?

app/models/maintenance_stats/stats/github/contributor_count_stats.rb
```diff
@@ -14,13 +14,15 @@ def get_stats
...
-          end.sum
+          @results.select{ |contributor| contributed?(contributor, weeks_ago) }.size
+        end
+
+        def contributed?(contributor, weeks_ago)
+          # todo: verify weeks are actually within the time period?
+          # this assumes the latest week is the last item in the weeks array for this contributor
+          # count up the weeks counting backwards from the end of the array
+          # return true if there is a commit in any of those weeks
+          contributor.weeks[-1*weeks_ago..contributor.weeks.length].sum(&:c) > 0
```
kbarrette: You can use `-1` to mean "end of array" here rather than `.length`

app/models/maintenance_stats/stats/github/release_stats.rb
kbarrette: `.iso8601` is there if you'd rather use that than `strftime`

app/models/maintenance_stats/queries/github/releases_query.rb
mikeyoung85: I'm using `nil` to indicate no data was returned from the API. At this line we have a response from GitHub so I want to initialize to an empty array unless this is after the first page of information.

app/models/maintenance_stats/queries/github/repository_contributor_stats_query.rb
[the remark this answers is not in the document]
mikeyoung85: `count` would be `0` the first time through that's why I added the +1. I can redo `count` to start at 1 to remove this.

app/models/maintenance_stats/stats/github/release_stats.rb
mikeyoung85: `iso8601` was returning something slightly different than what GitHub is sending back. I'm trying to keep the formatting the same as the stats that are already in the database.

app/models/maintenance_stats/queries/github/repository_contributor_stats_query.rb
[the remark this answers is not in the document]
kbarrette: Whoops, that's fine!

PR #2401, opened 2019-07-09
Exclude Maven Names Found From latest That Are Invalid

mikeyoung85:
This would prevent packages being added to Maven from the maven-repository.com/artifacts/latest scan if they do not have a valid name of `{group_id}:{artifact_name}`.

### DIFF
--- a/app/models/package_manager/maven.rb
+++ b/app/models/package_manager/maven.rb
@@ -40,9 +40,7 @@ module PackageManager
 
       (1..num).to_a.reverse.each do |number|
         page = get_html "https://maven-repository.com/artifact/latest?page=#{number}"
-        page.css('tr')[1..-1].each do |tr|
-          REDIS.sadd 'maven-names', tr.css('td')[0..1].map(&:text).join(':')
-        end
+        parse_names(page).each { |name| REDIS.sadd 'maven-names', name }
         REDIS.set 'maven-page', number
       end
     end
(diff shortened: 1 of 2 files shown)


mikeyoung85:
I don't have a way to really test this all works 100% correctly since maven-repository.com is just returning 500 errors at the moment.

app/models/package_manager/maven.rb
```diff
@@ -210,11 +206,24 @@ def self.latest_version(versions, name)
       end
     end
 
+    def self.parse_names(page)
+      # parse the names from maven-repository.com/artifact/latest pages
```
katzj: Is it worth having a test with some good and bad names to keep this from breaking?

mikeyoung85: Probably. I'll need to look into how to construct the right object for the page parsing. I'll add a spec for the MavenUrl class in there too to test the `legal_name?` method I added.

app/models/package_manager/maven.rb
```diff
@@ -210,11 +206,24 @@ def self.latest_version(versions, name)
...
+      names = page.css('tr')[1..-1]&.map do |tr|
+        tr.css('td')[0..1]&.map(&:text).join(':')
```
kbarrette: `&.map` might return nil, and `.join` will bomb

app/models/package_manager/maven.rb
```diff
@@ -210,11 +206,24 @@ def self.latest_version(versions, name)
...
+      names.select(&MavenUrl.method(:legal_name?)).uniq
+    end
+
     class MavenUrl
       def self.from_name(name)
         new(*name.split(':', 2))
       end
 
+      def self.legal_name?(name)
+        name.split(':').size == 2
```
kbarrette: Looks like `name` might be nil, so this will bomb - perhaps `compact` the `names` array?

PR #2410, opened 2019-07-23
Increase tree depth

kbarrette:
Two things I wanted to change here:
1. Increase max tree depth to 15 (which lets eg: `npm/gulp` get to the leaves of its tree)
2. Remove the `["MORE"]` magic string when truncating

To get there, I did some refactoring.

The "Add TreeResolver tests" commit adds tests for `TreeResolver` that pass before any changes to that class.

Probably best to review this commit-by-commit!

### DIFF
--- a/app/models/tree_resolver.rb
+++ b/app/models/tree_resolver.rb
@@ -1,5 +1,5 @@
 class TreeResolver
-  MAX_TREE_DEPTH = 10
+  MAX_TREE_DEPTH = 15
 
   def initialize(version, kind, date = nil)
     @version = version
@@ -57,7 +57,7 @@ class TreeResolver
     append_license_names(version)
     should_fetch = append_project_name(dependency)
 
-    return ["MORE"] if index > MAX_TREE_DEPTH
+    return if index > MAX_TREE_DEPTH
 
     dependencies = should_fetch ? fetch_dependencies(version, kind) : []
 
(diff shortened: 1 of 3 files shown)


app/models/tree_resolver.rb
```diff
@@ -25,53 +22,78 @@ def enqueue_tree_resolution
...
+  end
+
+  def license_names
+    @license_names.to_a
+  end
+
   private
 
-  def generate_dependency_tree
+  def generate_dependency_tree(_key)
```
jsonperl: _key ?

kbarrette: Rails cache `fetch` passes the key to the block

jsonperl: Ahhh, gotcha

kbarrette:
Hard to benchmark, but I can try something. Performance should be about the same, I think I only saved a few queries.

app/models/tree_resolver.rb
```diff
@@ -25,53 +22,78 @@ def enqueue_tree_resolution
...
-          end
-        end
-      }
-    end
+    return unless version
+
+    append_license_names(version)
+    should_fetch = append_project_name(dependency)
+
+    return if index > MAX_TREE_DEPTH
```
katzj: Should we include anything so that we know that the max depth has been exceeded and that the tree is thus truncated?

kbarrette: We talked about it, but I figured YAGNI applied here. I can add in a flag if you think it's important.

katzj: A log message is probably a fine super lightweight start. Then we can see if we're hitting it often

kbarrette: Done

> Max tree depth 15 reached when loading dependencies for Rubygems/16 for parent Rubygems/root

PR #2415, opened 2019-08-06
Get Commit Counts in one Query

mikeyoung85:
Some more cleanup of GitHub stats that I had stashed. This will condense the commit count API requests from 4 into 1 query.


app/models/maintenance_stats/stats/github/contributor_count_stats.rb
```diff
@@ -22,7 +22,7 @@ def contributed?(contributor, weeks_ago)
           # this assumes the latest week is the last item in the weeks array for this contributor
           # count up the weeks counting backwards from the end of the array
           # return true if there is a commit in any of those weeks
-          contributor.weeks[-1*weeks_ago..contributor.weeks.length].sum(&:c) > 0
+          contributor.weeks[-1*weeks_ago..-1].sum(&:c) > 0
```
mikeyoung85: This was leftover PR feedback that I think was never merged.

app/models/maintenance_stats/queries/github/releases_query.rb
```diff
@@ -32,8 +32,7 @@ def self.client_type
         def query(params: {})
           validate_params(params)
 
-          end_date = params[:end_date]
-          params = params.slice!(:end_date)
+          end_date = params.delete(:end_date)
```
mikeyoung85: This was leftover PR feedback that I think was never merged.

app/models/maintenance_stats/queries/github/commit_count_query.rb
```diff
@@ -29,7 +38,18 @@ def self.client_type
         def query(params: {})
           validate_params(params)
 
-          @client.query(COMMIT_COUNTS_QUERY, variables: params)
+          # generate the four dates needed
+          date_params = Hash.new { |k,v| k[v] = v.iso8601 }
```
kbarrette: The arguments to the block are the hash and the new key - I think you're going to end up with `:one_week.iso8601` here?

mikeyoung85: Yeah I don't think this actually works the way I thought it would. I'm going to remove since it isn't even being used.

PR #2418, opened 2019-08-13
Improve package name casing

kbarrette:
Not sure this is done yet, but I wanted to get some opinions.

In particular, I'm not sure if the last commit (project status) should be there - perhaps we want to leave that endpoint as it is? Thoughts?

### DIFF
--- a/app/controllers/api/projects_controller.rb
+++ b/app/controllers/api/projects_controller.rb
@@ -74,7 +74,7 @@ class Api::ProjectsController < Api::ApplicationController
                 raise ActionController::BadRequest.new("Unsupported subset")
               end
 
-    project = Project.find_with_includes!(platform, name, includes)
+    project = Project.find_best!(platform, name, includes)
     version = project.find_version!(version_name)
 
     project_json = serializer.new(project).as_json
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -73,31 +73,21 @@ class ApplicationController < ActionController::Base
   end
 
   def find_project
-    @project = Project.find_with_includes!(params[:platform], params[:name], [:repository, :versions])
+    @project = Project.find_best!(params[:platform], params[:name], [:repository, :versions])
     @color = @project.color
-  rescue ActiveRecord::RecordNotFound
-    raise if params[:name].blank? || params[:platform]&.downcase != "go"
 
-    resolved_name = PackageManager::Go.resolved_name(params[:name])
-    if resolved_name != params[:name] && Project.known?(params[:platform], resolved_name)
+    if @project.name != params[:name]
       redirect_to(
-        # Unescape since url_for automatically escapes our already-escaped resolved_name
+        # Unescape since url_for automatically escapes our already-escaped project name
         URI.unescape(
           url_for(
             params
               .to_unsafe_h
-              .merge({ "name" => CGI.escape(resolved_name) })
+              .merge({ "name" => CGI.escape(@project.name) })
           )
         )
       )
-    else
-      raise
     end
-  end
-
-  def find_project_lite
-    @project = Project.visible.platform(params[:platform]).where(name: params[:name]).first
-    raise ActiveRecord::RecordNotFound if @project.nil?
   end
 
   def current_platforms
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -1,11 +1,25 @@
 class ProjectsController < ApplicationController
   before_action :ensure_logged_in, only: [:your_dependent_repos, :mute, :unmute,
                                           :unsubscribe, :sync]
-  before_action :find_project, only: [:show, :sourcerank, :about, :dependents,
-                                      :dependent_repos, :your_dependent_repos,
-                                      :versions, :tags, :mute, :unmute, :unsubscribe,
-                                      :sync, :score, :refresh_stats]
-  before_action :find_project_lite, only: [:top_dependent_repos, :top_dependent_projects]
+  before_action :find_project, only: %i[
+    about
+    dependencies
+    dependent_repos
+    dependents
+    mute
+    refresh_stats
+    score
+    show
+    sourcerank
+    sync
+    tags
+    top_dependent_projects
+    top_dependent_repos
+    unmute
+    unsubscribe
+    versions
+    your_dependent_repos
+  ]
 
   def index
     if current_user
@@ -151,7 +165,6 @@ class ProjectsController < ApplicationController
   end
 
   def dependencies
-    find_project_lite
     find_version
     render layout: false
   end
--- a/app/models/package_manager/go.rb
+++ b/app/models/package_manager/go.rb
@@ -50,9 +50,9 @@ module PackageManager
     end
 
     # https://golang.org/cmd/go/#hdr-Import_path_syntax
-    def self.resolved_name(name)
-      return name if name.start_with?(*KNOWN_HOSTS)
-      return name if KNOWN_VCS.any?(&name.method(:include?))
+    def self.project_find_names(name)
+      return [name] if name.start_with?(*KNOWN_HOSTS)
+      return [name] if KNOWN_VCS.any?(&name.method(:include?))
 
       go_import = get_html('https://' + name + '?go-get=1')
         .xpath('//meta[@name="go-import"]')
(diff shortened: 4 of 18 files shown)


app/models/project.rb
```diff
@@ -513,30 +513,37 @@ def download_registry_users
...
-    if project.nil?
-      query = self.visible.lower_platform(platform).lower_name(name)
-      query = query.includes(*includes) unless includes.empty?
-      project = query.first
-    end
-    raise ActiveRecord::RecordNotFound if project.nil?
-    raise ActiveRecord::RecordNotFound if project.status == 'Hidden'
-    project
+
+    visible.lower_platform(platform).lower_name(name).first ||
```
mikeyoung85: The `includes` should be added to these queries

app/models/project.rb
```diff
@@ -513,30 +513,37 @@ def download_registry_users
...
+      visible.platform(platform).where(name: project_find_names).first!
```
mikeyoung85: These seem backwards from the original logic. I think we should only attempt the lowercase query if we don't find a project with `project_find_names`. It's also a bit odd that platform is downcased sometimes but not always.

app/models/repository_dependency.rb
```diff
@@ -23,9 +23,7 @@ class RepositoryDependency < ApplicationRecord
   delegate :filepath, to: :manifest
 
   def find_project_id
-    project_id = Project.platform(platform).where(name: project_name.try(:strip)).limit(1).pluck(:id).first
-    return project_id if project_id
-    Project.lower_platform(platform).lower_name(project_name.try(:strip)).limit(1).pluck(:id).first
+    Project.find_best(platform, project_name.strip)&.id
```
mikeyoung85: Is `project_name` potentially `nil` here? I just noticed the `try` for `strip` in the existing code.

app/models/package_manager/pypi.rb
```diff
@@ -83,7 +83,9 @@ def self.licenses(project)
     end
 
     def self.project_find_names(project_name)
-      [project_name, project_name.downcase].flat_map {|name| [name, name.gsub('-', '_'), name.gsub('_', '-')]}.uniq
+      [project_name, project_name.downcase]
```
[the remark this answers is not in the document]
kbarrette: Let me grep about that...

app/models/project.rb
kbarrette: I think I only want to do `project_find_names` if we don't find anything with the lowercase query - it's a potentially expensive operation (eg: making a network call for Go)

app/models/repository_dependency.rb
kbarrette: Will check

kbarrette:
@mikeyoung85 updated - still not sure what I want to do about the project status api, but I'll figure something out

kbarrette:
@mikeyoung85 I added `normalized_name` to the project status output - I think that might be the sane way to go here

app/models/package_manager/go.rb
```diff
@@ -50,9 +50,9 @@ def self.mapping(project)
     end
 
     # https://golang.org/cmd/go/#hdr-Import_path_syntax
-    def self.resolved_name(name)
-      return name if name.start_with?(*KNOWN_HOSTS)
-      return name if KNOWN_VCS.any?(&name.method(:include?))
+    def self.project_find_names(name)
+      return [name] if name.start_with?(*KNOWN_HOSTS)
```
mikeyoung85: You could condense this down to one line although it would be quite dense 😄

app/models/package_manager/pypi.rb
```diff
@@ -83,7 +83,11 @@ def self.licenses(project)
     end
 
     def self.project_find_names(project_name)
-      [project_name, project_name.downcase].flat_map {|name| [name, name.gsub('-', '_'), name.gsub('_', '-')]}.uniq
+      [
+        project_name,
+        project_name.gsub('-', '_'),
+        project_name.gsub('_', '-'),
```
mikeyoung85: 👍

app/serializers/project_status_serializer.rb
```diff
@@ -17,4 +24,8 @@ def show_stats?
   def name
     instance_options[:project_names][object.name]
   end
+
+  def normalized_name
+    object.name
```
mikeyoung85: Should this be reversed with `name` and maybe send back `projecrt_names`. The word normalized doesn't make sense to me here.

app/controllers/application_controller.rb
```diff
@@ -73,31 +73,28 @@ def find_platform(param = :id)
...
-    if resolved_name != params[:name] && Project.known?(params[:platform], resolved_name)
-      redirect_to(
-        # Unescape since url_for automatically escapes our already-escaped resolved_name
-        URI.unescape(
-          url_for(
-            params
-              .to_unsafe_h
-              .merge({ "name" => CGI.escape(resolved_name) })
-          )
+    raise if params[:name].blank?
```
mikeyoung85: I think we should just raise this right away.

app/controllers/application_controller.rb
```diff
@@ -73,31 +73,28 @@ def find_platform(param = :id)
...
+
+    project = Project.find_fuzzy!(params[:platform], params[:name])
+    raise if project.name == params[:name]
```
mikeyoung85: I know this was how it was originally written, but I think we could use `Project.find_best` in here and if the name does not equal `params[:name]` do the redirect. I think would make this look simpler.

kbarrette: Good thinking, lemme look at that

app/serializers/project_status_serializer.rb
kbarrette: Yeah, I didn't want to disturb the existing behavior, and used "normalized" because of `normalized_licenses` - I'll change wording

app/models/package_manager/go.rb
kbarrette: This could all be much cleaner if `PackageManager` classes were instantiated instead of all class methods :/

app/controllers/projects_controller.rb
```diff
@@ -1,11 +1,25 @@
 class ProjectsController < ApplicationController
   before_action :ensure_logged_in, only: [:your_dependent_repos, :mute, :unmute,
                                           :unsubscribe, :sync]
-  before_action :find_project, only: [:show, :sourcerank, :about, :dependents,
-                                      :dependent_repos, :your_dependent_repos,
-                                      :versions, :tags, :mute, :unmute, :unsubscribe,
-                                      :sync, :score, :refresh_stats]
-  before_action :find_project_lite, only: [:top_dependent_repos, :top_dependent_projects]
```
mikeyoung85: Is the performance of these two actions going to be a lot longer without the "lite" version of the query?

app/models/project.rb
```diff
@@ -345,6 +345,47 @@ def self.format_license(license)
     Spdx.find(license).try(:id) || license
   end
 
+  def self.find_best(*args)
```
mikeyoung85: Probably more of a styling thing, but it's strange seeing open ended arguments here but not in `find_best!`.

app/controllers/api/projects_controller.rb
```diff
@@ -74,7 +74,8 @@ def find_project_as_json_with_dependencies!(platform, name, version_name, subset
                 raise ActionController::BadRequest.new("Unsupported subset")
               end
 
-    project = Project.find_with_includes!(platform, name, includes)
+    project = Project.find_best(platform, name, includes)
+    raise ActiveRecord::RecordNotFound unless project
```
mikeyoung85: Why not just use `find_best!` here instead of raising on `nil`

app/controllers/projects_controller.rb
kbarrette: It's really just the includes - I can't imagine there was ever any real need for `lite`

kbarrette: I'll check the git logs though...

app/models/project.rb
kbarrette: I did it here because it's just passing all args through to `find_best!`

app/controllers/api/projects_controller.rb
kbarrette: Yep! Just swapped it!

app/controllers/projects_controller.rb
kbarrette: I can't find any reasoning

PR #2426, opened 2019-09-03
Block an ip that's scraping at an incredibly aggressive rate

katzj:


config/initializers/rack-attack.rb
```diff
@@ -17,3 +17,6 @@
 Rack::Attack.throttle('scrapers', :limit => 30, :period => 5.minutes) do |req|
   req.remote_ip if req.user_agent && req.user_agent.match(/Scrapy.*/)
```
jsonperl: Doesn't this mean we're pretty much blocking nobody?

jsonperl: *throttling

PR #2434, opened 2019-09-20
Adding check status url for conda so that we can see if a package exists in the anaconda packages, or if it should be removed from the site

tyrelsouza:
…sts in the anaconda packages, or if it should be removed from the site

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -491,6 +491,8 @@ class Project < ApplicationRecord
       update_attribute(:status, 'Removed')
     elsif platform.downcase != 'packagist' && [400, 404].include?(response.response_code)
       update_attribute(:status, 'Removed')
+    elsif platform.downcase == "conda" && response.response_code == 404
+      update_attribute(:status, 'Removed')
     elsif removed
       update_attribute(:status, nil)
     end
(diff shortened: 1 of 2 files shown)


tyrelsouza:
Looks like the test at https://github.com/librariesio/libraries.io/blob/8fbdf46865e1400ceda0bcb0746ed4b7c7863ff6/spec/workers/check_repo_status_worker_spec.rb#L14 covers this already, so I opted to not add another test.

app/models/project.rb
```diff
@@ -491,6 +491,8 @@ def check_status(removed = false)
       update_attribute(:status, 'Removed')
     elsif platform.downcase != 'packagist' && [400, 404].include?(response.response_code)
       update_attribute(:status, 'Removed')
+    elsif platform.downcase == "conda" && response.response_code == 404
```
mikeyoung85: I don't think you need this. The if check above it covers this.

tyrelsouza: Which one specifically? those two are for packagist..

mikeyoung85: `elsif platform.downcase != 'packagist' && [400, 404].include?(response.response_code)` covers it

tyrelsouza: OH it's `!=` Damn

tyrelsouza: For those following along: I enabled ligatures in my code editor, so `!=` looked like a wide `≠`, but my brain didn't read the `=\= as a  `!=` because it was so wide.

My bad!

PR #2439, opened 2019-10-02
Follow Redirects to Go Package Github URLs to Find Correct One

mikeyoung85:
Should translate from `github.com/DarthSim/imgproxy` to `https://github.com/imgproxy/imgproxy`


app/models/package_manager/go.rb
```diff
@@ -45,7 +45,8 @@ def self.mapping(project)
         name: project['Package'],
         description: project['Synopsis'],
         homepage: project['ProjectURL'],
-        repository_url: "https://#{project['Package']}"
+        # repository_url: "https://#{project['Package']}"
```
kbarrette: The comment can probably go

PR #2449, opened 2019-10-16
Expose license normalization

kbarrette:
First pass at this - thoughts?


db/migrate/20191016195047_project_licenses_normalized.rb
```diff
@@ -0,0 +1,5 @@
+class ProjectLicensesNormalized < ActiveRecord::Migration[5.2]
+  def change
+    add_column :projects, :license_normalized, :boolean, default: false
```
kbarrette: Not sure about this name - there are a lot of combinations of "license" and "normalize" floating around. I'm open to suggestions.

kbarrette:
@mikeyoung85 updated as suggested, and rebased

PR #2458, opened 2019-11-04
Add Latest Commit Date Stat

mikeyoung85:


app/models/maintenance_stats/stats/bitbucket/commit_count_stats.rb
```diff
@@ -22,6 +23,11 @@ def count_up_commits(dataset, since)
           return unless dataset.present?
           dataset.select {|commit| Date.parse(commit["date"]) >= since}.size
         end
+
+        def latest_commit(dataset)
+          latest = dataset&.sort_by{|commit| Date.parse(commit["date"])}&.reverse&.first
```
kbarrette: `reverse.first` == `last`?

kbarrette: What format is `date` in - it'd be nice to not have to parse it.

mikeyoung85: Will sorting work without Date parsing?

mikeyoung85: Dates are formatted like this: `2017-02-17T16:25:10+00:00`. I think thats the 8601 standard format.

mikeyoung85:
Updated sort to not parse

PR #2471, opened 2019-12-31
Create Endpoint to Begin Watching Projects Maintenance Stats in Bulk

mikeyoung85:
This creates a new endpoint that lets us take in a list of projects to watch for maintenance stats. I've reused the queries and work done in the StatusController to handle looking up the Projects and moved that code to a common controller class.

### DIFF
--- a/app/controllers/api/bulk_project_controller.rb
+++ b/app/controllers/api/bulk_project_controller.rb
@@ -1,0 +1,21 @@
+class Api::BulkProjectController < Api::ApplicationController
+  def project_status_queries
+    @project_status_queries ||= params[:projects]
+      .group_by { |project| project[:platform] }
+      .map { |platform, group| ProjectStatusQuery.new(platform, group.map { |p| p[:name] }) }
+  end
+
+  def projects
+    project_status_queries
+      .map(&:projects_by_name)
+      .flat_map(&:values)
+  end
+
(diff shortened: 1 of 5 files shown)


app/controllers/api/bulk_project_controller.rb
```diff
@@ -0,0 +1,21 @@
+class Api::BulkProjectController < Api::ApplicationController
+  def project_status_queries
+    @project_status_queries ||= params[:projects]
+      .group_by { |project| project[:platform] }
+      .map { |platform, group| ProjectStatusQuery.new(platform, group.map { |p| p[:name] }) }
+  end
+
+  def projects
+    project_status_queries
```
mikeyoung85: Should we be batching these calls in here?

app/controllers/api/maintenance_stats_controller.rb
```diff
@@ -1,13 +1,23 @@
...
   end
 
   def begin_watching
-    @project.update_maintenance_stats_async(priority: :high) unless @project.repository_maintenance_stats.exists?
+    begin_project_watch(@project)
     head :accepted
   end
+
+  def begin_watching_bulk
+    projects.map(&method(:begin_project_watch))
```
kbarrette: Should this be `each`?

PR #2477, opened 2020-01-12
Adds Dependency#version_project_id, and Project#dependent_projects_v2

tiegz:
This denormalizes the `Dependency#version#project` association as `Dependency#version_project`, to speed up queries that do lookups for dependent projects, e.g. `/npm/foo/dependents`

### Query using the existing `@project.dependent_projects`

``` mysql
SELECT  "projects".* FROM "projects"  
  INNER JOIN "versions" ON "projects"."id" = "versions"."project_id" 
  INNER JOIN "dependencies" ON "versions"."id" = "dependencies"."version_id" 
WHERE "dependencies"."project_id" = $1 
  GROUP BY projects.id 
  ORDER BY projects.rank DESC NULLS LAST;
```

### Query using the new `@project.dependent_projects_v2`

``` mysql
SELECT DISTINCT "projects".* FROM "projects"  
  INNER JOIN "dependencies" ON "projects"."id" = "dependencies"."version_project_id" 
WHERE "dependencies"."project_id" = $1 
  ORDER BY projects.rank DESC NULLS LAST;
```

I'm not sure why the existing one uses a `GROUP BY` instead of `DISTINCT`, so please lmk if anyone has background on that.

Hard to say how much this will improve the query until it's deployed bc there are millions of dependency rows on production.

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -76,7 +76,7 @@ class ProjectsController < ApplicationController
   end
 
   def dependents
-    @dependents = @project.dependent_projects_v2.visible.paginate(page: page_number)
+    @dependents = @project.dependent_projects.visible.paginate(page: page_number)
   end
 
   def dependent_repos
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -76,7 +76,7 @@ class ProjectsController < ApplicationController
   end
 
   def dependents
-    @dependents = @project.dependent_projects.visible.paginate(page: page_number)
+    @dependents = @project.dependent_projects_v2.visible.paginate(page: page_number)
   end
 
   def dependent_repos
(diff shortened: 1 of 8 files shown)


app/controllers/projects_controller.rb
```diff
@@ -76,7 +76,7 @@ def about
   end
 
   def dependents
-    @dependents = @project.dependent_projects.visible.paginate(page: page_number)
+    @dependents = @project.dependent_projects_v2.visible.paginate(page: page_number)
```
tiegz: My plan was to just deploy this, and if it helps, replace the other spots with this new method.

app/models/dependency.rb
```diff
@@ -2,6 +2,7 @@ class Dependency < ApplicationRecord
   include DependencyChecks
 
   belongs_to :version
+  belongs_to :version_project, class_name: 'Project'
```
tiegz: Also considered `dependent_project` as a name? But this is more obvious at first glance.

PR #2486, opened 2020-03-04
Update for CVEs

mellisio:


katzj:
This fails to build due to the bundler update

Gemfile
```diff
@@ -1,5 +1,5 @@
 source 'https://rubygems.org'
-ruby '2.6.2'
+ruby '2.6.5'
```
katzj: Also need to update the travis config

PR #2491, opened 2020-03-20
Initial support for licensing per version

mellisio:


app/models/package_manager/maven.rb
```diff
@@ -151,9 +151,11 @@ def self.versions(project)
 
     def self.extract_versions(versions)
       versions["response"]["docs"].map do |version|
+        license_string = licenses(get_pom(version["g"], version["a"], version["v"]))
```
jsonperl: Wow

app/models/package_manager/npm.rb
```diff
@@ -89,10 +89,14 @@ def self.licenses(latest_version)
     end
 
     def self.versions(project)
-      project["versions"].map do |k, _v|
+      # npm license fields are supposed to be SPDX expressions now https://docs.npmjs.com/files/package.json#license
```
katzj: Should this be accepting a `name` argument like all the others?

app/serializers/version_serializer.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class VersionSerializer < ActiveModel::Serializer
-  attributes :number, :published_at
+  attributes :number, :published_at, :spdx_expression
```
katzj: Should we also include `original_license` here?

mellisio: My reason against was that it could hugely impact the size of versions over the wire. Pypi in particular has a lot of raw licenses copied and pasted.

app/models/package_manager/npm.rb
mellisio: yes good catch.

app/serializers/version_serializer.rb
katzj: How do we handle that for project licenses?

mellisio: We don't really but at the moment you're just getting one license field per project call, with this change you'll be getting one per version

katzj: But if we don't return the original license, then we can do nothing for non-spdx licenses. There is a case for some of those that we want to handle (for example, vcr is now licensed under the Hippocratic license which can't be represented by spdx at present)

PR #2506, opened 2020-04-28
Bump Up the Repository Maintenance Stats Sidekiq Priorities a Bit

mikeyoung85:
We are putting a lot more requests in to update repository maintenance stats in the background and that queue has been backed up for a while now. This might help clear up the queue a bit and get through the stats quicker.


katzj:
I think the more impactful thing would be to bump up the number of sidekiq workers... this is just gonna juggle what's behind I think

PR #2509, opened 2020-04-30
Add NONE and NOASSERTION per spdx spec

mellisio:


app/models/version.rb
```diff
@@ -25,10 +25,20 @@ def save_project
...
-      self.spdx_expression = possible_license if Spdx.valid_spdx?(possible_license)
+      self.spdx_expression = handle_string_spdx_expression(possible_license)
+    end
+  end
+
+  def handle_string_spdx_expression(license_string)
+    if license_string == ""
+      "NONE"
+    elsif Spdx.valid_spdx?(license_string)
+      original_license
```
jsonperl, line 39: Shouldn't this return license_string? Maybe make that an early return at the top?
`return license_string if Spdx.valid_spdx?(license_string)`

PR #2519, opened 2020-05-15
Update spdx gem

mellisio:


Gemfile
```diff
@@ -84,7 +84,7 @@ gem "sidekiq-unique-jobs"
 gem "simple-rss"
 gem "simple_form"
 gem "sitemap_generator"
-gem "spdx", git: "https://github.com/librariesio/spdx", tag: "v2.0.6"
+gem "spdx", git: "https://github.com/librariesio/spdx", tag: "v2.0.8"
```
katzj, line 87: Is there a reason we can't use uploaded gems?

jsonperl: I think we got into this pattern because we were lacking ruby-gems creds. Can we store em?

katzj: I'm happy to give you and @mellisio push access to rubygems for the spdx gem :)

PR #2524, opened 2020-05-26
Add a couple suggested indices

mellisio:


db/migrate/20200526151754_add_index_on_issues_and_repository_users.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class AddIndexOnIssuesAndRepositoryUsers < ActiveRecord::Migration[5.2]
+  def change
+    commit_db_transaction
+    add_index :issues, [:last_synced_at], algorithm: :concurrently
```
jsonperl, line 6: Don't you have to disable_ddl_transaction! to run concurrent indexes?

PR #2528, opened 2020-06-05
Go Modules version and dependency support!

tiegz:
This adds support for "Releases" and "Dependencies" for Go Modules, which most Go projects are using now. It switches us from "go-search.org" (which doesn't work anymore) to "pkg.go.dev" (the new Go Module directory) and "proxy.golang.org" (the official Go Module proxy).

Example from k8s:

Screenshot 2020-06-05 13 04 22


app/models/package_manager/go.rb
```diff
@@ -21,32 +23,71 @@ class Go < Base
...
+          &.lines
+          &.map(&:strip)
+          &.reject(&:blank?)
+          &.map do |v|
+            info = get("http://proxy.golang.org/#{go_module}/@v/#{v}.info")
+            {
+              number: info["Version"],
+              published_at: info["Time"].presence && Time.parse(info["Time"])
+            }
+          end
```
tiegz, line 62: This is the only questionable part -- it's a choice between two options:

1) this, where we make a superfast request for each version to get it's actual published_at
2) or, make a single request to the pkg.go.dev page, and get each version with a Date-only-version of its published_at

🤷‍♂️

app/models/package_manager/go.rb
```diff
@@ -21,32 +23,71 @@ class Go < Base
...
+      "https://pkg.go.dev/#{name}#{"@#{version}" if version}?tab=doc"
     end
 
     def self.install_instructions(project, version = nil)
       "go get #{project.name}"
     end
 
     def self.project_names
-      get("http://go-search.org/api?action=packages")
+      # Currently the index only shows the last <=2000 modules from the date given. (https://proxy.golang.org/)
```
katzj, line 38: About how many a day are there usually?

app/models/package_manager/go.rb
```diff
@@ -21,32 +23,71 @@ class Go < Base
...
+      project_window = 1.day.ago.strftime("%FT%TZ")
+      get_raw("https://index.golang.org/index?since=#{project_window}&limit=2000")
+        .lines
+        .map { |line| JSON.parse(line)["Path"] }
     end
 
     def self.project(name)
-      get("http://go-search.org/api?action=package&id=#{name}")
+      if pkg_html = get_html("https://pkg.go.dev/#{name}?tab=doc")
+        go_module, _latest_version = pkg_html.css('a[data-test-id="DetailsHeader-infoLabelModule"]').first&.text&.split('@', 2)
```
katzj, line 47: This seems fragile... how often is `name != go_module`? Because then we could just use the version list from the proxy and figure out latest version ourselves

app/models/package_manager/go.rb
```diff
@@ -21,32 +23,71 @@ class Go < Base
...
+        go_module_html = get_html("https://pkg.go.dev/mod/#{go_module}")
+        # NB this requires a quick request for each version, but we could alternatively fetch from
+        # https://pkg.go.dev/mod/#{go_module}?tab=versions and the '.Versions-commitTime', selector,
+        # but the dates are a combination of date-only or natural language (e.g. '1 day ago')
+        versions = get_raw("http://proxy.golang.org/#{go_module}/@v/list")
```
katzj:
```suggestion
        versions = get_raw("https://proxy.golang.org/#{go_module}/@v/list")
```

app/models/package_manager/go.rb
katzj: I like the approach here of doing a request per version. This is a background job and so would rather get the full and correct info

app/models/package_manager/go.rb
```diff
@@ -21,32 +23,71 @@ class Go < Base
...
     def self.documentation_url(name, version = nil)
-      "http://godoc.org/#{name}"
+      "https://pkg.go.dev/#{name}#{"@#{version}" if version}?tab=doc"
     end
 
     def self.install_instructions(project, version = nil)
       "go get #{project.name}"
     end
 
     def self.project_names
```
katzj, line 37: I wonder if it's worth jumping on https://github.com/golang/go/issues/36785 and joining that discussion? We should at least be following it

tiegz: Good idea, hadn't seen that yet! Just posted some thoughts: https://github.com/golang/go/issues/36785#issuecomment-639794685

app/models/package_manager/go.rb
tiegz: Uh oh, just realized this is actually package-specific (not module-specific), and it's version-specific too. So this list is gonna change a lot. Left a note about that in the Go issue above tho.

app/models/package_manager/go.rb
tiegz: @katzj I just switched this to use the `pkg.go.dev` scraping instead, because I noticed `pkg.go.dev` will list files that aren't technically Go Modules, e.g. https://pkg.go.dev/github.com/BurntSushi/toml. Whereas the Go proxy url fails 😕 :

``` sh
 $ curl http://proxy.golang.org/github.com/BurntSushi/toml/@v/list
bad request: invalid escaped module path "github.com/BurntSushi/toml"
```

Seem ok?

app/models/package_manager/go.rb
tiegz: We're currently indexing both packages+modules, so `name` could be either the package or the parent module.

This part's not to get the latest_version tho, it's just to strip the latest_version from this module name. I might be wrong and it's not there, so I'll circle back in a bit and see if the splitting is necessary.

tiegz:
> Will we want to be re-looking at all go packages after this goes in to make the data more current/correct? I could even see trying to clean up some of the old stuff

Def! I'm afraid of overwhelming pkg.go.dev tho -- maybe something like this?

```
Project.where(platform: "Go").find_each.with_index do |p, i|
  PackageManagerDownloadWorker.perform_at((i * 1.second).from_now, platform, name)
end
```

With 1.8 million Go packages, that'd take about 21 days.

The slowest thing in an `update()` I think is fetching each `go.mod` file for each version, about 2 sec per request. As that go Issue above says, we can fetch deps from pkg.go.dev, but they're only for the latest version :\

PR #2530, opened 2020-06-10
Capture raw scm license data and serialize

jsonperl:


app/models/repository.rb
```diff
@@ -111,14 +111,7 @@ def description_with_language
   def normalize_license_and_language
     self.language = Linguist::Language[self.language].to_s
     self.language = nil if self.language.blank?
-    return if license.blank?
-    if license.downcase == 'other'
```
katzj, line 115: Do we really not want to do any normalization on the repository license?

PR #2551, opened 2020-07-26
Fix typo of 'change' in app/assets/javascripts/application.js

phil-davis:
This looks like "just a typo" that has no real consequence for an ordinary user!

Was introduced by https://github.com/phil-davis/libraries.io/commit/da555b600012656d2a2e89160b251a1c0dea9f35


katzj:
Thanks!

PR #2554, opened 2020-07-26
Fix scope :without_subscriptions

phil-davis:
What impact does this typo have?


katzj:
I instead went for removing the scope as it isn't used anywhere

PR #2555, opened 2020-07-26
Typo login_uniqueness_with_case_insensitive_host

phil-davis:
This is just a typo in a name. No real run-time effect, but might as well correct it.


katzj:
Thanks!

PR #2563, opened 2020-08-14
Update chartkick and pghero for cves

mellisio:


Gemfile.lock
```diff
@@ -129,7 +129,7 @@ GEM
     case_transform (0.2)
       activesupport
     charlock_holmes (0.7.6)
-    chartkick (3.3.1)
+    chartkick (3.4.0)
```
jsonperl, line 132: It doesn't look like chartkick is actually used? I see it required in admin.js but then I didn't see anything that took advantage of it. Maybe worth a "scream test"

PR #2566, opened 2020-08-18
Remove travis

mellisio:


katzj:
We still use travis for libraries testing. Although if someone wanted to switch it to circle, we totally could.

lm322 (commented):
84051640041f70f285f6e6dc5a57e580a90a5b67

PR #2567, opened 2020-08-19
Add repository license to project serializer

mellisio:


app/serializers/project_serializer.rb
```diff
@@ -33,4 +34,8 @@ class ProjectSerializer < ActiveModel::Serializer
   def show_updated_at?
     instance_options[:show_updated_at]
   end
+
+  def repository_license
+    object.repository&.license
```
katzj, line 39: Just double checking -- are we doing a `.includes(:repository)` on queries before this already? Otherwise it ends up a lot of extra queries

mellisio: The find_project in the application controller includes both versions and repository

PR #2583, opened 2020-09-23
Two-table online migration for repository_dependencies bigint.

tiegz:
This migrates `repository_dependencies` from an integer to bigint primary key.

The process is:
* create new table, with the change
* create triggers, which will copy any inserts/udpates/deletes from old to new
* copy rows over from old to new, in chunks
* rename new to "archived_TIMESTAMP_repository_dependencies" and old to "repository_dependencies"
* drop triggers

### Blockers
* [ ] new primary key default doesn't get updated
* [ ] what to do with the materialized view that still points to old table?

### Output 

```
== 20200921173826 ConvertRepositoryDependencyIdToBigInt: migrating ============
-- create_table("repository_dependencies_new", {:id=>:bigserial, :force=>:cascade})
   -> 0.0233s
-- Creating insert trigger function
-- Creating insert trigger
-- Creating update trigger function
-- Creating update trigger
-- Creating delete trigger function
-- Creating delete trigger
-- Copying old rows to new table: 1 to 2, in chunks of 100000.
.--  0% 
-- Renaming tables (archive will be archived_2020_09_23_10_24_10_244_repository_dependencies).
   -> 0.0034s
-- Deleting delete trigger
-- Deleting insert trigger
-- Deleting update trigger
-- Deleting insert trigger functions
-- Deleting update trigger functions
-- Deleting delete trigger functions
== 20200921173826 ConvertRepositoryDependencyIdToBigInt: migrated (0.0916s) ===
```

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_07_09_221449) do
+ActiveRecord::Schema.define(version: 2020_09_21_173826) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "pg_stat_statements"
@@ -258,7 +258,7 @@ ActiveRecord::Schema.define(version: 2020_07_09_221449) do
     t.index ["status"], name: "index_repositories_on_status"
   end
 
-  create_table "repository_dependencies", id: :serial, force: :cascade do |t|
+  create_table "repository_dependencies", id: :bigint, default: -> { "nextval('repository_dependencies_new_id_seq6'::regclass)" }, force: :cascade do |t|
     t.integer "project_id"
     t.integer "manifest_id"
     t.boolean "optional"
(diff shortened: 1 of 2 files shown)


db/schema.rb
```diff
@@ -258,7 +258,7 @@
     t.index ["status"], name: "index_repositories_on_status"
   end
 
-  create_table "repository_dependencies", id: :serial, force: :cascade do |t|
+  create_table "repository_dependencies", id: :bigint, default: -> { "nextval('repository_dependencies_new_id_seq6'::regclass)" }, force: :cascade do |t|
```
tiegz, line 261: I don't think we should have the hardcoded sequence name as default, so trying to fix this now.

db/schema.rb
```diff
@@ -417,11 +417,11 @@
...
               repositories.rank,
               repositories.stargazers_count,
-              repository_dependencies.project_id
+              archived_2020_09_23_09_53_18_556_repository_dependencies.project_id
              FROM (repositories
-               JOIN repository_dependencies ON ((repositories.id = repository_dependencies.repository_id)))
+               JOIN archived_2020_09_23_09_53_18_556_repository_dependencies ON ((repositories.id = archived_2020_09_23_09_53_18_556_repository_dependencies.repository_id)))
             WHERE (repositories.private = false)
-            GROUP BY repositories.id, repository_dependencies.project_id) t1
+            GROUP BY repositories.id, archived_2020_09_23_09_53_18_556_repository_dependencies.project_id) t1
```
tiegz, line 424: Also trying to figure out if we can swap the view source dynamically.

tiegz:
Closing in favor of https://github.com/librariesio/libraries.io/pull/2584

PR #2584, opened 2020-09-23
 RepositoryDependency id migration step 2: migrate from integer to bigint

tiegz:
* converts `repository_dependencies.id` from integer to bigint 

This migration assumes that we're ok waiting for the table lock, and just temporarily serving bad data for this table until it's done.

### DIFF
--- a/db/migrate/20200921173826_convert_repository_dependency_id_to_big_int.rb
+++ b/db/migrate/20200921173826_convert_repository_dependency_id_to_big_int.rb
@@ -1,0 +1,10 @@
+class ConvertRepositoryDependencyIdToBigInt < ActiveRecord::Migration[5.2]
+  def up
+    # NB this locks the table, so we're disabling related code temporarily, while it runs.
+    change_column :repository_dependencies, :id, :bigint
+  end
+
+  def down
+    change_column :repository_dependencies, :id, :integer
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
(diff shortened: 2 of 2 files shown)


db/migrate/20200921173826_convert_repository_dependency_id_to_big_int.rb
```diff
@@ -0,0 +1,10 @@
+class ConvertRepositoryDependencyIdToBigInt < ActiveRecord::Migration[5.2]
+  def up
+    # NB this locks the table, so we're disabling related code temporarily, while it runs.
+    change_column :repository_dependencies, :id, :bigint
```
tiegz, line 4: @katzj @jsonperl do you know if Circle has a timeout? If so, we could skip this in Circle and run it manually in a long-lived terminal.

tiegz: Ah nevermind, just remembered we don't deploy via Circle.

db/migrate/20200921173826_convert_repository_dependency_id_to_big_int.rb
```diff
@@ -0,0 +1,10 @@
...
+  end
+
+  def down
+    change_column :repository_dependencies, :id, :integer
```
jsonperl, line 8: 🤣

jsonperl (approved):
I think this works. Prolly get a 👍 from @katzj too.

katzj:
This needs to be done as a few PRs -- the first to disable reading which we merge and rollout. Then the actual migration (so that code isn't running still trying to read while the migration happens). And then reverting of the first.

PR #2593, opened 2020-10-06
Add Spring Libs Release as a Maven Repository Source

mikeyoung85:
This adds a new class that extends the Maven PackageManager class to provide overrides on URLs for package found in the spring-libs-release repository.


app/models/package_manager/maven_manager/spring_libs.rb
```diff
@@ -0,0 +1,61 @@
+module MavenManager
```
mikeyoung85, line 1: I'm not sure about this module setup. I would like it to live under the PackageManager module but was having difficulties with the Base package manager class referencing things outside of its module.

app/models/package_manager/maven_manager/spring_libs.rb
```diff
@@ -0,0 +1,61 @@
...
+  class SpringLibs < PackageManager::Maven
+    REPOSITORY_SOURCE_NAME = 'SpringLibs'
+
+    def self.repository_base
+      "https://repo.spring.io/libs-release-local"
+    end
+
+    def self.project_names
+      get("http://localhost:8080/springLibReleases/all")
```
mikeyoung85, line 10: Need to update these to the live URL

app/models/package_manager/maven_manager/spring_libs.rb
```diff
@@ -0,0 +1,61 @@
...
+    end
+
+    def self.recent_names
+      get("http://localhost:8080/springLibReleases/recent")
+    end
+
+    def self.load_names(limit = nil)
+      project_names.each { |name| REDIS.sadd("maven-spring-lib-names", name)}
```
mikeyoung85, line 18: I don't anything is referencing this redis location.

app/models/package_manager/maven_manager/spring_libs.rb
```diff
@@ -0,0 +1,61 @@
...
+    end
+
+    def self.check_status_url(project)
+      MavenUrl.from_name(project.name, repository_base).base
+    end
+
+    def self.versions(_project, name)
+      versions_page = get_raw(MavenUrl.from_name(name, repository_base).base)
+      link_regex = /([\w\-.]+)\/<\/a>/
+      found_versions = versions_page
```
mikeyoung85, line 44: I need to take another look through this. There isn't an API or anything, so this looks at all the link names available when you look at the package on the repository directory page and takes the version names from those.

Pseudo code is:
1. load page
2. look for any lines that have a `href` in them.
3. use regex to get text used in the link tag
4. remove the parent directory link from the group, `..`

app/models/package_manager/maven_manager/spring_libs.rb
```diff
@@ -0,0 +1,61 @@
...
+        .split("\n")
+        .filter {|line| line.include?("href") && line.match?(link_regex)}
+        .map{|line| line.match(link_regex)[1]}
+        .reject{|line| line == ".."}
+
+      retrieve_versions(found_versions.filter {|item| !item.ends_with?("-SNAPSHOT")}, name)
+    end
+
+    def self.formatted_name
```
mikeyoung85, line 53: These need to be set to Maven so the platform_class is set properly during package update.

app/models/package_manager/maven.rb
```diff
@@ -18,15 +19,41 @@ class Maven < Base
     }.freeze
 
     def self.package_link(project, version = nil)
+      if version
```
jsonperl, line 22: When would we be hitting this without a version? And are we gonna drop the ball on source then?

app/models/package_manager/maven.rb
```diff
@@ -18,15 +19,41 @@ class Maven < Base
...
       MavenUrl.from_name(name, repository_base).jar(version)
     end
 
     def self.check_status_url(project)
-      MavenUrl.from_name(project.name, repository_base).base
+      sources = project.versions.flat_map(&:repository_sources).uniq
+      if sources.include?(REPOSITORY_SOURCE_NAME) 
+        MavenUrl.from_name(project.name, repository_base).base
+      else
+        "MavenManager::#{sources.first}".constantize.check_status_url(project)
```
jsonperl, line 55: Could this get a little simpler if extracted out maven central and treated it exactly how we do the other repositories.

app/models/package_manager/maven.rb
mikeyoung85: I think the only spot this is actually getting called is from https://github.com/librariesio/libraries.io/blob/005d6de3e86c2354f78183dccc690a004d60fcf9/app/models/project.rb#L205

The other spot is from Base `package_link` but that has been overridden. I could remove the `if` check here and I think it would be the same result.

app/models/package_manager/maven.rb
mikeyoung85: That's not a bad idea. I'm struggling with the selection of the source in here. In my mind, I wanted Maven Central to be the default, but it doesn't need to be. The only perceived "main" source is from the link shown on the Maven section web page which is driven by https://github.com/librariesio/libraries.io/blob/af2c4dbe4c6f4626c12c096b9d569c0188397c0a/app/models/package_manager/maven.rb#L11

jsonperl: Def gets rid of all the ` if sources.include?(REPOSITORY_SOURCE_NAME) ` kinda stuff

app/models/package_manager/maven.rb
```diff
@@ -18,15 +19,41 @@ class Maven < Base
     }.freeze
 
     def self.package_link(project, version = nil)
+      if version
+        db_version = project.versions.find_by(number: version)
+        if db_version.repository_sources.include?(REPOSITORY_SOURCE_NAME)
+          return MavenUrl.from_name(project.name, repository_base).search(version)
+        else
+          repository_source = db_version.repository_sources.first
+          return "MavenManager::#{repository_source}".constantize.package_link(project, version)
```
jsonperl, line 28: I can't help but be slightly nervous about this. Maybe we can add something like this to get em?
```
def get_source(source_name)
  if MavenManager.const_defined?(source_name)
    return MavenManager.const_get(source_name)
  end

  raise "Unknown source #{source_name}" 
end
```

jsonperl: Though looking at that, it protects us from... nothing

app/models/package_manager/maven_manager/spring_libs.rb
jsonperl: I think rails would want it to be
```
module PackageManager
  module MavenManager
    class SpringLibs < Maven
``` 
?
Or maybe even 
`class PackageManager::MavenManager::SrpingLibs` ?

app/models/package_manager/maven.rb
```diff
@@ -17,27 +18,52 @@ class Maven < Base
       "http://www.eclipse.org/org/documents/edl-v10" => "Eclipse Distribution License (EDL), Version 1.0",
     }.freeze
 
+    PROVIDER_MAP = {
```
jsonperl, line 21: 👍

app/models/package_manager/maven.rb
```diff
@@ -17,27 +18,52 @@ class Maven < Base
...
+
       MavenUrl.from_name(project.name, repository_base).search(version)
     end
 
     def self.download_url(name, version = nil)
+      if version
+        project = Project.find_by(name: name, platform: "Maven")
+        db_version = project.versions.find_by(number: version)
+        unless db_version&.repository_sources.blank?
+          repository_source = db_version.repository_sources.first
```
jsonperl, line 43: Can't we just sort of default these now?
`repository_source = db_version.repository_sources&.first.presence? || "Maven"` kinda thing?

app/models/package_manager/maven.rb
```diff
@@ -17,27 +18,52 @@ class Maven < Base
...
+        end
+      end
+
       MavenUrl.from_name(name, repository_base).jar(version)
     end
 
     def self.check_status_url(project)
-      MavenUrl.from_name(project.name, repository_base).base
-    end
+      sources = project.versions.flat_map(&:repository_sources).compact.uniq
```
jsonperl, line 52: Should we just go for the last version?

app/models/package_manager/maven.rb
mikeyoung85: That might help make it more explicit. I wrote each of these to fail down to what was already here if there wasn't a repository source for the version, which should be the same as saying `|| "Maven"`.

app/models/package_manager/maven.rb
```diff
@@ -17,27 +18,52 @@ class Maven < Base
...
+      return PROVIDER_MAP[sources.first].check_status_url(project) unless sources.blank?
```
jsonperl, line 53: Same deal, maybe just default to Maven source here if we didn't find it?

app/models/package_manager/maven.rb
```diff
@@ -17,27 +18,52 @@ class Maven < Base
...
-      "https://repo1.maven.org/maven2"
+      MavenUrl.from_name(project.name, repository_base).base
     end
 
     def self.load_names(limit = nil)
       names = get("https://maven.libraries.io/all")
       names.each { |name| REDIS.sadd("maven-names", name)}
     end
 
+    def self.repository_base
```
jsonperl, line 63: Should this live in the maven provider?

app/models/package_manager/maven.rb
```diff
@@ -214,6 +240,10 @@ def self.latest_version(versions, name)
       end
     end
 
+    def self.repository_source_name
```
jsonperl, line 243: Not sure where this comes into place, should it live in the provider too?

app/models/package_manager/maven.rb
mikeyoung85: By publish date or by created_at date? I'm wondering if we can still do some sort of priority on sources. Maybe setting that in the Maven source map and pick the highest priority source from the list in `versions`.

app/models/package_manager/maven.rb
jsonperl: Or maybe even a default constant up top? DEFAULT_SOURCE = "Maven" and then just use that all over

app/models/package_manager/maven.rb
mikeyoung85: I still need to test what removing this stuff from here will do. Ideally it wouldn't live here.

app/models/package_manager/maven.rb
jsonperl: Not even really sure what a check status url is...

mikeyoung85: This is the URL that is used to check if the project is deprecated or not. So this just needs to not 404.

mikeyoung85: https://github.com/librariesio/libraries.io/blob/005d6de3e86c2354f78183dccc690a004d60fcf9/app/models/project.rb#L520

jsonperl (approved):
This all works for me! We chatted about some future stuff to do when the rake task is updated.

app/models/package_manager/maven.rb
mikeyoung85: I've made "MavenCentral" the default now which simplifies the logic. Pick the first repository source if it's there and if not just use the MavenCentral provider.

jsonperl: Awesome, looks good! 👍

PR #2594, opened 2020-10-07
Maven Download Rake Tasks For Multiple Repository Sources

mikeyoung85:
Adjust the `rake download:maven` tasks to take in an optional provider parameter to run the sync with.

### DIFF
--- a/app/models/package_manager/base.rb
+++ b/app/models/package_manager/base.rb
@@ -284,7 +284,7 @@ module PackageManager
 
     private_class_method def self.download_async(names)
       names.each_slice(1000).each_with_index do |group, index|
-        group.each { |name| PackageManagerDownloadWorker.perform_in(index.hours, self.name.demodulize, name) }
+        group.each { |name| PackageManagerDownloadWorker.perform_in(index.hours, self, name) }
       end
     end
   end
--- a/app/models/package_manager/base.rb
+++ b/app/models/package_manager/base.rb
@@ -284,7 +284,7 @@ module PackageManager
 
     private_class_method def self.download_async(names)
       names.each_slice(1000).each_with_index do |group, index|
-        group.each { |name| PackageManagerDownloadWorker.perform_in(index.hours, self, name) }
+        group.each { |name| PackageManagerDownloadWorker.perform_in(index.hours, self.name, name) }
       end
     end
   end
--- a/app/models/package_manager/maven.rb
+++ b/app/models/package_manager/maven.rb
@@ -234,10 +234,6 @@ module PackageManager
       "Maven"
     end
 
-    def self.provider(provider)
-      PROVIDER_MAP[provider]
-    end
-
     class MavenUrl
       def self.from_name(name, repo_base)
         new(*name.split(":", 2), repo_base)
--- a/app/models/package_manager/maven.rb
+++ b/app/models/package_manager/maven.rb
@@ -234,6 +234,10 @@ module PackageManager
       "Maven"
     end
 
+    def self.provider(provider)
+      PROVIDER_MAP[provider]
+    end
+
     class MavenUrl
       def self.from_name(name, repo_base)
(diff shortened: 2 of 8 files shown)


app/workers/package_manager_download_worker.rb
```diff
@@ -1,10 +1,15 @@
...
 
-  def perform(class_name, name)
-    return unless class_name.present?
-    logger.info("Beginning update for #{class_name}/#{name}")
-    "PackageManager::#{class_name}".constantize.update(name)
+  def perform(clazz, name)
+    return unless clazz.present?
+
+    # need to maintain compatibility with things that pass in the name of the class under PackageManager module
+    clazz = "PackageManager::#{clazz}".constantize if clazz.is_a? String
```
mikeyoung85, line 11: I think the only two spots calling this with a String are

https://github.com/librariesio/libraries.io/blob/46c85a9e730e25e279267e8dcbba307c46916942/lib/tasks/projects.rake#L112

https://github.com/librariesio/libraries.io/blob/4e733b309780632cc81616dc1c84838f5f0b42f3/app/controllers/hooks_controller.rb#L17

lib/tasks/download.rake
```diff
@@ -1,298 +1,302 @@
...
+  desc "Download all Julia packages asynchronously"
   task julia: :environment do
     PackageManager::Julia.import_async
   end
 
-  desc 'Download recent Maven packages asynchronously'
-  task maven: :environment do
-    PackageManager::Maven.load_names(50)
-    PackageManager::Maven.import_recent_async
+  desc "Download recent Maven packages asynchronously"
```
mikeyoung85, line 168: This is the only section with real changes. The rest is rubocop.

app/models/project.rb
```diff
@@ -170,7 +170,7 @@ def set_last_synced_at
   end
 
   def async_sync
-    PackageManagerDownloadWorker.perform_async(platform, name)
+    PackageManagerDownloadWorker.perform_async(platform_class, name)
```
mikeyoung85, line 173: The way this is written, we will not be able to correctly sync something not on Maven Central. This eventually ends up at `PackageManager::Maven.update()` so maybe we can do something in there to try and figure out the right provider to use.

app/models/package_manager/base.rb
```diff
@@ -280,7 +280,7 @@ def self.deprecation_info(_name)
 
     private_class_method def self.download_async(names)
       names.each_slice(1000).each_with_index do |group, index|
-        group.each { |name| PackageManagerDownloadWorker.perform_in(index.hours, self.name.demodulize, name) }
+        group.each { |name| PackageManagerDownloadWorker.perform_in(index.hours, self, name) }
```
jsonperl, line 283: Won't this try and serialize the class for the job?

app/models/package_manager/maven.rb
```diff
@@ -231,6 +231,10 @@ def self.repository_source_name
       "Maven"
     end
 
+    def self.provider(provider)
+      PROVIDER_MAP[provider]
```
jsonperl, line 235: Might as well update the internal stuff to use that?

app/models/package_manager/base.rb
mikeyoung85: Yeah that's a good point. Maybe it should take in the full class name as a string and I'll update each spot to pass in the full name instead of assuming it will be appended with `PackageManager::`.

lib/tasks/download.rake
```diff
@@ -1,298 +1,312 @@
...
+        sub_class.import_recent_async
+      rescue StandardError => e
+        puts e
+        next
+      end
+    end
+  end
+
+  desc "Download all Maven packages asynchronously"
+  task :maven_all, [:provider] => :environment do |_task, args|
```
jsonperl, line 181: Can remove the arg now

jsonperl (approved):

PR #2596, opened 2020-10-08
Add Atlassian Maven Provider

mikeyoung85:


app/models/package_manager/maven/atlassian.rb
```diff
@@ -0,0 +1,45 @@
...
+  def self.project_names
+    get("http://localhost:8080/atlassian/all")
+  end
+
+  def self.recent_names
+    get("http://localhost:8080/atlassian/recent")
+  end
+
+  def self.package_link(project, version = nil)
+    if version
```
jsonperl, line 19: Maybe just call download_url with `(project.name, version)`

app/models/package_manager/maven/atlassian.rb
```diff
@@ -0,0 +1,45 @@
...
+      MavenUrl.from_name(name, repository_base).base
+    end
+  end
+
+  def self.check_status_url(project)
+    MavenUrl.from_name(project.name, repository_base).base
+  end
+
+  def self.formatted_name
+    PackageManager::Maven.formatted_name
```
jsonperl, line 39: why do this instead of just strings?

mikeyoung85: Actually I can remove this from here. It was cleaned up in the other branch I had going.

jsonperl (approved):

PR #2599, opened 2020-10-13
Project Sync with Multiple Providers

mikeyoung85:
Right now the resync button is broken for any project that doesn't use the default Maven Central provider. This PR would run the sync through all provider classes that this project has in it's Versions.


jsonperl (approved):
LGTM

PR #2610, opened 2020-11-13
Run Go Download as import_recent

mikeyoung85:
Previously the go download task was only looking at new unknown packages. This would change it to update every project from the recent index listing on index.golang.org.

### DIFF
--- a/app/models/package_manager/go.rb
+++ b/app/models/package_manager/go.rb
@@ -5,22 +5,21 @@ module PackageManager
     HAS_VERSIONS = true
     HAS_DEPENDENCIES = true
     BIBLIOTHECARY_SUPPORT = true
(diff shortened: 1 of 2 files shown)


app/models/package_manager/go.rb
```diff
@@ -30,10 +29,14 @@ def self.documentation_url(name, version = nil)
       "https://pkg.go.dev/#{name}#{"@#{version}" if version}?tab=doc"
     end
 
-    def self.install_instructions(project, version = nil)
+    def self.install_instructions(project, _version = nil)
       "go get #{project.name}"
     end
 
+    def self.recent_names
```
mikeyoung85, line 36: All of these changes are rubocop formatting. This is the only change here.

jsonperl (approved):

PR #2614, opened 2020-11-19
Go Version Handling

mikeyoung85:
Make sure base module gets all versions assigned to it and limit modules with major release on the end, github.com/libraries/goexample/v3 for example, to have versions only assigned to that major release stream.


spec/models/package_manager/go_spec.rb
```diff
@@ -46,7 +46,7 @@
   describe "#mapping" do
     it "maps data correctly from pkg.go.dev" do
       VCR.use_cassette("pkg_go_dev") do
-        project = described_class.project("github.com/stretchr/testify")
+        project = described_class.project("github.com/urfave/cli")
```
jsonperl, line 49: This is my fave

mikeyoung85: `github.com/ourfave/cli`

jsonperl (approved):
This is perfecto 👌

spec/models/package_manager/go_spec.rb
```diff
@@ -55,6 +55,78 @@
...
+  describe "#update" do
+    it "should queue update for non versioned module" do
+      VCR.use_cassette("pkg_go_dev_v2") do
+        expect(PackageManagerDownloadWorker).to receive(:perform_async).with(described_class.name, "github.com/urfave/cli")
+
+        described_class.update("github.com/urfave/cli/v2")
+      end
+    end
+
+    it "should update both the major release module and base module" do
```
mikeyoung85, line 107: @jsonperl added a spec that does the full run and checks the results in the database

app/models/package_manager/go.rb
```diff
@@ -21,6 +21,8 @@ class Go < Base
       ".svn",
     ].freeze
 
+    VERSION_MODULE_REGEX = /(.+)\/(v\d)/.freeze
```
tiegz, line 24: I suppose the day could come when people exceed single digits 😱 

```suggestion
    VERSION_MODULE_REGEX = /(.+)\/(v\d+)/.freeze
```

jsonperl: Oh totally great point! I think we're close already.

jsonperl (approved):

PR #2616, opened 2020-11-20
Get Go Module Versions from proxy.golang.org

mikeyoung85:
Reverts back to getting versions and publish dates from go proxy instead of pkg.go.dev, but avoids getting versions we already know about. Keeps base module project up to date directly instead of calling update on it.


app/models/package_manager/go.rb
```diff
@@ -53,34 +53,55 @@ def self.update(name)
...
     end
 
     def self.versions(project, _name)
       return [] if project.nil?
+      return project[:versions] if project[:versions]
 
-      versions = project[:versions_html]&.css(".Versions-item")&.map do |v|
-        { number: v.css("a").first.text, published_at: Chronic.parse(v.css(".Versions-commitTime").first.text) }
-      end
+      known_versions = Project.find_by(platform: "Go", name: project[:name])&.versions || []
```
mikeyoung85, line 82: Still trying to think of a way to simplify this code a bit.

app/models/package_manager/go.rb
```diff
@@ -53,34 +53,55 @@ def self.update(name)
       # example: github.com/myexample/modulename/v2
       if (matches = name.match(VERSION_MODULE_REGEX))
         PackageManagerDownloadWorker.perform_async(self.name, matches[1])
```
mikeyoung85, line 55: We could remove this call to update the base module from here, but it may help fill in some missing data for now.

app/models/package_manager/go.rb
```diff
@@ -53,34 +53,54 @@ def self.update(name)
...
+        &.reject(&:blank?)
+        &.map do |v|
+          return known_versions.find_by(number: v).slice(:number, :published_at) if known_versions&.find_by(number: v).present?
+
+          info = get("http://proxy.golang.org/#{project[:name]}/@v/#{v}.info")
+          {
+            number: info["Version"],
+            published_at: info["Time"].presence && Time.parse(info["Time"]),
+          }
+        rescue Oj::ParseError
```
mikeyoung85, line 100: I think this rescue is in the right place. It needs to rescue from the `get()` call inside the map block. Rubocop wants to indent it 34 tabs in.

app/models/package_manager/go.rb
```diff
@@ -53,34 +53,54 @@ def self.update(name)
...
+
+        module_project = Project.find_by(platform: "Go", name: name)
+        base_module_project = Project.find_by(platform: "Go", name: matches[1])
+        return if module_project.nil? || base_module_project.nil?
+
+        module_project.versions.each do |version|
+          base_module_project.versions.find_or_create_by(number: version[:number]) do |new_version|
+            new_version.update(published_at: version[:published_at])
+            new_version.save
+          end
```
tiegz, line 65: This'll only get run for new records, but if you wanted to run it for either new-or-existing, you could use `tap` instead:

```suggestion
          base_module_project.versions.find_or_create_by(number: version[:number]).tap do |new_version|
            new_version.update(published_at: version[:published_at])
          end
```

app/models/package_manager/go.rb
```diff
@@ -53,34 +53,54 @@ def self.update(name)
...
-      end
-
-      if (matches = project[:name].match(VERSION_MODULE_REGEX))
-        versions = versions.select { |version| version[:number].start_with?(matches[2]) }
-      end
-
-      versions
+      return project[:versions] if project[:versions]
+
+      known_versions = Project.find_by(platform: "Go", name: project[:name])&.versions
```
tiegz, line 82: Might be worth just fetching the list of {number, published_at} here to avoid the N+1 on line 93:

```suggestion
      known_versions = Project
        .find_by(platform: "Go", name: project[:name])
        &.versions
        &.select(:number, :published_at)
```

app/models/package_manager/go.rb
```diff
@@ -53,34 +53,54 @@ def self.update(name)
...
+      # NB fetching versions from the html only gets dates without timestamps, but we could alternatively use the go proxy too:
+      #   1) Fetch the list of versions: https://proxy.golang.org/#{module_name}/@v/list
+      #   2) And for each version, fetch https://proxy.golang.org/#{module_name}/@v/#{v}.info
+
+      get_raw("http://proxy.golang.org/#{project[:name]}/@v/list")
+        &.lines
+        &.map(&:strip)
+        &.reject(&:blank?)
+        &.map do |v|
+          return known_versions.find_by(number: v).slice(:number, :published_at) if known_versions&.find_by(number: v).present?
```
tiegz, line 93: Followup to my last comment, this could then be:

```suggestion
          known_version = known_versions.find { |(number, _published_at)| number == v }
          return known_version.slice(:number, :published_at) if known_version.present?
```

app/models/package_manager/go.rb
jsonperl: I think the intent was to not worry about the versions we've already captured in the past. As an optimization we could even do something like:
```
        new_base_versions = module_project.versions - base_module_project.versions.where(id: module_project.versions).pluck(:number)
        new_base_versions.each do |version|
          base_module_project.versions.create(number: version[:number], published_at: version[:published_at])
        end
```

app/models/package_manager/go.rb
```diff
@@ -53,34 +53,54 @@ def self.update(name)
       # example: github.com/myexample/modulename/v2
       if (matches = name.match(VERSION_MODULE_REGEX))
```
jsonperl, line 54: Might be nice to put this whole thing in another #update_base_version if name.match(VERSION_MODULE_REGEX) kinda thing

app/models/package_manager/go.rb
jsonperl: Yea it's probably not necessary? Esp if we go back and do some cleanup.

app/models/package_manager/go.rb
jsonperl: Add a .to_h for one that's indexed by the the version number!

app/models/package_manager/go.rb
```diff
@@ -53,34 +53,54 @@ def self.update(name)
...
-      versions
+      return project[:versions] if project[:versions]
+
+      known_versions = Project.find_by(platform: "Go", name: project[:name])&.versions
+
+      # NB fetching versions from the html only gets dates without timestamps, but we could alternatively use the go proxy too:
+      #   1) Fetch the list of versions: https://proxy.golang.org/#{module_name}/@v/list
+      #   2) And for each version, fetch https://proxy.golang.org/#{module_name}/@v/#{v}.info
+
+      get_raw("http://proxy.golang.org/#{project[:name]}/@v/list")
```
jsonperl, line 88: Maybe put the proxy URL in a const so we can try others?

app/models/package_manager/go.rb
mikeyoung85: I think I will keep it in here for now and then remove it once I figure out how to get all the base modules into Libraries.

mikeyoung85:
@tiegz @jsonperl thank you for the comments. I've updated the PR.

jsonperl:
> @tiegz @jsonperl thank you for the comments. I've updated the PR.

🙌 Team work makes the dream work! 🙌

jsonperl (approved):

PR #2619, opened 2020-11-24
Update Go Versioned with Base Modules

mikeyoung85:
This task would go through and start updates for Go modules that have a `/vx` on the end of the name. The intention is to get the base module project in place if it doesn't exist and then run the update on the versioned module to get versions added correctly to both projects.

### DIFF
--- a/lib/tasks/projects.rake
+++ b/lib/tasks/projects.rake
@@ -180,4 +180,18 @@ namespace :projects do
       end
     end
   end
+
+  desc 'Update Go Base Modules'
+  task :update_go_base_modules, [:version] => :environment do |_task, args|
+    # go through versioned module names and rerun update on them to get their versions
+    # added to the base module Project
+    Project.where(platform: "Go").where("name like ?", "%/v#{args[:version]}").find_in_batches do |projects|
+      projects.each do |project|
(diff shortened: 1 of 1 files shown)


lib/tasks/projects.rake
```diff
@@ -180,4 +180,18 @@ namespace :projects do
...
   end
+
+  desc 'Update Go Base Modules'
+  task :update_go_base_modules, [:version] => :environment do |_task, args|
+    # go through versioned module names and rerun update on them to get their versions
+    # added to the base module Project
+    Project.where(platform: "Go").where("name like ?", "%/v#{args[:version]}").find_in_batches do |projects|
+      projects.each do |project|
+        matches = PackageManager::Go::VERSION_MODULE_REGEX.match(project.name)
+        Project.create(platform: "Go", name: matches[1])
```
jsonperl, line 191: Wouldn't we want a find_or_create? (Or create_or_find)

jsonperl: I stand corrected, this is fine and is non-explody

mikeyoung85: I was banking on `create` silently failing based on the unique indexes in the database and then continuing on afterwards. I'll switch it out to find_or_create_by since that shows the intention a little better.

jsonperl (approved):

PR #2620, opened 2020-11-25
Verify Go Projects

mikeyoung85:
Added a worker and rake task to go through and verify Go project names against the Go proxy and remove projects that are not found on there. I am going to open this as a draft and let you guys think about if you want to run through these projects.

### DIFF
--- a/app/models/package_manager/go.rb
+++ b/app/models/package_manager/go.rb
@@ -173,5 +173,10 @@ module PackageManager
     def self.get_repository_url(project)
       request("https://#{project['Package']}").to_hash[:url].to_s
     end
+
+    def self.valid_project?(name)
+      response = request("#{PROXY_BASE_URL}/#{name}/@v/list")
+      response.status != 410
+    end
   end
 end
--- a/lib/tasks/projects.rake
+++ b/lib/tasks/projects.rake
@@ -195,20 +195,28 @@ namespace :projects do
     end
   end
 
-  # bin/rake projects:verify_go_projects[2020-01-01,2020-12-31]
   desc 'Verify Go Projects'
-  task :verify_go_projects, [:after_date, :before_date] => :environment do |_task, args|
-    limit = args.limit || 1000
-    before_date = Date.parse(args.before_date)
-    after_date = Date.parse(args.after_date)
-    puts "Date range: #{before_date} - #{after_date}"
+  task :verify_go_projects, [:count] => :environment do |_task, args|
+    loop do
+      count ||= 1000
+      start_id = REDIS.get("go:update:latest_updated_id").presence || 0
+      puts "Start id: #{start_id}, count: #{start_id.to_i + count}"
+      projects = Project
+        .where("id > ?", start_id)
+        .order(:id)
+        .limit(count)
 
-    Project
-      .where(platform: "Go")
-      .where("created_at BETWEEN ? AND ?", after_date, before_date)
-      .find_each do |project|
-        puts "Queueing verification for #{project.name}"
-        GoProjectVerificationWorker.perform_async(project.name)
+      if projects.count.zero?
+        puts "Done!"
+        exit
+      elsif (queue_size = Sidekiq::Queue.new('small').size) > 0
+        puts "Waiting for #{queue_size} jobs..."
+        sleep 5
+      else
+        projects
+          .each { |p| PackageManagerDownloadWorker.perform_async("PackageManager::Go", p.name) }
+        REDIS.set("go:update:latest_updated_id", projects.last.id)
       end
+    end
   end
 end
--- a/lib/tasks/projects.rake
+++ b/lib/tasks/projects.rake
@@ -195,11 +195,13 @@ namespace :projects do
     end
   end
 
+  # bin/rake projects:verify_go_projects[2020-01-01,2020-12-31]
   desc 'Verify Go Projects'
-  task :verify_go_projects, [:after_date, :before_date, :limit] => :environment do |_task, args|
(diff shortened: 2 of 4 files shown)


lib/tasks/projects.rake
```diff
@@ -194,4 +194,21 @@ namespace :projects do
       end
     end
   end
+
+  desc 'Verify Go Projects'
+  task :verify_go_projects, [:before_date, :limit] => :environment do |_task, args|
```
mikeyoung85, line 199: I thought going by created_at date would be the most sensible way of scrolling through the projects since the older ones are more likely to be incorrect.

app/models/package_manager/go.rb
```diff
@@ -173,5 +173,10 @@ def self.project_find_names(name)
     def self.get_repository_url(project)
       request("https://#{project['Package']}").to_hash[:url].to_s
     end
+
+    def self.valid_project?(name)
+      response = request("#{PROXY_BASE_URL}/#{name}/@v/list")
+      response.status != 410
```
jsonperl, line 179: 410 is "GONE" btw 🤣

tiegz (commented):
Just to be clear, this will also remove all "module packages" and only keep "modules", so it'll remove e.g. "github.com/kubernetes/kubernetes/cmd/genman".

We have about 1.9 million Go projects in total, and the request to the proxy takes about 1-4-ish seconds on average.

Alternative to the rake task, we could also bring in the `BackfillBatchWorker` and use that to automate the job queuing?

lib/tasks/projects.rake
```diff
@@ -194,4 +194,21 @@ namespace :projects do
...
+    limit = args.limit || 1000
+    before_date = Date.parse(args.before_date) || Date.today + 1.day
```
tiegz, line 201: Wondering if this would need an `after_date` too. I don't think there's a way to tell if we've already checked a project or not.

jsonperl:
> Just to be clear, this will also remove all "module packages" and only keep "modules", so it'll remove e.g. "github.com/kubernetes/kubernetes/cmd/genman".
> 
> We have about 1.9 million Go projects in total, and the request to the proxy takes about 1-4-ish seconds on average.
> 
> Alternative to the rake task, we could also bring in the `BackfillBatchWorker` and use that to automate the job queuing?

Yea yikes that's like... a couple months yea?

jsonperl:
> Just to be clear, this will also remove all "module packages" and only keep "modules", so it'll remove e.g. "github.com/kubernetes/kubernetes/cmd/genman".

That's the idea really. Modules are the thing that has a version and releases right? (And what's specified in go.mod files). Packages are just the bits you use in the code, but from a versioning and "data about a package" perspective I think we just want to stick to modules.

tiegz:
> Yea yikes that's like... a couple months yea?

Yeah 😬 although in parallel it should be cool. Just updated the rake task to take a date range, so we can knock out ranges one-at-a-time.

> That's the idea really. Modules are the thing that has a version and releases right?

👍 much better. I'm not sure how this will affect Depci yet tho.

jsonperl:
Date ranges seem like a decent plan. Nothing is harmed by the cruft being here, we can just kick em off periodically.

lib/tasks/projects.rake
```diff
@@ -194,4 +194,29 @@ namespace :projects do
...
+        exit
+      elsif (queue_size = Sidekiq::Queue.new('small').size) > 0
+        puts "Waiting for #{queue_size} jobs..."
+        sleep 5
+      else
+        projects
+          .each { |p| PackageManagerDownloadWorker.perform_async("PackageManager::Go", p.name) }
+        REDIS.set("go:update:latest_updated_id", projects.last.id)
+      end
+    end
```
tiegz, line 220: @jsonperl here's what I was thinking: create the loop in the rake task so we can keep it running in tmux, and avoid re-booting the Rails env every invocation.

lib/tasks/projects.rake
```diff
@@ -195,20 +195,28 @@ namespace :projects do
...
-
-    Project
-      .where(platform: "Go")
-      .where("created_at BETWEEN ? AND ?", after_date, before_date)
-      .find_each do |project|
-        puts "Queueing verification for #{project.name}"
-        GoProjectVerificationWorker.perform_async(project.name)
+  task :verify_go_projects, [:count] => :environment do |_task, args|
+    loop do
+      count ||= 1000
```
jsonperl, line 201: this should be `count = args.count || 1000`

lib/tasks/projects.rake
jsonperl: Can't you just do that in bash? I'm not a huge fan of assuming a SIGINT for using this thing or it all gets queued.

`watch -n 60 bundle exec rake projects:verify_go_projects[1000]` would do the trick no?

tiegz: Loop removed

lib/tasks/projects.rake
```diff
@@ -194,4 +194,25 @@ namespace :projects do
...
+      .where("id > ?", start_id)
+      .order(:id)
+      .limit(args.count)
+
+    if projects.count.zero?
+      puts "Done!"
+      exit
+    else
+      projects
+        .each { |p| PackageManagerDownloadWorker.perform_async("PackageManager::Go", p.name) }
```
jsonperl, line 214: Hmm isn't this the wrong job? This PR added a `GoProjectVerificationWorker`

tiegz: Yes! My bad, I thought I fixed that yesterday.

jsonperl (approved):

PR #2627, opened 2020-12-09
Maven: skip versions that have missing poms, rather than raising err.

tiegz:
When parsing versions of a Maven project, we may come across a version with a missing pom (or in the case below, a missing parent pom). 

Currently `retrieve_versions` will blow up if any versions are missing, so instead of blowing up and not updating *any* versions, let's update the versions that work still and ignore the broken version.

As precedent, we're [skipping bad versions](https://github.com/librariesio/libraries.io/blob/master/app/models/package_manager/go.rb#L110-L111) in Go currently too.

#### Working
Pom: https://repo1.maven.org/maven2/org/opencastproject/opencast-common/8.5/opencast-common-8.5.pom
Parent pom: https://repo1.maven.org/maven2/org/opencastproject/base/8.5/base-8.5.pom

#### Broken
Pom: https://repo1.maven.org/maven2/org/opencastproject/opencast-common/7.7/opencast-common-7.7.pom
Missing parent pom: https://repo1.maven.org/maven2/org/opencastproject/base/7.7/base-7.7.pom

This results in Opencast's versions not getting updated to the latest (8.9):

Screen Shot 2020-12-09 at 2 51 00 PM

### DIFF
--- a/spec/models/package_manager/maven_spec.rb
+++ b/spec/models/package_manager/maven_spec.rb
@@ -128,7 +128,10 @@ describe PackageManager::Maven do
 
       expect(described_class.project("javax.faces:javax.faces-api")).to eq(expected)
     end
+  end
 
+
+  describe ".versions" do
     it "skips versions that can't be parsed" do
       expect(described_class)
         .to receive(:get_raw)
--- a/spec/models/package_manager/maven_spec.rb
+++ b/spec/models/package_manager/maven_spec.rb
@@ -128,6 +128,29 @@ describe PackageManager::Maven do
 
       expect(described_class.project("javax.faces:javax.faces-api")).to eq(expected)
     end
+
+    it "skips versions that can't be parsed" do
+      expect(described_class)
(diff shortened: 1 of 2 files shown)


spec/models/package_manager/maven_spec.rb
```diff
@@ -128,6 +128,29 @@
...
+
+      # TODO: these are probably bugs... it's using the version 3.2.0/etc of a depdendency and looking that up on itself
+      allow(described_class)
+        .to receive(:get_pom)
+          .with("com.google.api.grpc", "proto-google-common-protos", "3.2.0")
+          .and_raise(Ox::ParseError.new(""))
+      allow(described_class)
+        .to receive(:get_pom)
+          .with("com.google.api.grpc", "proto-google-common-protos", "${api.version}")
+          .and_raise(Ox::ParseError.new(""))
```
tiegz, line 150: Probably gonna look into this bug separately next.

mellisio (approved):

PR #2630, opened 2020-12-15
Kick off CheckStatusWorker in Project#async_sync.

tiegz:
* removes the `Project#sync` method because I don't think it's called from anywhere
* instead of checking deprecation/removal status via ^, we can check it by kicking off `CheckStatusWorker` when an async sync happens on the project

NB There **are** currently deprecated and removed projects in the database, but I think that's because Projects **sometimes** get marked as such by their parent repository [here](https://github.com/librariesio/libraries.io/blob/41159948ff612c0813f909b73a4bda697bb92e7f/app/models/repository.rb#L272-L287). In NPM's case, deprecation happens via the NPM CLI, so we'd never get triggered to update the Repository's status. This should update it more often on Projects.

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -152,25 +152,13 @@ class Project < ApplicationRecord
     save
   end
 
-  def sync
-    check_status
-    if status == 'Removed'
-      set_last_synced_at
-      return
-    end
-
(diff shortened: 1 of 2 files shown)


jsonperl (approved):
Works for me!

PR #2632, opened 2020-12-17
Log active connection count

jsonperl:
Wanted to get an eye on what active connections look like when we see stability issues.


katzj (commented):
Maybe stick in the base application controller and not the api one?

jsonperl:
> Maybe stick in the base application controller and not the api one?

Good call

katzj (approved):

PR #2634, opened 2020-12-17
Fix argument in Go rake task.

tiegz:
Hah! `count` is a method on args, not the actual arg. Silly mistake.


jsonperl (approved):
Haha so it was limiting it to 1? :D

PR #2636, opened 2020-12-17
Only consider package deprecated if the last version is deprecated.

tiegz:
This maps to NPM's logic, where the latest version is displayed for the package landing page, and the deprecation message shows up at the top if that version is deprecated.

For example, buildmail:

* https://www.npmjs.com/package/buildmail <-- deprecated, bc it's really 4.0.1
* https://www.npmjs.com/package/buildmail/v/4.0.1 <-- deprecated
* https://www.npmjs.com/package/buildmail/v/3.0.0-beta.0. <-- not deprecated

### DIFF
--- a/app/models/package_manager/npm.rb
+++ b/app/models/package_manager/npm.rb
@@ -40,11 +40,11 @@ module PackageManager
     end
 
     def self.deprecation_info(name)
-      versions = project(name)["versions"].values
+      last_version = project(name)["versions"].values.last
 
       {
-        is_deprecated: versions.all? { |v| v["deprecated"] },
-        message: versions.last["deprecated"],
+        is_deprecated: !!last_version["deprecated"],
+        message: last_version["deprecated"],
       }
     end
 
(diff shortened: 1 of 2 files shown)


app/models/package_manager/npm.rb
```diff
@@ -40,11 +40,11 @@ def self.project(name)
     end
 
     def self.deprecation_info(name)
-      versions = project(name)["versions"].values
+      last_version = project(name)["versions"].values.last
```
jsonperl, line 43: We sure these are in timeorder? Could max_by just to be sure.

jsonperl (approved):
Nice-n-easy 👍

app/models/package_manager/npm.rb
tiegz: We were using timeorder before to get the deprecation message, and in my experience they've been ordered. But I'll take a quick look to see if we can sort by semver to be sure.

jsonperl: If they generally seem like they're in the right order coming in, I'd trust that over an attempted semver sort 🤣

tiegz:
Same! 😬

PR #2642, opened 2020-12-23
This makes things easier to run with a default setup db

jsonperl:
I'm just running a a docker service for old PG and this makes things way easier to run off the stock image.


katzj (commented):
Can we make this able to be overridden by an environment variable for people who aren't using docker to run postgres?

katzj (approved):

PR #2643, opened 2020-12-23
Add google cloud tracing to active model serializers

jsonperl:


config/initializers/active_model_serializers.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+module ASMTrace
+  def serializable_hash(*)
+    if Rails.env.production?
+      Google::Cloud::Trace.in_span "active_model_serializers#serializable_hash##{self.class.name}" do |_span|
```
tiegz, line 6: If this doesn't work, we may need to [set the trace context first](https://github.com/googleapis/google-cloud-ruby/blob/master/google-cloud-trace/lib/google/cloud/trace.rb#L242).

I assume Rails already sets one, so I'll try to track down to verify.

tiegz (approved):

config/initializers/active_model_serializers.rb
jsonperl: It'd be inside the stackdriver gem (or something stackdriver requires). Though I assume it must since we're getting AR traces already...

tiegz: Yeah, stackdriver is just an empty thing that loads the google-cloud-trace gem's railtie. Digging thru the railtie to figure out what it does.

katzj (approved):
The trace side seems fine to me

PR #2644, opened 2020-12-27
Optimize status check endpoint

jsonperl:
An attempt at optimization for the check endpoint. 

Initially I assumed this was going to be an N+1 fiesta, but in digging around here, it seems like queries are optimized quite nicely thanks to some work done by @kbarrette in https://github.com/librariesio/libraries.io/pull/2399 and https://github.com/librariesio/libraries.io/pull/2418. Turns out the bulk of the work happens in serialization. This adds a stripped down serializer that avoids active model serializers entirely, and tries to optimize a little bit (merge! vs merge, etc...). We also force usage of Oj for json serialization in this endpoint, something we probably want to explore doing universally. I've left a couple things that seemed useful that I hacked in to do some optimization.

A test run on my machine using 500 bootstrapped projects shows a 2.4x improvement, but I suspect it'll be better than that in production. Only one way to find out!

```
/api/check?v2=true
  1.341680   0.018838   1.360518 (  1.389088)

/api/check
  3.230206   0.034801   3.265007 (  3.286004)
```

### DIFF
--- a/app/serializers/optimized_project_serializer.rb
+++ b/app/serializers/optimized_project_serializer.rb
@@ -1,0 +1,97 @@
+# frozen_string_literal: true
+
+# Note: This is _not_ an active model serializer. To use, instantiate and
+# call #serialize
+class OptimizedProjectSerializer
+  PROJECT_ATTRIBUTES = %w[
+    dependent_repos_count
+    dependents_count
+    deprecation_reason
+    description
+    forks
+    homepage
+    keywords
+    language
+    latest_download_url
+    latest_download_url
+    latest_release_number
+    latest_release_published_at
+    latest_stable_release_number
+    latest_stable_release_published_at
+    license_normalized
+    license_set_by_admin
+    licenses
+    normalized_licenses
+    package_manager_url
+    platform
+    rank
+    repository_url
+    score
+    stars
+    status
+  ].freeze
+
+  VERSION_ATTRIBUTES = %w[
+    number
+    published_at
+    spdx_expression
+    original_license
+    researched_at
+    repository_sources
+  ].freeze
+
+  MAINTENANCE_STAT_ATTRIBUTES = %w[
+    category
+    value
+    updated_at
+  ].freeze
+
+  def initialize(projects, requested_name_map, internal_key = false)
+    @projects = projects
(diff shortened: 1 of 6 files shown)


spec/requests/api/status_spec.rb
```diff
@@ -1,185 +1,218 @@
...
+          json = JSON.parse(response.body)
+          expect(json.size).to eq(2)
+          expect(json.select { |p| p[:name] == "bcrypt" && p[:platform] == "NPM" }).to be
+          expect(json.select { |p| p[:name] == "bcrypt" && p[:platform] == "Pypi" }).to be
+        end
       end
     end
   end
+
+  xit "is performant with a bunch of projects" do
```
jsonperl, line 190: Leaving this here but not running normally. I expect to iterate a little bit here in the future and would like to continue to use this.

app/serializers/optimized_project_serializer.rb
```diff
@@ -0,0 +1,97 @@
+# frozen_string_literal: true
+
+# Note: This is _not_ an active model serializer. To use, instantiate and
+# call #serialize
+class OptimizedProjectSerializer
```
katzj, line 5: Is it worth looking at using jb in libraries also?

config/initializers/google_cloud_trace.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+# Fake google cloud trace in_span method so we don't have to check our environment
```
katzj, line 3: Is it worth sticking this in stackdriver.rb like the rest of the trace initialization? Also, I'm confused that we don't have to do this on the tidelift side?

katzj (approved):
I'm okay with starting down this path. A couple of questions that are worth considering but we can do 'em after this is in

app/serializers/optimized_project_serializer.rb
jsonperl: Totally worth exploring.

config/initializers/google_cloud_trace.rb
jsonperl: Hmm yea I’ll take a look at doing that in a follow up.

PR #2646, opened 2020-12-28
Set up pushing static assets to GCS

katzj:
This is a lot of plumbing so that static assets are assets/GIT_HASH/ and that they then end up pushed to google cloud storage by cloud build

### DIFF
--- a/config/initializers/git_revision.rb
+++ b/config/initializers/git_revision.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+Rails.application.config.git_revision = if Rails.env.development?
+    `git rev-parse HEAD`.strip
+  else
+    ENV["REVISION_ID"]
+  end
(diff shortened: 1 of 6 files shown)


config/initializers/git_revision.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+Rails.application.config.git_revision = if Rails.env.development?
+    `git rev-parse HEAD`
+  else
+    ENV["REVISION_ID"].presence
```
jsonperl, line 6:
```suggestion
    ENV["REVISION_ID"]
```

config/initializers/git_revision.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+Rails.application.config.git_revision = if Rails.env.development?
+    `git rev-parse HEAD`
```
jsonperl, line 4: Looks like this has a newline, maybe `git rev-parse HEAD`.strip

jsonperl (approved):

kbarrette (approved):

PR #2649, opened 2020-12-29
Further check endpoint optimizations

jsonperl:
Ditching AR includes and plucking what we need for versions and maintenance stats makes this over 4x faster now in a local test with 500 projects.

```
/api/check?v2=true
  0.848421   0.020203   0.868624 (  0.892780)

/api/check
  3.559839   0.027464   3.587303 (  3.628964)
```

### DIFF
--- a/app/controllers/api/status_controller.rb
+++ b/app/controllers/api/status_controller.rb
@@ -12,6 +12,8 @@ class Api::StatusController < Api::BulkProjectController
   end
 
   def check_legacy
+    @includes = %i[repository versions repository_maintenance_stats]
+
     render(
       json: projects,
       each_serializer: ProjectStatusSerializer,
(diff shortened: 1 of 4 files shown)


katzj (approved):

app/controllers/api/status_controller.rb
```diff
@@ -23,9 +25,33 @@ def check_legacy
   end
 
   def check_new
+    @includes = %i[repository]
+
     # This forces serialization with Oj
     # TODO: Investigate calling Oj.mimic_JSON on initializationh to swap universally
     serializer = OptimizedProjectSerializer.new(projects, project_names, internal_api_key?)
```
jsonperl, line 32: Oj was doing a bad job serializing TimeWithZone and I didn't want to fight too hard

PR #2652, opened 2020-12-30
Add some new indexes on projects and repositories based on query patterns

katzj:


kbarrette (approved):
Change looks fine.

Does indexing on a boolean gain us much here? I guess it immediately restricts the query to the small subset of forked/private repos?

katzj:
I don't know that it's gonna get us a ton -- but hard to really tell without trying it. And if not, can just drop the index 🤷‍♂️

PR #2658, opened 2021-01-06
Add package-deprecation support for pypi yanking.

tiegz:
Pypi is similar to NPM in that it only allows specific releases to be "yanked", but having the latest release yanked is treated as if the whole package is yanked.

NB: according to PEP 592, yanked packages are still downloadable, and they won't be used by package managers unless they're the only possible release that satisfies a manifest.

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -518,6 +518,9 @@ class Project < ApplicationRecord
       update_attribute(:status, 'Removed')
     elsif platform.downcase != 'packagist' && [400, 404].include?(response.response_code)
       update_attribute(:status, 'Removed')
+    elsif platform.downcase == 'pypi' && [404].include?(response.response_code)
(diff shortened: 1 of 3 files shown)


app/models/project.rb
```diff
@@ -516,6 +516,9 @@ def check_status(removed = false)
     response = Typhoeus.head(url)
     if platform.downcase == 'packagist' && response.response_code == 302
       update_attribute(:status, 'Removed')
+    elsif platform.downcase == 'pypi' && response.response_code == 404
+      # TODO: remove this stanza once this bug is fixed: https://github.com/pypa/warehouse/issues/3709#issuecomment-754973958
+      update_attribute(:status, 'Deprecated')
```
tiegz, line 521: I didn't add a test for this because it's temporary, but I did confirm that it works in the console.

katzj (approved):

jsonperl (commented):
Here's a post merge 👍

PR #2685, opened 2021-01-22
First step to remove a bunch of queues. More queues can significantly…

mellisio:


app/workers/repository_maintenance_stat_worker.rb
```diff
@@ -1,25 +1,8 @@
 class RepositoryMaintenanceStatWorker
     include Sidekiq::Worker
-    sidekiq_options queue: :repo_maintenance_stat, retry: 3
+    sidekiq_options queue: :low, retry: 3
 
     def perform(repo_id)
         Repository.find(repo_id).gather_maintenance_stats
     end
-
-    def self.enqueue(repo_id, priority: :medium)
```
katzj, line 9: These priorities were to let us intentionally get the first set of maintenance stats "fast" and refreshes to be able to be lower priority

mellisio: I'm not seeing very many of these queued relative to the others, can I just make them all critical?

jsonperl:
I don't have a lot of historical reference here so I'm gonna trust you and @katzj to make the call. I think having true queue priority (this) is much more sensible then whatever psuedo priority stuff we get with weights. Though this means that NOTHING gets processed while the critical queue is draining, which changes the system dynamics a bit, and I don't know if that's a problem or not.

app/workers/repository_maintenance_stat_worker.rb
katzj: Making them all critical makes it so we can overload the system in a way that we really don't want to allow. Especially with the changed processing order.

PR #2687, opened 2021-01-25
Handle unfetchable Go versions that have name collisions.

tiegz:
Ignore versions that have case-insensitive name collisions instead of blowing up when we try to fetch their info.

* [example Go Module version with name collision](https://github.com/ysweid/aws-sdk-go/tree/v1.12.68/models/apis)
* [breaking url for ^](https://proxy.golang.org/github.com/ysweid/aws-sdk-go/@v/v1.12.68.info)
* [Go issue](https://github.com/golang/go/issues/33778)

### DIFF
--- a/app/models/package_manager/go.rb
+++ b/app/models/package_manager/go.rb
@@ -106,13 +106,14 @@ module PackageManager
             known_versions[v].slice(:number, :published_at)
           else
             info = get("#{PROXY_BASE_URL}/#{project[:name]}/@v/#{v}.info")
-            # Known Go Modules issue where versions with case-insensitive name collisions break go get
+            # Store nil published_at for known Go Modules issue where case-insensitive name collisions break go get
             # e.g. https://proxy.golang.org/github.com/ysweid/aws-sdk-go/@v/v1.12.68.info
-            next if info.nil?
+            version = info.nil? ? v : info["Version"]
+            published_at = info && info["Time"].presence && Time.parse(info["Time"])
 
             {
-              number: info["Version"],
-              published_at: info["Time"].presence && Time.parse(info["Time"]),
+              number: version,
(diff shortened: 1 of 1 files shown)


app/models/package_manager/go.rb
```diff
@@ -106,6 +106,10 @@ def self.versions(project, _name)
             known_versions[v].slice(:number, :published_at)
           else
             info = get("#{PROXY_BASE_URL}/#{project[:name]}/@v/#{v}.info")
```
jsonperl, line 108: Any way you can think of to correct to the proper case before updating? Some sort of case insensitive search mechanism?

tiegz: It's unfortunately an issue with the files in the Go Module and not with the name itself so I think we have to wait on a fix from Golang 😬 Here's the response for the example one above, where it has two files in the same folder with same-name-different-casing:

`not found: create zip: case-insensitive file name collision: "models/apis/GuardDuty/2017-11-28/api-2.json" and "models/apis/guardduty/2017-11-28/api-2.json"`

tiegz: (or alternatively we could keep the version but just display its published_at as `?` or something?)

jsonperl: Seems like a really really small amount of packages huh? (Like 2 maybe?)

jsonperl (approved):

app/models/package_manager/go.rb
tiegz: I think this example had 3 versions with name collisions, but we've got a ton of these errors in bugsnag: https://app.bugsnag.com/tidelift/libraries-dot-io/errors/60076c589e7ce30018c6b586

PR #2688, opened 2021-01-25
Add per release go sync

mellisio:


app/workers/package_manager_download_worker.rb
```diff
@@ -65,7 +65,12 @@ def perform(platform_name, name, version = nil)
     raise "Platform '#{platform_name}' not found" unless platform
 
     # need to maintain compatibility with things that pass in the name of the class under PackageManager module
-    logger.info("Package update for platform=#{key} name=#{name}")
-    platform.update(name)
+    if platform::SUPPORTS_SINGLE_VERSION_UPDATE && version.present?
+      logger.info("Version update for platform=#{key} name=#{name} version=#{version}")
+      platform.update_version(name, version)
```
jsonperl, line 70: What if update_version just knew how to handle this internally instead of branching here and exposing a constant?

jsonperl: And the signatures all change to `platform.update(name, version = nil)` 

I s'pose that's a lot more files to touch though.

app/models/package_manager/go.rb
```diff
@@ -52,8 +53,20 @@ def self.project_names(since = 1.day.ago)
...
-      super(name)
+    def self.one_version(name, version_string)
+      info = get("#{PROXY_BASE_URL}/#{name}/@v/#{version_string}.info")
+      # Store nil published_at for known Go Modules issue where case-insensitive name collisions break go get
+      # e.g. https://proxy.golang.org/github.com/ysweid/aws-sdk-go/@v/v1.12.68.info
+      version_string = info.nil? ? version_string : info["Version"]
+      published_at = info && info["Time"].presence && Time.parse(info["Time"])
+      {
+        number: version_string,
+        published_at: published_at,
```
jsonperl, line 64: Hey @mellisio, thoughts on trying to avoid this if we send published_at to the job from depper?

mellisio: Unless we intend to start sending the license too I don't think it matters. One thing we could do would be to send published_at, number, license then we wouldn't even have to hit an API on libraries side.

jsonperl (approved):

PR #2691, opened 2021-01-26
Make Version deprecatable, and implement Rubygems "yanking" deprecation.

tiegz:
Q: should we start showing Version statuses publicly on the project pages too?

### DIFF
--- a/app/models/package_manager/base.rb
+++ b/app/models/package_manager/base.rb
@@ -110,9 +110,9 @@ module PackageManager
       end
 
       if self::HAS_VERSIONS && sync_versions
-        versions(project, dbproject.name).each do |vers|
-          add_version(dbproject, vers)
-        end
+        versions(project, dbproject.name)
+          .each { |v| add_version(dbproject, v) }
+          .tap { |vs| deprecate_versions(dbproject, vs) }
       end
 
       save_dependencies(mapped_project) if self::HAS_DEPENDENCIES
@@ -134,6 +134,7 @@ module PackageManager
 
       dbproject = Project.find_by!(name: mapped_project[:name], platform: db_platform)
       add_version(dbproject, one_version(dbproject.name, version))
+      # TODO handle deprecation here too
       save_dependencies(mapped_project) if self::HAS_DEPENDENCIES
       finalize_dbproject(dbproject)
     end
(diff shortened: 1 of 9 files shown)


app/models/project.rb
```diff
@@ -9,7 +9,7 @@ class Project < ApplicationRecord
   include BitbucketProject
 
   HAS_DEPENDENCIES = false
-  STATUSES = ['Active', 'Deprecated', 'Unmaintained', 'Help Wanted', 'Removed']
+  STATUSES = ['Active', 'Deprecated', 'Unmaintained', 'Help Wanted', 'Removed', 'Hidden'].freeze
```
tiegz, line 12: 'Hidden' is used on many projects in the database -- IMO we should document it here (and then I fix up some logic for this below)

app/models/version.rb
```diff
@@ -3,8 +3,11 @@
 class Version < ApplicationRecord
   include Releaseable
 
-  validates_presence_of :project_id, :number
-  validates_uniqueness_of :number, scope: :project_id
+  STATUSES = ['Deprecated', 'Removed'].freeze
```
jsonperl, line 6: Assuming you checked these? (After the convo in slack) :D

app/views/admin/projects/show.html.erb
```diff
@@ -26,7 +26,7 @@
 
     <%= form.input :repository_url %>
     <%= form.input :normalized_licenses, include_blank: true, selected: @project.normalized_licenses, label: 'License', collection: Project.popular_licenses(:facet_limit => 300).map{|l| format_license(l['key']) } %>
-    <%= form.input :status, include_blank: true, collection: Project::STATUSES + ['Hidden'] %>
```
jsonperl, line 29: 🤣

jsonperl (approved):
This looks great!

app/models/package_manager/base.rb
```diff
@@ -146,6 +147,16 @@ def self.add_version(dbproject, version_hash)
       existing.save
     end
 
+    def self.deprecate_versions(dbproject, version_hash)
+      case dbproject.platform.downcase
+      when "rubygems" # yanked gems will be omitted from project JSON versions
```
johnbintz-tidelift, line 152: :pray:

jsonperl: I'm assuming these will get stashed in their respective package manager (at some point) yea?

app/models/version.rb
tiegz: Yeah, these are the values for the new column, but I did check the `Project#version` values in the db.

app/models/package_manager/base.rb
tiegz: I kept it in here so it's next to the `add_version` method, but it would def make sense to extract to each class later too 👍

jsonperl: I think that's generally been the pattern. Along with a constant check 🤷‍♂️

PR #2708, opened 2021-02-02
GitHub issue stats via graphql

katzj:
This switches the maintenance stats collection for issues to make queries against graphql instead of looking at our issues database.


app/models/maintenance_stats/queries/github/issues_query.rb
```diff
@@ -0,0 +1,42 @@
...
+        ISSUES_QUERY = Rails.application.config.graphql.client.parse <<-GRAPHQL
+          query($owner: String!, $repo_name: String!, $one_year: GitTimestamp!){
+            repository(owner: $owner, name: $repo_name){
+              openIssues: issues(states: OPEN, filterBy:{since:$one_year}) {
+                totalCount
+              },
+              closedIssues:issues(states:CLOSED, filterBy:{since:$one_year}){
+                totalCount
+              },
+              openPrs:pullRequests(states:OPEN){
```
katzj, line 14: Ugh. `pullRequests` doesn't support `filterBy`. But we already were just looking at everything in the previous incarnation of this so it's not a regression (see https://github.com/librariesio/libraries.io/blob/master/app/models/maintenance_stats/stats/github/db_issues_stats.rb)

mikeyoung85: I thought we had something with the V3 client that would paginate back one year, but I can't find that anywhere looking through the history.

app/models/maintenance_stats/stats/github/issue_stats.rb
```diff
@@ -0,0 +1,64 @@
+module MaintenanceStats
```
mikeyoung85, line 1: Could we add a spec for this? The other stats have simple specs that check if the data is in the right spots, https://github.com/librariesio/libraries.io/blob/master/spec/models/maintenance_stats/github/commit_count_stats_spec.rb

app/models/maintenance_stats/queries/github/issues_query.rb
tiegz: Could use this ugly workaround until they support it -- you'd have to interpolate the repo name and date in Ruby too:

``` graphql
query ($owner: String!, $repo_name: String!, $one_year: GitTimestamp!) {
  openPrs: search(query: "repo:#{repo_name} is:pr is:open created:>#{one_year.to_date}", type: ISSUE) {
    issueCount
  }
  closedPrs: search(query: "repo:#{repo_name} is:pr is:closed created:>#{one_year.to_date}", type: ISSUE) {
    issueCount
  }
  repository(owner: $owner, name: $repo_name) {
    openIssues: issues(states: OPEN, filterBy: {since: $one_year}) {
      totalCount
    }
    closedIssues: issues(states: CLOSED, filterBy: {since: $one_year}) {
      totalCount
    }
  }
}
```

katzj: The graphql client doesn't let you do interpolation like that for security reasons 🤦 

Oh well, we'll just go with this.

katzj:
Alright, tests are happy and added one looking at what we get from graphql for issue stats

mikeyoung85 (approved):

tiegz (approved):

jsonperl (approved):

PR #2720, opened 2021-02-08
Trace endpoints for better instrumentation.

tiegz:
It's easier to track down bottlenecks when you can drill into specific endpoints.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -8,7 +8,14 @@ class ApplicationController < ActionController::Base
 
   before_action :log_active_connections
 
+  around_action :trace_span
+
   private
+
+  def trace_span(&block)
+    Google::Cloud::Trace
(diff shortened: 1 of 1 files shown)


katzj (approved):

jsonperl:
🎉

app/controllers/application_controller.rb
```diff
@@ -8,8 +8,15 @@ class ApplicationController < ActionController::Base
 
   before_action :log_active_connections
 
+  around_action :trace_span
+
   private
 
+  def trace_span(&block)
+    Google::Cloud::Trace
```
jsonperl, line 16: Is there anything we need to install / setup to get activerecord tracing? I'm not sure how that works.

tiegz: I think we get that out of the box here: https://github.com/googleapis/google-cloud-ruby/blob/master/google-cloud-trace/lib/google/cloud/trace/rails.rb#L73-L79

PR #2721, opened 2021-02-08
Adds a task to backfill Project#latest_stable_release_number

tiegz:
Looks like `Project#latest_stable_release_number` was added a while ago: https://github.com/librariesio/libraries.io/pull/2231. If we backfill this and start using it instead of looking it up in Ruby, we can fix an N+1 in our DependencySerializer (used in Api::Repositories#dependencies)

### DIFF
--- a/lib/tasks/one_off.rake
+++ b/lib/tasks/one_off.rake
@@ -24,28 +24,16 @@ namespace :one_off do
     end
   end
 
-  desc "backfill missing clojars packages with a slash, and remove packages with a dot"
-  task cleanup_clojar_projects: :environment do
+  desc "backfill all latest_stable_release_numbers"
+  task backfill_latest_stable_release_numbers: :environment do
+    count = 0
     Project.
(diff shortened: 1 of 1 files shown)


katzj:
I'd think we should be able to rely on it in that serializer... we rely on it in other places and I am pretty sure I did teh backfill a couple years ago when I added this. I'd imagine the things without a stable release number don't have a latest release number.

tiegz:
@katzj I'd found a bunch that didn't have the same value for the synthetic method vs the db-method, but when I double-checked it just now it looks like they're just out-of-date:

```
> pp no_match.map { |n| [n.read_attribute(:latest_stable_release_number), n.latest_stable_release_number] }; nil
[["1.6.0", "2.0.0"], ["3.5.6", "3.5.7"], ["2.0.0", "2.0.1"], ...]
```

tiegz:
Closing this one, but maybe we can look into the syncronization problem with some of these separately.

PR #2722, opened 2021-02-08
Start using db-based latest_stable_release_number instead of Ruby version

tiegz:
To be merged after this backfill is run: https://github.com/librariesio/libraries.io/pull/2721


katzj (approved):
I don't think this is getting used by the setter but double check me before hitting merge?

tiegz:
@katzj confirmed that the setter is not using this method 👍

PR #2723, opened 2021-02-09
Add JB gem, and a first JB template for Api::Projects#dependencies.

tiegz:
An attempt at speeding up `/api/projects/:platform/:name/:version/dependencies`:

* adds `jb` gem for simple JSON templates
* adds `v2=true` param that switches from ProjectSerializer to JB template in this endpoint

Sample project with 1820 versions, roughly before and after:

```
> Benchmark.ms { ProjectSerializer.new(p).as_json }
=> 277.5426210137084
> Benchmark.ms { p.versions.map { |v| v.slice(:number, :published_at, :spdx_expression, :original_license, :researched_at, :repository_sources) }.as_json }
=> 82.78486994095147
```

### DIFF
--- a/app/controllers/api/projects_controller.rb
+++ b/app/controllers/api/projects_controller.rb
@@ -66,15 +66,14 @@ class Api::ProjectsController < Api::ApplicationController
   end
 
   def dependencies
-    subset = params.fetch(:subset, "default")
+    @subset = params.fetch(:subset, "default")
 
     if params[:v2] == "true"
       @project = Project.find_best!(params[:platform], params[:name], [:repository, :versions])
       @version = @project.find_version!(params[:version])
-      @subset = subset
-      # render app/views/api/projects/dependencies.json.b
+      # render app/views/api/projects/dependencies.json.jb
     else
-      project_json = find_project_as_json_with_dependencies!(params[:platform], params[:name], params[:version], subset)
+      project_json = find_project_as_json_with_dependencies!(params[:platform], params[:name], params[:version], @subset)
       render json: project_json
     end
   end
--- a/app/controllers/api/projects_controller.rb
+++ b/app/controllers/api/projects_controller.rb
@@ -67,9 +67,16 @@ class Api::ProjectsController < Api::ApplicationController
 
   def dependencies
     subset = params.fetch(:subset, "default")
-    project_json = find_project_as_json_with_dependencies!(params[:platform], params[:name], params[:version], subset)
 
-    render json: project_json
+    if params[:v2] == "true"
+      @project = Project.find_best!(params[:platform], params[:name], [:repository, :versions])
+      @version = @project.find_version!(params[:version])
+      @subset = subset
+      # render
+    else
+      project_json = find_project_as_json_with_dependencies!(params[:platform], params[:name], params[:version], subset)
+      render json: project_json
+    end
   end
 
   def dependencies_bulk
--- a/app/controllers/api/projects_controller.rb
+++ b/app/controllers/api/projects_controller.rb
@@ -72,7 +72,7 @@ class Api::ProjectsController < Api::ApplicationController
       @project = Project.find_best!(params[:platform], params[:name], [:repository, :versions])
       @version = @project.find_version!(params[:version])
       @subset = subset
-      # render
+      # render app/views/api/projects/dependencies.json.b
     else
       project_json = find_project_as_json_with_dependencies!(params[:platform], params[:name], params[:version], subset)
       render json: project_json
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -600,11 +600,11 @@ class Project < ApplicationRecord
   end
 
   def find_version!(version_name)
-    version = if version_name == 'latest'
-                versions.sort.first
-              else
-                versions.find_by_number(version_name)
-              end
+    if version_name == 'latest'
+      version_name = versions.pluck(:number).sort.last
+    end
+
+    version = versions.find_by_number(version_name)
 
     raise ActiveRecord::RecordNotFound if version.nil?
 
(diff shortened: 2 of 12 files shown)


app/serializers/project_serializer.rb
```diff
@@ -11,7 +11,6 @@ class ProjectSerializer < ActiveModel::Serializer
     latest_download_url
     latest_release_number
     latest_release_published_at
-    latest_stable_release
```
tiegz, line 14: @katzj I noticed you'd added the two attributes below [about two years ago](https://github.com/librariesio/libraries.io/pull/2247), so it seemed safe to remove `latest_stable_release`? LMK if this is a bad idea tho.

app/views/api/projects/dependencies.json.jb
```diff
@@ -0,0 +1,30 @@
+dependencies_fields = {
+  dependencies_for_version: @version.number,
+  dependencies: (@version.dependencies.includes(:project) || []).
+    map {|dependency| DependencySerializer.new(dependency) }.
+    as_json
```
jsonperl, line 5: I don't _think_ we need the as_json here? But not exactly sure how it figures it out. Perhaps I've been deluding myself.

app/views/api/projects/dependencies.json.jb
```diff
@@ -0,0 +1,30 @@
...
+}
+
+case @subset
```
jsonperl, line 8: I'm missing where @subset gets set, is this missing the controller checkin?

jsonperl: I was looking at a commit, not the PR haha, ignore.

app/controllers/api/projects_controller.rb
```diff
@@ -67,9 +67,16 @@ def updated
 
   def dependencies
     subset = params.fetch(:subset, "default")
```
jsonperl, line 69: Can probably just add the @ here and skip 74

app/controllers/api/projects_controller.rb
```diff
@@ -67,9 +67,16 @@ def updated
...
-    project_json = find_project_as_json_with_dependencies!(params[:platform], params[:name], params[:version], subset)
 
-    render json: project_json
+    if params[:v2] == "true"
+      @project = Project.find_best!(params[:platform], params[:name], [:repository, :versions])
```
jsonperl, line 72: Not for now: This find_best! method is goofy and is a pile of spaghetti to track. Ultimately we index on 'lower((platform)::text), lower((name)::text)` so why not just skip to that. 🤷‍♂️

app/controllers/api/projects_controller.rb
```diff
@@ -67,9 +67,16 @@ def updated
...
+      @version = @project.find_version!(params[:version])
```
jsonperl, line 73: Another pretty large optimization here would be this https://github.com/librariesio/libraries.io/blob/100cafcc0bae517dd43b15ea51a7ba00fca7bd7f/app/models/project.rb#L599 where we grab and hydrate all versions to then throw them all out.

app/views/api/projects/dependencies.json.jb
```diff
@@ -0,0 +1,30 @@
...
+when "default"
+  @project.
+    slice(Project::API_FIELDS).
+    merge({
+      versions: @project.versions.map do |v|
```
jsonperl, line 13: If we wanna squeeze a little extra out of this we could pluck all the necessary fields. We'd have to twiddle repository sources after the fact but could be worth it. Something along the lines of https://github.com/librariesio/libraries.io/blob/100cafcc0bae517dd43b15ea51a7ba00fca7bd7f/app/serializers/optimized_project_serializer.rb#L83

spec/requests/api/subscriptions_spec.rb
```diff
@@ -44,7 +44,7 @@
       put "/api/subscriptions/#{subscription.project.platform}/#{subscription.project.name}?api_key=#{user.api_key}", params: { subscription: { include_prerelease: true } }
       expect(response).to have_http_status(:success)
       expect(response.content_type).to eq('application/json')
-      expect(response.body).to be_json_eql subscription.reload.to_json(only: [:include_prerelease, :created_at, :updated_at], include: {project: {only: Project::API_FIELDS, methods: [:package_manager_url, :stars, :forks, :keywords, :latest_stable_release, :latest_download_url, :repository_license], include: {versions: {only: [:number, :published_at]} }}})
+      expect(response.body).to be_json_eql subscription.reload.to_json(only: [:include_prerelease, :created_at, :updated_at], include: {project: {only: Project::API_FIELDS, methods: [:package_manager_url, :stars, :forks, :keywords, :latest_download_url, :repository_license], include: {versions: {only: [:number, :published_at]} }}})
```
jsonperl, line 47: Yikes these specs 😬

jsonperl (approved):
Looks good, couple thoughts

app/controllers/api/projects_controller.rb
tiegz: Ooh, that's a good find.

tiegz: Actually, this one might be tricky to do now, and we'll use the `versions` associations in the JB template anyway so I'm gonna look at this later.

spec/requests/api/subscriptions_spec.rb
tiegz: Urgh, yeah, maybe I should clean this up.

app/models/project.rb
```diff
@@ -595,11 +600,11 @@ def download_registry_users
...
-    version = if version_name == 'latest'
-                versions.sort.first
-              else
-                versions.find_by_number(version_name)
-              end
+    if version_name == 'latest'
+      version_name = versions.pluck(:number).sort.last
+    end
+
+    version = versions.find_by_number(version_name)
```
tiegz, line 607: Did it, w00t.

app/controllers/api/projects_controller.rb
jsonperl: We only use them for the non-limited version

PR #2731, opened 2021-02-16
Tweak GCP intializer.

tiegz:
Just realized that there's a `Google::Cloud.configure.use_trace` variable that gets set in the Railtie, and it determines if the middleware is added. Removing the monkeypatch will mean that some real Trace code will run inside `in_span`, but it should no-op anyway.

### DIFF
--- a/config/initializers/google_cloud_trace.rb
+++ b/config/initializers/google_cloud_trace.rb
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
-require "google/cloud/trace"
+require "google/cloud/trace/rails"
 
 puts "Google::Cloud::Trace enabled." if Google::Cloud.configure.use_trace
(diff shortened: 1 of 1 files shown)


config/initializers/google_cloud_trace.rb
```diff
@@ -1,15 +1,5 @@
...
-    module Cloud
-      class Trace
-        def self.in_span(_span_name)
-          yield
-        end
-      end
-    end
-  end
-end
+require "google/cloud/trace/rails"
```
tiegz, line 3: https://github.com/googleapis/google-cloud-ruby/blob/1a7cfef70298276a9203658748e9d70edfcfb2d5/google-cloud-trace/lib/google/cloud/trace/rails.rb#L38

jsonperl: Where does `Cloud.configure.use_trace` get set?

tiegz: It only gets set to `true` if credentials are valid and if it's Production: https://github.com/googleapis/google-cloud-ruby/blob/1a7cfef70298276a9203658748e9d70edfcfb2d5/google-cloud-trace/lib/google/cloud/trace/rails.rb#L133-L142

jsonperl (approved):

PR #2735, opened 2021-02-17
Don't require rubocop by default.

tiegz:
Rubocop loads about 20mb into memory, and we don't need it most of the time (especially in production).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -128,6 +128,6 @@ end
 
 gem "scenic", "~> 1.4"
 
-gem "rubocop", "~> 0.80.1"
+gem "rubocop", "~> 0.80.1", require: false
 
 gem "chartkick", "~> 3.4"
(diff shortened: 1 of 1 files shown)


katzj:
What about just moving into the development/test section?

tiegz:
@katzj that was my first thought but the Rubocop readme recommends just not requiring it 🤷 Maybe bc it's run in test mode in CI?

tiegz:
(nevermind, missed that you said test too. I'll try it out)

Gemfile
```diff
@@ -103,6 +103,7 @@ end
 group :development, :test do
   gem "pry"
   gem "rspec-rails"
+  gem "rubocop", "~> 0.80.1", require: false
```
tiegz, line 106: Gonna keep the `require: false` because we don't need it for Rails server, console, etc.

katzj (approved):

jsonperl (approved):

PR #2736, opened 2021-02-17
PackageManagerDownloadWorker refactor: step 1

tiegz:
Cleanup, refactor, and wrangle an N+1.

### DIFF
--- a/app/models/package_manager/base.rb
+++ b/app/models/package_manager/base.rb
@@ -88,35 +88,8 @@ module PackageManager
     end
 
     def self.map_project(project)
-      mapped_project = mapping(project)
-      mapped_project = mapped_project.delete_if { |_key, value| value.blank? } if mapped_project.present?
-      mapped_project
-    end
-
-    def self.save(project, sync_versions: true)
-      return unless project.present?
-
-      mapped_project = map_project(project)
-      return false unless mapped_project.present?
-
-      dbproject = Project.find_or_initialize_by({ name: mapped_project[:name], platform: db_platform })
-      if dbproject.new_record?
-        dbproject.assign_attributes(mapped_project.except(:name, :releases, :versions, :version, :dependencies, :properties))
-        dbproject.save
-      else
-        dbproject.reformat_repository_url
-        attrs = mapped_project.except(:name, :releases, :versions, :version, :dependencies, :properties)
-        dbproject.update_attributes(attrs)
-      end
-
-      if self::HAS_VERSIONS && sync_versions
-        versions(project, dbproject.name)
-          .each { |v| add_version(dbproject, v) }
-          .tap { |vs| deprecate_versions(dbproject, vs) }
-      end
-
-      save_dependencies(mapped_project) if self::HAS_DEPENDENCIES
-      finalize_dbproject(dbproject)
+      mapping(project)
+        &.delete_if { |_key, value| value.blank? }
     end
 
     def self.update_version(name, version)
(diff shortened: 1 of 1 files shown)


app/models/package_manager/base.rb
```diff
@@ -88,35 +88,8 @@ def self.platform_name(platform)
     end
 
     def self.map_project(project)
-      mapped_project = mapping(project)
-      mapped_project = mapped_project.delete_if { |_key, value| value.blank? } if mapped_project.present?
-      mapped_project
-    end
-
-    def self.save(project, sync_versions: true)
```
tiegz, line 96: There's not really a reason for this to be outside of the `update()` method, so I consolidated them.

app/models/package_manager/base.rb
```diff
@@ -217,26 +205,25 @@ def self.new_names
...
-                 dependencies(name, version.number, mapped_project)
+                 dependencies(name, db_version.number, mapped_project)
                rescue StandardError
                  []
                end
-        next unless deps&.any? && version.dependencies.empty?
 
         deps.each do |dep|
-          next if dep[:project_name].blank? || version.dependencies.find_by_project_name(dep[:project_name])
+          next if dep[:project_name].blank? || db_version.dependencies.any? { |d| d.project_name == dep[:project_name] }
```
tiegz, line 219: Here's the db win in this PR: we've avoided an N+1 above with `db_project.versions.includes(:dependencies)`, but then reintroduced the N+1 and made it N^2+1 ( by doing `version.dependencies.find_by_project_name` for each dep for each version 😆 . This should fix.

app/models/package_manager/base.rb
```diff
@@ -132,42 +105,57 @@ def self.update_version(name, version)
...
     end
 
-    def self.finalize_dbproject(dbproject)
-      dbproject.reload
-      dbproject.download_registry_users
-      dbproject.last_synced_at = Time.now
-      dbproject.save
-      dbproject
+    def self.finalize_db_project(db_project)
+      db_project.reload
```
jsonperl, line 134: Do we need the reload?

app/models/package_manager/base.rb
```diff
@@ -132,42 +105,57 @@ def self.update_version(name, version)
...
+      db_project.reformat_repository_url unless db_project.new_record?
+      db_project.update(mapped_project.except(:name, :releases, :versions, :version, :dependencies, :properties))
+
+      if self::HAS_VERSIONS && sync_versions
+        versions(project, db_project.name)
+          .each { |v| add_version(db_project, v) }
+          .tap { |vs| deprecate_versions(db_project, vs) }
+      end
+
+      save_dependencies(mapped_project) if self::HAS_DEPENDENCIES
```
jsonperl, line 157: Looks like we do this in line 111 too?

app/models/package_manager/base.rb
```diff
@@ -132,42 +105,57 @@ def self.update_version(name, version)
...
       proj = project(name)
-      save(proj, sync_versions: sync_versions) if proj.present?
+      return unless project.present?
+
+      mapped_project = map_project(project)
+      return false unless mapped_project.present?
+
+      db_project = Project.find_or_initialize_by({ name: mapped_project[:name], platform: db_platform })
+      db_project.reformat_repository_url unless db_project.new_record?
+      db_project.update(mapped_project.except(:name, :releases, :versions, :version, :dependencies, :properties))
```
tiegz, line 149: Simplified this stanza that was originally in `self.save()`

app/models/package_manager/base.rb
jsonperl: Ah that's the version specific side of this.

app/models/package_manager/base.rb
```diff
@@ -132,42 +105,57 @@ def self.update_version(name, version)
...
+
+      if self::HAS_VERSIONS && sync_versions
```
jsonperl, line 151: Wonder if a later cleanup would be relying on all the stuff in update_version and just iterating through versions in the update method.

app/models/package_manager/base.rb
tiegz: I bet we don't -- in fact I could probably just move this logic into the main `update()` method and try to consolidate all the UPDATEs.

app/models/package_manager/base.rb
jsonperl: Now *that* is what I'm talking about! 💪

app/models/package_manager/base.rb
jsonperl: Do it!

tiegz: (I'm gonna push this to a followup PR to keep these PRs chunked)

jsonperl: Screen Shot 2021-02-17 at 2 18 07 PM

jsonperl (approved):

PR #2738, opened 2021-02-17
 PackageManagerDownloadWorker refactor: step 2

tiegz:
This combines the `PackageManager::Base.update` and `PackageManager::Base.update_version` methods because they're doing similar things, and `update_version` was wrapping `update` anyway.

This should also fix an issue with Maven where we fetched all the versions twice (i.e. if a package has 2000 versions, we'll make 4000 requests total). 😬

### DIFF
--- a/app/models/package_manager/base.rb
+++ b/app/models/package_manager/base.rb
@@ -105,7 +105,7 @@ module PackageManager
       return false unless mapped_project.present?
 
       db_project = Project.find_or_initialize_by({ name: mapped_project[:name], platform: db_platform })
-      db_project.reformat_repository_url unless db_project.new_record?
+      db_project.reformat_repository_url if sync_version == :all && !db_project.new_record?
       db_project.update(mapped_project.except(:name, :releases, :versions, :version, :dependencies, :properties))
 
       if self::HAS_VERSIONS
--- a/app/models/package_manager/base.rb
+++ b/app/models/package_manager/base.rb
@@ -94,7 +94,7 @@ module PackageManager
 
     def self.update(name, sync_version: :all)
       if sync_version != :all && !self::SUPPORTS_SINGLE_VERSION_UPDATE
-        logger.warn("#{db_platform}.update(#{name}, sync_versions: #{sync_version}) called but not supported on platform")
+        logger.warn("#{db_platform}.update(#{name}, sync_version: #{sync_version}) called but not supported on platform")
         return
       end
 
--- a/app/models/package_manager/base.rb
+++ b/app/models/package_manager/base.rb
@@ -92,25 +92,43 @@ module PackageManager
         &.delete_if { |_key, value| value.blank? }
     end
 
-    def self.update_version(name, version)
-      unless self::SUPPORTS_SINGLE_VERSION_UPDATE
-        logger.warn("#{db_platform}.update_version(#{name}, #{version}) called but not supported on platform")
+    def self.update(name, sync_version: :all)
+      if sync_version != :all && !self::SUPPORTS_SINGLE_VERSION_UPDATE
+        logger.warn("#{db_platform}.update(#{name}, sync_versions: #{sync_version}) called but not supported on platform")
         return
       end
 
-      update(name, sync_versions: false)
       raw_project = project(name)
+      return unless raw_project.present?
+
       mapped_project = map_project(raw_project)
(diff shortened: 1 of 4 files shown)


app/models/package_manager/maven.rb
```diff
@@ -158,10 +158,14 @@ def self.dependencies(name, version, project)
...
-      xml_versions = Nokogiri::XML(xml_metadata).css("version").map(&:text)
-      retrieve_versions(xml_versions.filter { |item| !item.ends_with?("-SNAPSHOT") }, name)
+    def self.versions(project, name)
+      if project && project[:versions]
+        return project[:versions]
+      else
+        xml_metadata = get_raw(MavenUrl.from_name(name, repository_base, NAME_DELIMITER).maven_metadata)
+        xml_versions = Nokogiri::XML(xml_metadata).css("version").map(&:text)
+        retrieve_versions(xml_versions.filter { |item| !item.ends_with?("-SNAPSHOT") }, name)
+      end
```
tiegz, line 168: @jsonperl turns out this double-fetch was mostly a problem with Maven, so this uses the pre-fetched `project` var if it's available, instead of refetching the versions.

tiegz: (this also only speeds up the `sync_version: :all` case, and I'm going to work on speeding up the single-version update after this PR)

app/models/package_manager/base.rb
```diff
@@ -92,25 +92,43 @@ def self.map_project(project)
         &.delete_if { |_key, value| value.blank? }
     end
 
-    def self.update_version(name, version)
-      unless self::SUPPORTS_SINGLE_VERSION_UPDATE
-        logger.warn("#{db_platform}.update_version(#{name}, #{version}) called but not supported on platform")
+    def self.update(name, sync_version: :all)
+      if sync_version != :all && !self::SUPPORTS_SINGLE_VERSION_UPDATE
+        logger.warn("#{db_platform}.update(#{name}, sync_version: #{sync_version}) called but not supported on platform")
```
tiegz, line 97: Should we raise here instead?

jsonperl: Warn seems fine, it's not really an error I think. It should honestly just silently *not* do a version update, then we don't have to inspect that logic anywhere but here.

app/models/package_manager/base.rb
```diff
@@ -92,25 +92,43 @@ def self.map_project(project)
...
+          # TODO handle deprecation here too
+        end
       end
 
-      db_project = Project.find_by!(name: mapped_project[:name], platform: db_platform)
-      add_version(db_project, one_version(db_project.name, version))
-      # TODO handle deprecation here too
       save_dependencies(mapped_project) if self::HAS_DEPENDENCIES
       finalize_db_project(db_project)
+    rescue SystemExit, Interrupt
```
jsonperl, line 124: What the heck is this about (I realize it's just copied)

app/models/package_manager/base.rb
```diff
@@ -92,25 +92,43 @@ def self.map_project(project)
...
-        logger.warn("#{db_platform}.update_version(#{name}, #{version}) called but not supported on platform")
+    def self.update(name, sync_version: :all)
+      if sync_version != :all && !self::SUPPORTS_SINGLE_VERSION_UPDATE
+        logger.warn("#{db_platform}.update(#{name}, sync_version: #{sync_version}) called but not supported on platform")
         return
       end
 
-      update(name, sync_versions: false)
       raw_project = project(name)
+      return false unless raw_project.present?
```
jsonperl, line 102: I'm looking at this all a bit cross-eyed and it feels like it could be broken down into smaller functions, but combining seems like a good step in the right direction before refactoring/extracting.

jsonperl (approved):

app/models/package_manager/base.rb
tiegz: 6 years ago 😬 https://github.com/librariesio/libraries.io/pull/58/files

jsonperl: Yea... Delete it :D

If it splodes, it splodes! 

https://media.giphy.com/media/lZhymdRsuFDmU/giphy-downsized.gif

jsonperl: Oh this is merged :D 

We'll delete it later!

PR #2740, opened 2021-02-18
 PackageManagerDownloadWorker refactor: step 3

tiegz:
TL;DR: this fetches the latest version for Maven packages from maven_metadata.xml, instead of fetching all the versions and picking the newest one.

We need to use `latest_version()` for `mapping()`, so we need the latest version AFAIK.

In the next step, we can make `versions()` optional, so we can get project data without making N requests for the versions.

Here's the original commit that started fetching the latest version to use as the project's POM: https://github.com/librariesio/libraries.io/commit/3b77f008b27cd05870b0bb534c1236e0a1213e9d#diff-ac475c62f3e3faba6f62b8831d6d8b85af275fdb812328fd9cfaec0179718e20R39

### DIFF
--- a/app/models/package_manager/base.rb
+++ b/app/models/package_manager/base.rb
@@ -121,8 +121,6 @@ module PackageManager
 
       save_dependencies(mapped_project) if self::HAS_DEPENDENCIES
       finalize_db_project(db_project)
-    rescue SystemExit, Interrupt
-      exit 0
     rescue StandardError => e
       if ENV["RACK_ENV"] == "production"
         Bugsnag.notify(e)
--- a/app/models/package_manager/maven.rb
+++ b/app/models/package_manager/maven.rb
@@ -75,9 +75,9 @@ module PackageManager
       path = sections.join("/")
 
       versions = versions(nil, name)
-      latest_version = latest_version(name)
+      latest = latest_version(name)
 
-      return {} unless latest_version.present?
+      return {} unless latest.present?
 
       {
         name: name,
@@ -85,7 +85,7 @@ module PackageManager
         group_id: sections[0],
         artifact_id: sections[1],
         versions: versions,
-        latest_version: latest_version,
+        latest_version: latest,
       }
     rescue StandardError
       {}
@@ -164,7 +164,7 @@ module PackageManager
       if project && project[:versions]
         return project[:versions]
       else
-        xml_metadata = get_raw(MavenUrl.from_name(name, repository_base, NAME_DELIMITER).maven_metadata)
+        xml_metadata = maven_metadata(name)
         xml_versions = Nokogiri::XML(xml_metadata).css("version").map(&:text)
         retrieve_versions(xml_versions.filter { |item| !item.ends_with?("-SNAPSHOT") }, name)
       end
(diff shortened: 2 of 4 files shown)


app/models/package_manager/maven.rb
```diff
@@ -230,20 +232,20 @@ def self.licenses(xml)
...
-        Project
-          .find_by(name: name, platform: formatted_name)
-          &.versions
-          &.max_by(&:published_at)
-          &.number
-      end
+    def self.latest_version(name)
+      maven_metadata = get_raw(MavenUrl.from_name(name, repository_base, NAME_DELIMITER).maven_metadata)
+      latest = Nokogiri::XML(maven_metadata)
+        .css("versioning > latest, versioning > release, metadata > version")
```
tiegz, line 238: I checked all the Maven providers' examples and `versioning > latest` looks reliable. I did find one weird example that had 1 version but no `versioning > latest`, so I'm providing two fallbacks that could be there too.

app/models/package_manager/maven.rb
```diff
@@ -230,20 +232,20 @@ def self.licenses(xml)
...
-        # TODO: this is in place to handle packages that are no longer on maven-repository.com
-        # this could be removed if we switched to a package data provider that supplied full information
-        Project
-          .find_by(name: name, platform: formatted_name)
-          &.versions
-          &.max_by(&:published_at)
-          &.number
-      end
+    def self.latest_version(name)
+      maven_metadata = get_raw(MavenUrl.from_name(name, repository_base, NAME_DELIMITER).maven_metadata)
```
tiegz, line 236: This fetches the maven metadata separately from `version()` (so we'll call it twice -- should be cheap, but is a good use-case for maintaining state in PackageManager::Base, where we could cache the maven metadata in the instance).

app/models/package_manager/base.rb
```diff
@@ -121,8 +121,6 @@ def self.update(name, sync_version: :all)
 
       save_dependencies(mapped_project) if self::HAS_DEPENDENCIES
       finalize_db_project(db_project)
-    rescue SystemExit, Interrupt
```
jsonperl, line 124: 👋

app/models/package_manager/maven.rb
jsonperl, line 236: seems useful to make a `.maven_metadata` method?

app/models/package_manager/maven.rb
```diff
@@ -230,20 +232,20 @@ def self.licenses(xml)
...
+      latest = Nokogiri::XML(maven_metadata)
+        .css("versioning > latest, versioning > release, metadata > version")
+        .map(&:text)
+        .first
+
+      # TODO: this is in place to handle packages that are no longer on maven-repository.com
```
jsonperl, line 242: We're not using maven-repository now right? Maybe just delete this, and if it fails we can explore. Also if we need it, that query below can be a single pluck instead of hydrating hundreds/thousands of version objects to grab a single version number.

jsonperl: `Version.where(project: Project.where(name: project.name, platform: project.platform)).order(published_at: :desc).limit(1).pluck(:number)`

app/models/package_manager/maven.rb
tiegz: True!

jsonperl (approved):

mellisio (approved):

PR #2742, opened 2021-02-19
PackageManagerDownloadWorker refactor: step 5

tiegz:
Sometimes we get packages with a list of dependencies that don't include versions (e.g. [this Maven one](https://repo1.maven.org/maven2/org/wso2/carbon/identity/framework/org.wso2.carbon.identity.authenticator.thrift/5.10.54/org.wso2.carbon.identity.authenticator.thrift-5.10.54.pom)). This skips the dependency creation if there's no version requirement bc there's a validation on `Dependency#requirements` anyway. This should save us a  ROLLBACK in these cases.

This also converts the save methods to `!` versions, so we don't silently fail like this anymore while updating packages.


app/models/package_manager/base.rb
```diff
@@ -204,12 +204,12 @@ def self.save_dependencies(mapped_project)
                end
 
         deps.each do |dep|
-          next if dep[:project_name].blank? || db_version.dependencies.any? { |d| d.project_name == dep[:project_name] }
+          next if dep[:project_name].blank? || dep[:requirements].blank? || db_version.dependencies.any? { |d| d.project_name == dep[:project_name] }
```
tiegz, line 207: 💥

app/models/package_manager/base.rb
```diff
@@ -187,10 +186,12 @@ def self.new_names
       names - existing_names
     end
 
-    def self.save_dependencies(mapped_project)
+    def self.save_dependencies(mapped_project, sync_version: :all)
       name = mapped_project[:name]
       db_project = Project.find_by(name: name, platform: db_platform)
-      db_project.versions.includes(:dependencies).each do |db_version|
+      db_versions = db_project.versions.includes(:dependencies)
+      db_versions = versions.where(id: sync_version) unless sync_version == :all
```
tiegz, line 193: 💥

jsonperl: 🎉

jsonperl (approved):
This is gonna be a massive win

PR #2746, opened 2021-02-22
Autocorrect Style/FrozenStringLiteralComment

tiegz:
Ty Rubocop


jsonperl (approved):

jsonperl (approved):
Couple tests are failing now `can't modify frozen String` which makes me slightly nervous about a deploy, though we'd probably get bug notifications pretty quickly 🤷‍♂️

PR #2751, opened 2021-02-24
Fallback to license_url if no license specified for nuget packages

jsonperl:


app/models/package_manager/nu_get.rb
```diff
@@ -111,10 +111,15 @@ def self.description(item)
 
     def self.versions(project, _name)
       project[:releases].map do |item|
+        license = [
```
jsonperl, line 114: This is the only meat of this PR

tiegz (approved):

spec/models/package_manager/nu_get_spec.rb
```diff
@@ -1,36 +1,63 @@
...
+  end
+
+  describe "with release data" do
+    def stub_releases(releases_data = {})
+      allow(described_class)
+        .to receive(:get_releases)
+        .and_return(releases_data)
+    end
+
+    # TODO: license urls are deprecated, but for now we'll fallback and include them as the license
```
mellisio, line 45: What do people tend to put in there? Is it like a link to MIT?

mellisio (approved):

spec/models/package_manager/nu_get_spec.rb
jsonperl: It's generally been ms shortlinks like http://go.microsoft.com/fwlink/?LinkId=329770

PR #2752, opened 2021-02-24
Removed load_names from maven package manager

jsonperl:
This isn't used, I deleted the set from redis.

### DIFF
--- a/app/models/package_manager/maven.rb
+++ b/app/models/package_manager/maven.rb
@@ -53,17 +53,12 @@ module PackageManager
       PROVIDER_MAP[source].check_status_url(project)
     end
 
-    def self.load_names(_limit = nil)
-      names = get("https://maven.libraries.io/mavenCentral/all")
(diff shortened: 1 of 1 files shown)


app/models/package_manager/maven.rb
```diff
@@ -182,8 +177,8 @@ def self.retrieve_versions(versions, name)
             published_at: Time.parse(pom.locate("publishedAt").first.text),
             original_license: license_list,
           }
-        rescue Ox::Error
-          next
+      rescue Ox::Error
+        next
```
tiegz, line 181: I think this `rescue` is for the map block itself.

jsonperl: I didn't do it, did rubocop break it??

jsonperl: Na, just goofy indenting. I don't wanna fight with him 👮‍♂️

tiegz (approved):

PR #2754, opened 2021-02-25
Quiet some PackageManagerDownloadWorker race condition noise.

tiegz:
This has been noisier since we started using `save!` methods instead of `save` during package updates.

Should silence https://app.bugsnag.com/tidelift/libraries-dot-io/errors/6036fff1d519f90017f7229b

### DIFF
--- a/app/models/package_manager/base.rb
+++ b/app/models/package_manager/base.rb
@@ -125,12 +125,6 @@ module PackageManager
 
       save_dependencies(mapped_project, sync_version: sync_version) if self::HAS_DEPENDENCIES
       finalize_db_project(db_project)
-    rescue StandardError => e
-      if ENV["RACK_ENV"] == "production"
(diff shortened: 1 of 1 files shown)


app/models/package_manager/base.rb
```diff
@@ -125,12 +125,6 @@ def self.update(name, sync_version: :all)
 
       save_dependencies(mapped_project, sync_version: sync_version) if self::HAS_DEPENDENCIES
       finalize_db_project(db_project)
-    rescue StandardError => e
-      if ENV["RACK_ENV"] == "production"
-        Bugsnag.notify(e)
-      else
-        raise
-      end
```
tiegz, line 133: I think this might've been added before it was run async? But should be fine to remove it AFAICT.

mellisio (approved):

PR #2756, opened 2021-02-26
Start using optimized .../dependencies endpoint code

tiegz:
I tested this API call and it seems to work, but I forgot to flip the switch.


tiegz:
Here are two other places we're doing this too:

https://github.com/tidelift/dependencyci/blob/main/app/models/user.rb#L192-L222
https://github.com/tidelift/dependencyci/blob/main/app/services/hubspot_service.rb#L81-L85

mellisio (approved):

jsonperl (approved):

PR #2757, opened 2021-02-27
PackageManager::CocoaPods: don't set 'versions' on project if the project is missing.

tiegz:
We're getting `{"versions" => {}}` for CocoaPods that don't exist, but we want just a blank `{}` so we can do an early-return because the project doesn't exist.

Fixes https://app.bugsnag.com/tidelift/libraries-dot-io/errors/6036a95d2447bb0018ea48f2


tiegz:
Sidenote: we've seen about 19k of these so far (and would've been getting them beforehand). Might be something weird going on with that Cocoapods service?

katzj (approved):

PR #2764, opened 2021-03-09
Fix for Elm ingestion.

tiegz:
Fixes https://app.bugsnag.com/tidelift/libraries-dot-io/errors/60010c298ed8fc0017c964c9

Also renamed a bunch of args to be more clear.

### DIFF
--- a/app/models/package_manager/elm.rb
+++ b/app/models/package_manager/elm.rb
@@ -39,7 +39,7 @@ module PackageManager
     end
 
     def self.project(name)
-      projects[name]
+      get("http://package.elm-lang.org/packages/#{name}/latest/elm.json")
     end
 
     def self.mapping(project)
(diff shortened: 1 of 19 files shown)


app/models/package_manager/elm.rb
```diff
@@ -56,25 +56,16 @@ def self.versions(project, _name)
       end
     end
 
-    def self.dependencies(name, version, project)
-      find_and_map_dependencies(name, version, project)
-    end
-
-    def self.find_dependencies(name, version)
-      url = "https://raw.githubusercontent.com/#{name}/#{version}/elm-package.json"
```
tiegz, line 64: This url doesn't work: https://raw.githubusercontent.com/elm/bytes/1.0.2/elm-package.json

app/models/package_manager/elm.rb
```diff
@@ -56,25 +56,16 @@ def self.versions(project, _name)
...
-      begin
-        response = request(url)
-        if response.status == 200
-          contents = response.body
-          dependencies = Bibliothecary.analyse_file("elm-package.json", contents).first.try(:fetch, :dependencies)
-          dependencies
-        else
-          []
+    def self.dependencies(name, version, _mapped_project)
+      get("http://package.elm-lang.org/packages/#{name}/#{version}/elm.json")
```
tiegz, line 60: This was the only JSON endpoint I could find that returns the data (it's the one that's used as an XHR to fetch the data on web)

jsonperl (approved):

PR #2779, opened 2021-04-08
Handle missing POMs 

tiegz:
Some libs in the Spring Artifactory instance (also some Maven libs) do not have POMs, even tho they have zips, pom.asc etc. For example:

https://repo.spring.io/webapp/#/artifacts/browse/tree/General/libs-release-local/org/springframework/cloud/spring-cloud-function-docs/3.1.2

This gracefully moves on during the update if it is missing.

https://app.clubhouse.io/tidelift/story/16019/pom-parsing-error-in-package-manager-download-worker

### DIFF
--- a/app/models/package_manager/maven.rb
+++ b/app/models/package_manager/maven.rb
@@ -25,6 +25,13 @@ module PackageManager
       "SpringLibs" => SpringLibs,
     }.freeze
 
+    class POMNotFound < StandardError
+      attr_reader :url
+      def initialize(url)
+        @url = url
+        super("POM not found: #{@url}")
+      end
+    end
 
     def self.repository_base
       PROVIDER_MAP["default"].repository_base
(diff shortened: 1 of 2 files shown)


app/models/package_manager/maven.rb
```diff
@@ -153,7 +163,10 @@ def self.retrieve_versions(versions, name)
     end
 
     def self.download_pom(group_id, artifact_id, version)
-      pom_request = request(MavenUrl.new(group_id, artifact_id, repository_base).pom(version))
+      url = MavenUrl.new(group_id, artifact_id, repository_base).pom(version)
+      pom_request = request(url)
+      raise POMNotFound.new(url) if pom_request.status == 404
```
jsonperl, line 168: Will we have any useful information if we can't get the pom?

tiegz: Unfortunately not. We can only rely on maven_metadata.xml for getting the list of versions, and we get the rest of the data in `mapping(raw_project)` from the latest_version's POM. 

I guess we could iterate thru each version and find the newest one with a POM, but that wouldn't work in this specific Spring example because none of the versions of a POM.

jsonperl (approved):

PR #2780, opened 2021-04-09
Skip version updates if POM is not found.

tiegz:
Followup to https://github.com/librariesio/libraries.io/pull/2779

When pulling specific versions and the POM is missing, skip the update on that version.

Fixes https://app.bugsnag.com/tidelift/libraries-dot-io/errors/606f904b075d7d00079ebe85?event_id=606f904b00770c3cbd710000&i=em&m=nw

### DIFF
--- a/spec/models/package_manager/maven_spec.rb
+++ b/spec/models/package_manager/maven_spec.rb
@@ -139,6 +139,16 @@ describe PackageManager::Maven do
     end
   end
 
+  describe ".one_version" do
+    it "retrieves a single version" do
+      allow(described_class)
+        .to receive(:download_pom)
+        .and_raise(PackageManager::Maven::POMNotFound.new("https://a-maven-central-url"))
+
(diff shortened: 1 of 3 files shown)


spec/models/package_manager/maven_spec.rb
```diff
@@ -139,6 +139,16 @@
     end
   end
 
+  describe ".one_version" do
+    it "retrieves a single version" do
+      allow(described_class)
+        .to receive(:download_pom)
+        .and_raise(PackageManager::Maven::POMNotFound.new("https://a-maven-central-url"))
+
+      expect(PackageManager::Maven::MavenCentral.one_version("org.foo:bar", "1.0.0")). to eq(nil)
```
tiegz, line 148: MavenCentral is the only Maven one that we allow single-version updates for now.

mellisio (approved):

PR #2793, opened 2021-04-23
Raise exception on version update failures so we can retry later

jsonperl:


tiegz (approved):
Nice, can't wait to see how often this blows up!

PR #2795, opened 2021-04-27
Less noise when we get a Packagist 404.

tiegz:
Fixes https://app.bugsnag.com/tidelift/libraries-dot-io/errors/6007059462591d001892e3a9?event_id=60871a7f00780ff373c30000&i=em&m=fq

### DIFF
--- a/app/models/package_manager/packagist.rb
+++ b/app/models/package_manager/packagist.rb
@@ -27,7 +27,8 @@ module PackageManager
     end
 
     def self.project(name)
-      get("https://packagist.org/packages/#{name}.json")["package"]
+      get("https://packagist.org/packages/#{name}.json")
+        &.fetch("package")
     end
 
     def self.deprecation_info(name)


app/models/package_manager/packagist.rb
```diff
@@ -27,7 +27,8 @@ def self.recent_names
     end
 
     def self.project(name)
-      get("https://packagist.org/packages/#{name}.json")["package"]
+      get("https://packagist.org/packages/#{name}.json")
```
jsonperl, line 30: Should this explode tho, and not silently? Maybe just a better error?

tiegz: For package updates, our convention is to [move on silently](https://github.com/librariesio/libraries.io/blob/main/app/models/package_manager/base.rb#L97) if the project is blank. I think it makes sense to just return a `blank?` value from this if the project is 404'ing, and let the "check_project_status" code mark it as "Removed" or whatever later on. I made a similar change [to Sublime](https://github.com/librariesio/libraries.io/blob/main/app/models/package_manager/sublime.rb#L18-L22) recently too.

jsonperl (approved):

PR #2797, opened 2021-04-27
Project tag updater

jsonperl:
Little baby-step to try and figure out what's going on here and optimize it. We collected a bunch of stats that definitely point to tag updating being the culprit here. The second commit has the meat, the first is just linting.

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -506,6 +506,15 @@ class Project < ApplicationRecord
     end
   end
 
+  def update_tags
+    return unless repository
+
+    repository.download_tags
+  rescue StandardError => e
+    Rails.logger.error("Version#update_tags #{e.inspect}")
+    nil
+  end
+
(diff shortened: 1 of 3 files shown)


jsonperl:
Added some more stats too in https://github.com/librariesio/libraries.io/pull/2797/commits/f0b434c3d15fc02d2f3cdfc197cce7774645d73c

app/models/project.rb
```diff
@@ -449,53 +461,75 @@ def normalize_licenses
...
+    end
+  end
+
+  def update_tags
+    return unless repository
+
+    benchmark = nil
+
+    qcount = QueryCounter.count do
+      benchmark = Benchmark.measure do
```
tiegz, line 517: Could also use this if you want to avoid the multiplication below:

```suggestion
      ms = Benchmark.ms do
```

tiegz (approved):

PR #2808, opened 2021-05-10
Fix tags, drop timeout to 1.5 seconds in webhook

mellisio:


app/models/tag.rb
```diff
@@ -45,16 +45,9 @@ def has_projects?
...
-      subscriptions = subscriptions.include_prereleases if prerelease?
-
-      subscriptions.group_by(&:notification_user).each do |user, _user_subscriptions|
-        next if user.nil?
-        next if user.muted?(project)
-        next if !user.emails_enabled?
-        VersionsMailer.new_version(user, project, self).deliver_later
-      end
+    project.mailing_list(include_prereleases: prerelease?).each do |user|
+      next if user.muted?(project)
```
jsonperl, line 49: Can you add this muted check in project.rb L652?

jsonperl, line 49: Aren't the project_mutes already checked in your updates in project.rb

mellisio: ah yes, I forgot I did that

jsonperl (approved):

PR #2819, opened 2021-05-24
bibliothecary 6.9.7 ~> 6.12.2

tiegz:
Looks like we haven't bumped bibliothecary in a few months, so here's a diff:

https://github.com/librariesio/bibliothecary/compare/v6.9.7...v6.12.2

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,7 +6,7 @@ ruby "2.6.5"
 gem "active_model_serializers"
 gem "api-pagination"
 gem "asciidoctor"
-gem "bibliothecary", "~> 6.12.2"
+gem "bibliothecary"
 gem "bitbucket_rest_api", git: "https://github.com/librariesio/bitbucket"
 gem "bootsnap", require: false
 gem "bootstrap-sass"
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -6,7 +6,7 @@ ruby "2.6.5"
 gem "active_model_serializers"
 gem "api-pagination"
 gem "asciidoctor"
-gem "bibliothecary"
+gem "bibliothecary", "~> 6.12.2"
```
jsonperl, line 9: Any reason to specify rather than just a bundle update.

jsonperl (approved):

Gemfile
tiegz: Strangely that wasn't updating the gem for me! I wonder if it's because it's a new minor version?

jsonperl: May have been that rubygems didn't have it yet?

tiegz: Wait a sec, I tried again and it worked. I bet I was just running `bundle` before 🤦🏻

jsonperl (approved):

PR #2821, opened 2021-05-25
Remove Atom, Jam, Shards and Sublime.

tiegz:
Records will remain in db for the time being, so I'm adding an explicit 404 when that happens so we don't get 500s instead.


app/controllers/application_controller.rb
```diff
@@ -90,6 +90,10 @@ def find_platform(param = :id)
 
   def find_project
     @project = Project.find_best!(params[:platform], params[:name], [:repository, :versions])
+
+    # There could be projects in the db whose package managers have since been removed
+    raise ActiveRecord::RecordNotFound unless @project.platform_class_exists?
```
jsonperl, line 95: 👍🏻

jsonperl (approved):

PR #2824, opened 2021-05-26
Search: exclude platforms removed from code but still exist in db/ES.

tiegz:
Some records are now throwing 500s in search because they exist in the db/ES but their PackageManager::Base platform was removed. Until we remove those records in a cleanup, this ignores those removed platforms in all project searches explicitly. This should also exclude those removed platforms from faceting too 👍🏻 

Fixes https://app.bugsnag.com/tidelift/libraries-dot-io/errors/60ae50aca940100007305441

### DIFF
--- a/app/models/concerns/project_search.rb
+++ b/app/models/concerns/project_search.rb
@@ -1,5 +1,11 @@
 # frozen_string_literal: true
 module ProjectSearch
+  # Whenever a PackageManager::Base platform is removed, it's easier for now to
+  # just add it to this list, until the records/indices are cleared out of that platform.
+  # NB these are the casings from the database, which sometimes don't match the PackageManager::Base
+  # formatted_name casings, e.g. Pypi vs PyPI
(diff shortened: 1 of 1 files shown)


app/models/concerns/project_search.rb
```diff
@@ -1,5 +1,11 @@
 # frozen_string_literal: true
 module ProjectSearch
+  # Whenever a PackageManager::Base platform is removed, it's easier for now to
+  # just add it to this list, until the records/indices are cleared out of that platform.
+  # NB these are the casings from the database, which sometimes don't match the PackageManager::Base
+  # formatted_name casings, e.g. Pypi vs PyPI
```
tiegz, line 6: Ideally we could just safelist all platforms from `PackageManager::Base.platforms.map(&:formatted_name)` but they don't always map 1:1 with `projects.platform` in the database, so we can keep this explicit blocklist array here instead.

jsonperl (approved):

PR #2838, opened 2021-07-02
Updated readme

jhan217:
Updated readme with a couple things I discovered while setting up the project on an Apple M1-chipped computer + Dockerized Postgres instance.

### DIFF
--- a/docs/development-setup.md
+++ b/docs/development-setup.md
@@ -17,22 +17,46 @@ rbenv install 2.6.5
 Next, you'll need to make sure that you have PostgreSQL, Elasticsearch 2.4 and Redis installed. This can be done easily on OSX using [Homebrew](http://mxcl.github.io/homebrew/) or postgres can be installed by using [http://postgresapp.com](http://postgresapp.com). Please also see these [further instructions for installing Postgres via Homebrew](http://www.mikeball.us/blog/setting-up-postgres-with-homebrew/).
 
 ```bash
-brew cask install phantomjs homebrew/cask-versions/adoptopenjdk8
-brew install postgres elasticsearch@2.4 redis icu4c cmake
+brew install --cask phantomjs homebrew/cask-versions/adoptopenjdk8
+brew install postgres redis icu4c cmake
 ```
+
+Since this repo uses an old version of Elasticsearch, it is no longer supported on Homebrew. You can use the ElasticSearch
+2.4.5 container from Dockerhub, but the image won't work on Apple M1 chipped machines. You can download the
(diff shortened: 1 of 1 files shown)


docs/development-setup.md
```diff
@@ -17,24 +17,48 @@ rbenv install 2.6.5
...
+to start up ElasticSearch. When developing in the repo and you encounter Faraday timeouts on updates, you may need
+to restart your local ElasticSearch cluster.
+
 On Debian-based Linux distributions you can use apt-get to install Postgres:
 
 ```bash
 sudo apt-get install postgresql postgresql-contrib libpq-dev libicu-dev
 ```
 
+If you are using a Docker container for Postgres, you'll need to edit the `config/database.yml`
```
jsonperl, line 46: Can we just edit the database.yml so it'll work in all cases?

jhan217: I'm not sure what editing the `database.yml` will do for the non-dockerized postgres instances. It would also probably affect people who already have a working non-docker local environment (with `postgres` as the user instead of `librariesio`).

jsonperl: Would this work?
```
  database: libraries_development
  username: <%= ENV["PG_USERNAME"] || "postgres" %>
  password: <%= ENV["PG_PASSWORD"] %>
  host: <%= ENV["PG_HOST"] %> || "/var/run/postgres"
```

jsonperl: I just worry local changes will accidentally keep getting committed since it's source controlled, and maybe this lets us run both ways with some env settings.

jsonperl: Hangon a sec, I'm digging into the socket setup

jhan217: I'm happy to make the change but I think either way someone's going to have a local changes due to the different users/environment variable names specified in the docker-compose.yml and database.yml files. 

https://github.com/librariesio/libraries.io/blob/main/config/database.yml#L27
https://github.com/librariesio/libraries.io/blob/main/docker-compose.yml#L19-L20

Updating the README / letting people configure things as they want seems to be the least amount of risk in terms of breaking existing setups (esp. since I'm not familiar with all the non-M1/Docker setups).

jsonperl: Ok updated to use the default socket location from homebrew. Works for me with a homebrew install, if it works for your Docker postgres, we're probably good.

jsonperl (approved):

jsonperl (approved):

jsonperl (approved):

docs/development-setup.md
jhan217: We decided to not update `database.yml` since the test configs would either work for the local development environment or circle, but not both.

PR #2840, opened 2021-07-08
Add epl2 url to LICENSE_STRINGS

jhan217:
This is a quick fix to resolve the missing license information for the org.jacoco:org.jacoco.* packages. I was able to get the EPL2 license information locally with this change. 

See https://app.clubhouse.io/tidelift/story/18585/license-missing-for-maven-org-jacoco-org-jacoco-packages#activity-18662 for more details.

### DIFF
--- a/app/models/package_manager/maven.rb
+++ b/app/models/package_manager/maven.rb
@@ -157,6 +157,12 @@ module PackageManager
           rescue StandardError
             license_list = nil
           end
+
+          if license_list.blank? && pom.respond_to?("project") && pom.project.locate("parent").present?
+            parent_version = extract_pom_value(pom.project, "parent/version")
+            Rails.logger.info("[POM has parent no license] name=#{name} parent_version=#{parent_version} child_version=#{version}")
+          end
+
           {
             number: version,
             published_at: Time.parse(pom.locate("publishedAt").first.text),
--- a/app/models/package_manager/maven.rb
+++ b/app/models/package_manager/maven.rb
@@ -13,6 +13,7 @@ module PackageManager
     LICENSE_STRINGS = {
       "http://www.apache.org/licenses/LICENSE-2.0" => "Apache-2.0",
       "http://www.eclipse.org/legal/epl-v10" => "Eclipse Public License (EPL), Version 1.0",
+      "http://www.eclipse.org/legal/epl-2.0" => "Eclipse Public License (EPL), Version 2.0",
       "http://www.eclipse.org/org/documents/edl-v10" => "Eclipse Distribution License (EDL), Version 1.0",
     }.freeze
     NAME_DELIMITER = ":"
@@ -29,6 +30,7 @@ module PackageManager
 
     class POMNotFound < StandardError
       attr_reader :url
+
       def initialize(url)
         @url = url
         super("Missing POM: #{@url}")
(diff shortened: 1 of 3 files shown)


app/models/package_manager/maven.rb
```diff
@@ -167,7 +171,7 @@ def self.retrieve_versions(versions, name)
     def self.download_pom(group_id, artifact_id, version)
       url = MavenUrl.new(group_id, artifact_id, repository_base).pom(version)
       pom_request = request(url)
-      raise POMNotFound.new(url) if pom_request.status == 404
+      raise POMNotFound, url if pom_request.status == 404
```
jhan217, line 174: This was a rubocop autocorrect

app/models/package_manager/maven.rb
```diff
@@ -145,6 +147,8 @@ def self.versions(raw_project, name)
     end
 
     def self.retrieve_versions(versions, name)
+      # TODO: This should also look for the parent POM and if it exists + the child POM
```
jhan217, line 150: Wondering if it's worth logging the number of version packages with parents so we get a sense of what that volume looks like / if we can pick off any additional missing licenses with this type of fix.

jsonperl: I think if there's a quick way to log something, go for it! Maybe something along the lines of 
`Rails.logger.info("[POM Parent Found] name=#{name} version=#{version}")`

app/models/package_manager/maven.rb
jsonperl: Yuck! :D

jsonperl (approved):

spec/fixtures/proto-google-common-protos-0.1.9.pom
```diff
@@ -53,3 +53,4 @@
     </dependency>
   </dependencies>
 </project>
+<publishedAt>Fri, 11 Oct 2019 20:00:00 GMT</publishedAt>
```
jhan217, line 56: From what I can tell this was stubbed out in the rest of the specs and doesn't show up on the POM files on maven (e.g. https://search.maven.org/artifact/org.jacoco/org.jacoco.core/0.8.7/pom and https://search.maven.org/artifact/org.jacoco/org.jacoco.build/0.8.7/pom), but it does show up in the xml we pull in.

spec/models/package_manager/maven_spec.rb
```diff
@@ -108,7 +108,7 @@
       it "should return nil" do
         allow(described_class).to receive(:download_pom).and_raise(PackageManager::Maven::POMNotFound.new("https://a-spring-url"))
 
-        expect(described_class.mapping({group_id: "org", artifact_id: "foo", version: "1.0.0"})).to eq(nil)
```
jhan217, line 111: Rubocop auto-corrected a bunch of whitespaces, the only major changes here were adding the specs on https://github.com/librariesio/libraries.io/pull/2840/files#diff-06c2fce0a53ffddef834b0f150c296f813f1c482f19ad291130a115870ebebf7R205

app/models/package_manager/maven.rb
```diff
@@ -153,6 +157,12 @@ def self.retrieve_versions(versions, name)
           rescue StandardError
             license_list = nil
           end
+
+          if license_list.blank? && pom.respond_to?("project") && pom.project.locate("parent").present?
```
jhan217, line 161: Not sure if I should wrap this in a `begin/rescue` block so it doesn't affect actual license fetching in case there's a POM parsing issue. Mostly hoping to quickly see from the logs if there are cases where the child version is off from the listed parent pom version.

jsonperl (approved):

PR #2842, opened 2021-07-13
Add new flash

katzj:


app/views/shared/flashes/upstream-webinar-1.html.erb
```diff
@@ -0,0 +1,8 @@
+Live webinar: How is Bloomberg supporting the Python community by shifting left? Join us 8/4.
```
jhan217, line 1: Worth adding the year to that date?

app/views/shared/flashes/upstream-webinar-1.html.erb
```diff
@@ -0,0 +1,8 @@
+Live webinar: How is Bloomberg supporting the Python community by shifting left? Join us 8/4. 
+<strong>
+  <a href="https://upstream.live/webinar/supporting-the-python-community-by-shifting-left?utm_source=libraries&utm_medium=referral&utm_campaign=upstream-webinar"
+    target="_blank"
```
jhan217, line 4: Do we control the `https://upstream.live` domain or should we add the `rel="noopener noreferrer"`/`rel="external"` tags?

katzj: We do

app/views/shared/flashes/upstream-webinar-1.html.erb
katzj: We swap them out before it matters (I just get the content from Amy :)

jhan217 (approved):

Issue #2865, opened 2021-10-02
One of my packages is being incorrectly listed as deprecated


serhatcelik:
My "avsub" package is being incorrectly listed as deprecated. I have previously yanked a specific version with bugs. It seems like the package being deprecated forever.

tiegz:
Thanks @serhatcelik , we've just fixed your package on Libraries.

PR #2866, opened 2021-10-04
Re-check Pypi packages for deprecated/removed status regularly.

tiegz:
* Removes the `removed` arg in CheckStatusWorker and CheckRepoStatusWorker (I don't see a reason why they're necessary, based on the orignial commit https://github.com/librariesio/libraries.io/commit/926f18c56ac8499a980dbbbb1bb64c5808db7113#diff-2bba7f75aa1e0fe9dac0d85c3619999931b5fe2a93918e2bdb0abdb2e1f951fdL8)
* Resets `status: 'Deprecated'` to `status: nil` if a project no longer deprecated
* Check pypi when rechecking removed/deprecated packages too

This should prevent cases like https://github.com/librariesio/libraries.io/issues/2865 where Pypi packages weren't considered deprecated anymore but didn't get reset back to `status: nil`.

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -109,6 +109,7 @@ class Project < ApplicationRecord
   scope :removed, -> { where('projects."status" = ?', "Removed") }
   scope :unmaintained, -> { where('projects."status" = ?', "Unmaintained") }
   scope :hidden, -> { where('projects."status" = ?', "Hidden") }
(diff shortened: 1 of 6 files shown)


tiegz:
Just remembered the reason for the `removed` args was probably (?) because there's double-syncing on Project#status from both Project#check_status and Repository#check_status, so I'm leaving this as is for now.

PR #2867, opened 2021-10-04
When re-checking removed projects, also check deprecated.

tiegz:
This should help cases where a project is no longer deprecated, e.g. https://github.com/librariesio/libraries.io/issues/2865

### DIFF
--- a/lib/tasks/projects.rake
+++ b/lib/tasks/projects.rake
@@ -53,18 +53,19 @@ namespace :projects do
     end
   end
 
-  desc 'Check to see if projects have been removed'
+  desc 'Check to see if projects are still removed/deprecated'
   task check_removed_status: :environment do
     exit if ENV['READ_ONLY'].present?
-    ['npm', 'rubygems', 'packagist', 'cpan', 'clojars', 'cocoapods',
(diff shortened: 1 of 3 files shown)


lib/tasks/projects.rake
```diff
@@ -53,18 +53,19 @@ namespace :projects do
     end
   end
 
-  desc 'Check to see if projects have been removed'
+  desc 'Check to see if projects are still removed/deprecated'
   task check_removed_status: :environment do
     exit if ENV['READ_ONLY'].present?
-    ['npm', 'rubygems', 'packagist', 'cpan', 'clojars', 'cocoapods',
+    # Check if removed/deprecated projects are still deprecated/removed
+    ['pypi', 'npm', 'rubygems', 'packagist', 'cpan', 'clojars', 'cocoapods',
```
tiegz, line 60: NB we didn't have pypi in here yet, so we were never re-checking their statuses.

katzj (approved):

tyrelsouza (approved):

lib/tasks/projects.rake
tiegz, line 60: NB we didn't have pypi in here yet, so we were never re-checking their statuses.

PR #2903, opened 2022-02-11
Fix API tree endpoint by changin params[:version] to params[:number]

tiegz:
The two endpoints to get a dep tree use different params for version:

```
                api_version_tree GET      /api/:platform/:name/:version/tree(.:format)              api/tree#show {:format=>:json, :platform=>/[\w\-]+/, :name=>/[^\/]+/, :version=>/[\w.\-]+/}
                    version_tree GET      /:platform/:name/:number/tree(.:format)                   tree#show {:number=>/[\w.\-%]+/, :name=>/[^\/]+/}
```

~~This adds the `params[:version]` to the tree-resolver-initialization logic so the first route can fetch the specific version instead of "latest stable" version.~~

This changes `params[:version]` to `params[:number]` to fix the API tree endpoint and make API routes consistent with Web routes.

### DIFF
--- a/app/controllers/api/projects_controller.rb
+++ b/app/controllers/api/projects_controller.rb
@@ -70,7 +70,7 @@ class Api::ProjectsController < Api::ApplicationController
     @subset = params.fetch(:subset, "default")
 
     @project = Project.find_best!(params[:platform], params[:name], [:repository, :versions])
-    @version = @project.find_version!(params[:version])
+    @version = @project.find_version!(params[:number])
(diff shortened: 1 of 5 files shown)


johnbintz-tidelift:
What would be the overhead w/ changing the route param name?

tiegz:
> What would be the overhead w/ changing the route param name?

Looked pretty simple, so I just did that (and api::projects#dependencies as well to be consistent)

havocp (approved):

PR #2908, opened 2022-03-03
Add version scheme detection task

papiro:

- [X] Have you followed the guidelines for [contributors](http://docs.libraries.io/contributorshandbook)?
- [X ] Have you checked to ensure there aren't other open pull requests on the repository for a similar change?
- [ X] Have you written new tests for your changes?
- [ X] Have you successfully run the project with your changes locally?

In investigating supporting multiple versioning schemes the idea arose to gather some numbers on scheme counts "in the wild".

This PR contains a Rake task.
The purpose of this task is to try to detect (as intelligently as possible) the spread of versioning schemes in use in a list of provided platforms and package names. It is not finished since I am working on consuming the package list from a file right now, but I wanted to submit for review of the algorithm.

### DIFF
--- a/lib/tasks/version_scheme_counter.rake
+++ b/lib/tasks/version_scheme_counter.rake
@@ -53,7 +53,7 @@ namespace :version do
 
     package_platforms, package_names = packages[0].zip(*packages)
 
-    Project.where(platform: package_platforms, name: package_names).find_each(batch_size: 1000) do |project|
+    Project.includes(:versions).where(platform: package_platforms, name: package_names).find_each(batch_size: 1000) do |project|
       local_tallies = tallies.clone
 
       if !project.versions_count?
--- a/lib/tasks/version_scheme_counter.rake
+++ b/lib/tasks/version_scheme_counter.rake
@@ -44,6 +44,23 @@ module VersionSchemeDetection
       first_part.length >= 4 || (first_part.length == 2 && first_part.to_i > 0)
     end
   }
+
+  def self.build_project_where_clause(packages)
+    # downcase platform
+    packages.map! { |package| [package[0].downcase, package[1]] }
+
+    project_table = Project.arel_table
+
+    packages.reduce(Arel::Nodes::False.new) do |clause, package|
+      clause.or(
+        project_table.grouping(
+          project_table.lower(project_table[:platform]).eq(package[0]).and(
+            project_table[:name].eq(package[1])
+          )
+        )
+      )
+    end
+  end
 end
 
 tallies = { semver: 0, pep440: 0, maven: 0, osgi: 0, calver: 0, unknown: 0, no_versions: 0 }
@@ -59,17 +76,11 @@ namespace :version do
     warnings = []
 
     packages = CSV.read(args[:package_list])
-    # downcase platform
-    packages.map! { |package| [package[0].downcase, package[1]] }
-
-    # The following makes a query like "(x AND y) OR (x AND y)"
-    package_matcher = "(LOWER(platform) = ? AND name = ?)"
-    where_clause = "#{package_matcher} OR " * (packages.length - 1)
-    where_clause = "#{where_clause} #{package_matcher}"
 
     Project
       .includes(:versions)
-      .where([where_clause, *packages.flatten]).find_each(batch_size: 1000) do |project|
+      .where(VersionSchemeDetection.build_project_where_clause(packages))
+      .find_each(batch_size: 1000) do |project|
       project_platform = project.platform
       project_name = project.name
 
--- a/lib/tasks/version_scheme_counter.rake
+++ b/lib/tasks/version_scheme_counter.rake
@@ -138,6 +138,9 @@ namespace :version do
     end if global_tallies[:unknown_schemes].length
     puts warnings
 
-    File.write(File.join(__dir__, "output", "version_scheme_count.json"), JSON.pretty_generate(global_tallies))
+    File.write(File.join(__dir__, "output", "version_scheme_count.json"), JSON.pretty_generate({
+                                                                                                 **global_tallies,
+                                                                                                 warnings: warnings
+                                                                                               }))
   end
 end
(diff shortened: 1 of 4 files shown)


lib/tasks/version_scheme_counter.rake
```diff
@@ -0,0 +1,108 @@
...
+        global_tallies[:unknown_versions].push([project.platform, project.name, project.versions.pluck(:number)])
+        detected_scheme = :unknown
+      elsif maxes.include?(:semver)
+        detected_scheme = :semver
+      elsif maxes.include?(:calver)
+        detected_scheme = :calver
+      elsif maxes.include?(:maven) && project.platform.downcase == "maven"
+        detected_scheme = :maven
+      elsif maxes.include?(:pep440) && project.platform.downcase == "pypi"
+        detected_scheme = :pep440
```
papiro, line 95: Need to determine a fall-through tie-breaker

lib/tasks/version_scheme_counter.rake
```diff
@@ -0,0 +1,108 @@
+# frozen_string_literal: true
+#
+# [
+#   ["Maven", "io.redlink.geocoding:proxy-commons"],
+#   ["Packagist", "phraseanet/php-sdk"]
+# ].each { |project| PackageManager.const_get(project[0]).update project[1]}
+#
+#
+# Version scheme detection methodology:
```
papiro, line 9: The algorithm in plain english

lib/tasks/version_scheme_counter.rake
```diff
@@ -0,0 +1,108 @@
...
+          :semver,
+          :pep440,
+          :maven,
+          :calver,
+        ].each do |scheme|
+          if VersionSchemeDetection::VALIDATORS[scheme.upcase].call(version_number)
+            local_tallies[scheme] += 1
+            matchable = true
+          end
+        end
```
tiegz, line 77: Just to 2x check, this doesn't break on purpose so that we catch all matching schemes for a version? Makes sense, to get the best approximation

lib/tasks/version_scheme_counter.rake
```diff
@@ -0,0 +1,108 @@
...
+namespace :version do
+  desc 'Tests a sampling of project\'s versions to count occurrences of different versioning schemes'
+  task scheme_counter: :environment do
+    # projects = Project.order(Arel.sql("RANDOM()")).limit(100)
+
+    global_tallies = tallies.clone.merge({
+      unknown_versions: []
+    })
+
+    Project.find_each(batch_size: 1000) do |project|
```
tiegz: Could do an `includes(:versions)` here to preload

lib/tasks/version_scheme_counter.rake
```diff
@@ -0,0 +1,108 @@
...
+        end
+      end
+
+      max_count = local_tallies.values.max
+      maxes = local_tallies.select{|_, tally| tally == max_count}.keys
+
+      if maxes.length == 1
+        detected_scheme = maxes[0]
+      elsif maxes.include?(:unknown)
+        global_tallies[:unknown_versions].push([project.platform, project.name, project.versions.pluck(:number)])
```
tiegz, line 90: If you do the `includes(:versions)` suggestion from above, could change the pluck to a map here too

lib/tasks/version_scheme_counter.rake
tiegz: Just to 2x here too, the ordering basically determines the priority that we give a certain scheme match, in case there are 2+ schemes that match the most?

papiro: yes, that's correct

lib/tasks/version_scheme_counter.rake
papiro: yep, that's right

lib/tasks/version_scheme_counter.rake
papiro: What is the difference between the two? I've seen `.map(&:number)` before but I thought it was only needed when referring to a non-getter method on the object?

tiegz: The `pluck()` does a new query for the column you need (`SELECT number FROM versions WHERE versions.package_id = ?`) and doesn't hydrate any ActiveRecord models, so it's just a list of values.

So if you don't do the `includes(:versions)` earlier, it won't have any preloaded versions and this'll be the most efficient way.

But if you `includes(:versions)` earlier, it'll preload your versions (and hydrate their ActiveRecord models) with all the numbers in a single query, and it'll already have the `:number` values in memory, so doing a `pluck(:number)` query for each project would be extraneous work.

lib/tasks/version_scheme_counter.rake
```diff
@@ -0,0 +1,135 @@
...
+      unknown_schemes: []
+    })
+    warnings = []
+
+    packages = CSV.read(args[:package_list])
+    # downcase platform
+    packages.map! { |package| [package[0].downcase, package[1]] }
+
+    # The following makes a query like "(x AND y) OR (x AND y)"
+    package_matcher = "(LOWER(platform) = ? AND name = ?)"
```
johnbintz-tidelift, line 66: There's a more programmatic and less string-oriented way you can build these SQL statements. We use Arel nodes on the Catalog model in DepCI to build a complex grouping of `OR` queries that is a bit easier to work with: https://github.com/tidelift/dependencyci/blob/main/app/models/catalog.rb#L634-L645

johnbintz-tidelift (approved):
This looks good! I'll be curious to see what comes out of the analysis.

lib/tasks/version_scheme_counter.rake
```diff
@@ -0,0 +1,146 @@
...
+    Project
+      .includes(:versions)
+      .where(VersionSchemeDetection.build_project_where_clause(packages))
+      .find_each(batch_size: 1000) do |project|
+      project_platform = project.platform
+      project_name = project.name
+
+      local_tallies = tallies.clone
+
+      if !project.versions_count?
```
johnbintz-tidelift, line 89: I am surprised that linting tools didn't complain about this one:

```suggestion
      unless project.versions_count?
```

PR #2914, opened 2022-03-11
Group Go Projects Around Repository URL to Avoid Different Cased Modules

mikeyoung85:
This will try and lookup existing Go projects that match based on the repository URL and use the existing name instead of creating a new one. We have a lot of Go projects that have different casing but all point to the same module.

One example:
https://pkg.go.dev/github.com/Go-sqL-Driver/MySQL
https://pkg.go.dev/github.com/Go-SQL-Driver/MySQL

### DIFF
--- a/app/models/package_manager/go.rb
+++ b/app/models/package_manager/go.rb
@@ -77,7 +77,8 @@ module PackageManager
       project = super(name, sync_version: sync_version)
       # call update on base module name if the name is appended with major version
       # example: github.com/myexample/modulename/v2
-      update_base_module(name) if name.match(VERSION_MODULE_REGEX)
+      # use the returned project name in case it finds a Project via repository_url
+      update_base_module(project.name) if project.name.match(VERSION_MODULE_REGEX)
 
       project
     end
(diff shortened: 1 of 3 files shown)


app/models/package_manager/go.rb
```diff
@@ -142,8 +143,26 @@ def self.mapping(raw_project)
...
         url = raw_project[:overview_html]&.css(".UnitMeta-repo a")&.first&.attribute("href")&.value
 
+        # find an existing project with the same repository and replace the name with the existing project
+        # this will avoid creating duplicate Projects with various casing
+
+        # if this is a versioned module, make sure to find the right versioned project
+        regex_matches = raw_project[:name].match(VERSION_MODULE_REGEX)
+        if regex_matches
+          # try and find a versioned name matching this repository_url
+          existing_project_name = Project.where(platform: "Go").where("lower(repository_url) = :repo_url and name like :name", repo_url: url.downcase, name: "%/#{regex_matches[2]}").first&.name
```
tiegz, line 153: Was there an advantage to downcasing the repo url? I think pkg.go.dev should give us the canonical casing. 

One of my concerns with this whole thing too is that there's a git repository out there that might be case-sensitive and treat `repo.com/foo/bar` as a different repo than `repo.com/FOO/BAR`.

tiegz (approved):

PR #2918, opened 2022-03-21
Sort Releases By Publish Date

mikeyoung85:
This would set the latest release and tags for Project's using their publish date instead of by their ID order.

### DIFF
--- a/db/migrate/20220318153151_add_published_at_index_to_version.rb
+++ b/db/migrate/20220318153151_add_published_at_index_to_version.rb
@@ -1,0 +1,5 @@
+class AddPublishedAtIndexToVersion < ActiveRecord::Migration[5.2]
+  def change
+    add_index :versions, [:project_id, :published_at]
+  end
+end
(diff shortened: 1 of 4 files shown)


db/migrate/20220318153151_add_published_at_index_to_version.rb
```diff
@@ -0,0 +1,5 @@
+class AddPublishedAtIndexToVersion < ActiveRecord::Migration[5.2]
+  def change
+    add_index :versions, [:project_id, :published_at]
```
mikeyoung85, line 3: I'm actually not sure if this index is worth it. The queries should use the existing index by `project_id` and then sort from those results, which means it should only be sorting a few hundred at the most.

spec/models/project_spec.rb
```diff
@@ -348,4 +349,46 @@ def create_repo_sub(user)
       expect(project.mailing_list.count).to eq 1
     end
   end
+
+  describe "latest_release" do
```
mikeyoung85, line 353: Most of the other changes are rubocop. These are added specs.

havocp (approved):
Makes sense to me. Does this normally not really matter because we create releases in the same order that they are published and then the ids sort properly? But I guess can matter when for some reason we create them in a different order?

djpowers (approved):
LGTM

db/migrate/20220318153151_add_published_at_index_to_version.rb
katzj: Not sure if it is worth it either. But if we add, needs to be done with `algorithm: concurrently` to avoid downtime

PR #2919, opened 2022-03-22
version scheme detection recovery from pre-existing output file

papiro:


lib/tasks/version_scheme_counter.rake
```diff
@@ -45,6 +45,8 @@ module VersionSchemeDetection
     end
   }
 
+  TALLIES = { semver: 0, pep440: 0, maven: 0, osgi: 0, calver: 0, unknown: 0, no_versions: 0, total: 0 }
```
johnbintz-tidelift, line 48: To be extra safe when working with this constant, should this Hash be frozen?

```suggestion
  TALLIES = { semver: 0, pep440: 0, maven: 0, osgi: 0, calver: 0, unknown: 0, no_versions: 0, total: 0 }.freeze
```

johnbintz-tidelift (approved):

tiegz (approved):

lib/tasks/version_scheme_counter.rake
papiro: That's a good point, may as well.

PR #2927, opened 2022-04-02
Ignore versions with a missing project in Versions#index

tiegz:
Fixes https://app.bugsnag.com/tidelift/libraries-dot-io/errors/6234d918bcf6b60009a6cd53

### DIFF
--- a/app/controllers/api/versions_controller.rb
+++ b/app/controllers/api/versions_controller.rb
@@ -8,9 +8,12 @@ class Api::VersionsController < Api::BulkProjectController
   def index
     @max_results = [Integer(params[:max_results] || MAX_RESULTS), MAX_RESULTS].min
 
-    @versions = Version.includes(:project).where(
-      "versions.updated_at > ?",
-      Time.parse(params.require(:since))
-    ).order(:updated_at)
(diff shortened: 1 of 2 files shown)


app/controllers/api/versions_controller.rb
```diff
@@ -8,9 +8,12 @@ class Api::VersionsController < Api::BulkProjectController
   def index
     @max_results = [Integer(params[:max_results] || MAX_RESULTS), MAX_RESULTS].min
 
-    @versions = Version.includes(:project).where(
-      "versions.updated_at > ?",
-      Time.parse(params.require(:since))
-    ).order(:updated_at)
+    @versions = Version
+      .includes(:project)
```
mikeyoung85, line 12: Should this be a join? I have a feeling the includes turns the query into a join anyways...

mikeyoung85 (approved):

app/controllers/api/versions_controller.rb
tiegz: Yeah, once you start adding conditions to the includes it'll use a join instead 👍🏻

PR #2930, opened 2022-04-04
Set a reasonable timeout when looking up canonical name of go module

tiegz:
This adds a 2-second timeout to `PackageManager::Go.project_find_names()`, which is currently timing out (60 sec?) on a module that has been moved (`vbom.ml/util` -> `github.com/fvbommel/util`).

### DIFF
--- a/app/models/package_manager/go.rb
+++ b/app/models/package_manager/go.rb
@@ -205,7 +205,8 @@ module PackageManager
       return [name] if KNOWN_VCS.any?(&name.method(:include?))
 
       begin
-        go_import = get_html("https://#{name}?go-get=1")
+        # https://go.dev/ref/mod#serving-from-proxy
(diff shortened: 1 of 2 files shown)


app/models/package_manager/go.rb
```diff
@@ -205,7 +205,8 @@ def self.project_find_names(name)
       return [name] if KNOWN_VCS.any?(&name.method(:include?))
 
       begin
-        go_import = get_html("https://#{name}?go-get=1")
+        # https://go.dev/ref/mod#serving-from-proxy
+        go_import = get_html("https://#{name}?go-get=1", {request: {timeout: 2}})
```
tiegz, line 209: This was 3 years old so I looked for a better way to find this name, but couldn't find one 🤷🏻 This is at least part of the spec on how to define  a module host+name

PR #2932, opened 2022-04-04
Wrap Project#normalize_licenses in a few places to fix nil license bug in view

tiegz:
Fixes https://app.bugsnag.com/tidelift/libraries-dot-io/errors/5ffca366b35b870018f87a07

### DIFF
--- a/app/views/tree/_dep.html.erb
+++ b/app/views/tree/_dep.html.erb
@@ -9,7 +9,7 @@
   <em class='text-muted'>
     <span class="tip" title="Specified version range"><%= requirements %></span>
     <%= '-' if requirements.present? %>
-    <%= project.normalize_licenses.join(', ') %>
+    <%= project.normalized_licenses.join(', ') %>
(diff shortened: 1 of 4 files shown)


app/views/tree/_dep.html.erb
```diff
@@ -9,7 +9,7 @@
   <em class='text-muted'>
     <span class="tip" title="Specified version range"><%= requirements %></span>
     <%= '-' if requirements.present? %>
-    <%= project.normalize_licenses.join(', ') %>
+    <%= project.normalized_licenses.join(', ') %>
```
tiegz, line 12: @katzj @havocp do you know the background on `normalized_licenses` vs `license_normalized`?

tiegz: (the latter is set by `normalize_licenses`, so I'm confused by the singular/plural diff there)

papiro (approved):

PR #2937, opened 2022-05-25
Drupal: scrape each release page for published_at instead of releases index.

tiegz:
Drupal removed the "submitted on" date from their index of releases, so we have to fetch each release page separately to fetch the real published_at date.


Brad-- (commented):
lgtm!
image

Brad-- (approved):

PR #2941, opened 2022-05-31
Update packagist package endpoint to v2 to get fresh version data.

tiegz:
We've found that some Packagist package versions aren't being updated, and it turns out that the V1 package endpoint has been deprecated: https://blog.packagist.com/deprecating-composer-1-support/

This PR starts using the V2 package endpoint, which has a slightly different shape: https://packagist.org/apidoc#get-package-data .


app/models/package_manager/packagist.rb
```diff
@@ -52,8 +52,10 @@ def self.recent_names
     end
 
     def self.project(name)
-      get("https://packagist.org/packages/#{name}.json")
-        &.fetch("package")
+      # The main v2 endpoint excludes dev versions, so if that list
+      # of versions is empty, fallback to the dev list of versions.
+      get("https://repo.packagist.org/p2/#{name}.json").dig("packages", name).presence ||
```
mikeyoung85, line 57: Do we need to encode the name in here or is it already coming in encoded?

tiegz: It looks like we don't do any URI escaping for any of the PackageManager::Base subclasses (except for a one-off slash-escaping in NPM). Bad URIs would get raised as `URI::InvalidURIError`, and we do have a small flow of those, but it doesn't seem like a big issue (ruby URI parsing vulnerabilities notwithstanding).

app/models/package_manager/packagist.rb
```diff
@@ -66,31 +68,27 @@ def self.deprecation_info(name)
...
+
       {
         name: latest_version["name"],
         description: latest_version["description"],
-        homepage: latest_version["home_page"],
+        homepage: latest_version["homepage"],
         keywords_array: Array.wrap(latest_version["keywords"]),
         licenses: latest_version["license"].join(","),
-        repository_url: repo_fallback(raw_project["repository"], latest_version["home_page"]),
+        repository_url: repo_fallback(latest_version["source"].first&.fetch("url"), latest_version["homepage"]),
```
mikeyoung85, line 84: In the example file they have on their API docs site it looks like this will link to the `.git` url like `https://github.com/Seldaek/monolog.git`. Do you know if that URL gets normalized and removes the `.git` on the end?

tiegz: Ah, good find! It does not get normalized, so I'll strip that

tiegz: Woops, nevermind, the `repo_fallback` method actually does `URLParser.try_all`, which cleans any VCS urls.

PR #2952, opened 2022-06-16
Silence some errors from removed platform projects.

tiegz:
We still do a few things with projects where the platform has been removed (e.g. version notification mailers), but they're currently blowing up because `Project#platform_name` is busted. This fix returns the platform name directly if we know it has been removed.

Fixes https://app.bugsnag.com/tidelift/libraries-dot-io/errors/615b59f7cb008b00075ec41a


tiegz:
Actually it doesn't really make sense to keep sending the version notifications if the links in them don't work anymore. Will revisit this

PR #2955, opened 2022-06-20
Cache unreachable go hosts when looking up redirected go libraries.

tiegz:
The Go codepath for `Project.find_with_package_manager` does a lookup for uncommon Go hosts to get their fully qualified import names in case they're redirects (e.g. `go.uber.org/multierr` -> redirects to -> `github.com/uber-go/multierr`), via https://github.com/librariesio/libraries.io/pull/2386

Recently we've been seeing thousands of unreachable hosts via `Faraday::ConnectionFailed` errors per day, so this caches the host so we remember if it's unreachable. 

https://app.bugsnag.com/tidelift/libraries-dot-io/errors/62966d29b7863b000831d2b7

PS I wish I had a better fix to prevent doing these lookups for all Go package lookups, but I think the only answer is a legit `aliases` column, which is a bigger fix.


tiegz:
Noting that this is still a problem, and seems to be exacerabated by pen-testers throwing garbage Go package names at libraries.

mikeyoung85 (approved):

PR #2957, opened 2022-07-05
self.dependencies requires versions in mapped_project for some package managers

katzj:
We need versions in mapped_project for some package managers for
dependencies to work. Bring it back for those specific ones.

### DIFF
--- a/app/models/package_manager/packagist.rb
+++ b/app/models/package_manager/packagist.rb
@@ -83,6 +83,7 @@ module PackageManager
         keywords_array: Array.wrap(latest_version["keywords"]),
         licenses: latest_version["license"]&.join(","),
         repository_url: repo_fallback(latest_version["source"]&.fetch("url"), latest_version["homepage"]),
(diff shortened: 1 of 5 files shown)


app/models/package_manager/packagist.rb
```diff
@@ -83,6 +83,7 @@ def self.mapping(raw_project)
         keywords_array: Array.wrap(latest_version["keywords"]),
         licenses: latest_version["license"]&.join(","),
         repository_url: repo_fallback(latest_version["source"]&.fetch("url"), latest_version["homepage"]),
+        versions: raw_project["versions"],
```
tiegz, line 86: since I was in here recently, I think you'd have to do this to get the versions now bc `raw_project` will just be a list of sorted versions:

```suggestion
        versions: raw_project.pluck("version"),
```

(example payload [here](﻿﻿﻿https://repo.packagist.org/p2/tiat/router.json))

tiegz (changes requested):

tiegz (approved):

PR #2961, opened 2022-07-12
Adds and implements Project#repository_url_set_by_admin

tiegz:
This adds `Project#repository_url_set_by_admin?` to denote that the repository_url was explicitly set by an admin, and to prevent it from being overwritten by PackageManagerDownloadWorker. 

It's similar to the `Project#license_set_by_admin?` that was added in https://github.com/librariesio/libraries.io/pull/2448.

(the one thing I'm not doing is adding this to the API, bc I'm not sure if we need it?)


tiegz:
I think this might be the wrong approach, so I'm closing and opening a different fix.

PR #2991, opened 2022-09-08
Perf: project maintained scope

travatomic:
The Project#maintained scope is a bit of a hot path, and is not using indexed queries. These queries are often 3+ seconds. This change will use an index only scan for many of the queries, bringing them down to under 1 second.

* Change the `Project#maintained` scope to use inclusive matching rather than exclusive matching
* Add an index to support the Project#maintained scope. Unfortunately since this includes querying on NULL, this index is a bit tricky. It would be better if Project#status disallowed null and used a proper value for the initial state, but that would cause other things to break. I'm going for the lowest impact performance change here.


djpowers (approved):
Seems reasonable to me 👍

tiegz (approved):

PR #2997, opened 2022-09-26
Change VersionUpdateFailure behavior to be less exceptional.

tiegz:
Currently, when we get a newly published version and sync a `PackageManagerDownloadWorker` job, we'll raise a `VersionUpdateFailure` (and ignore that error in bugsnag) if we can't find the version. This is most likely because the package repo has a stale cache, so we let the job try again and it is usually a long enough delay for the new version to exist on the repo.

There are some problems with this:
* since we know this is because of repo staleness, this doesn't really count as "exceptional" enough to raise an error
* this leads us to thousands of regular jobs in the "retries" queue normally, which conceals real errors.
* we're also logging the backtraces and passing around the backtraces for these thousands of errors, which is a lot of strings.
* regarding ☝🏻 , I've noticed the sidekiq workers that OOM have a ton of these errors in their logs, so they might be relieved to stop handling this error all the time.

I propose this strategy instead:
* if `PackageManagerDownloadWorker` can't find the version, *requeue* the job in 2 seconds, so that it goes into the "scheduled" queue 
* if, after 30 attempts (at least 60 seconds later), we still can't fetch the version, assume something else is wrong and raise VersionUpdateFailure
* allow the VersionUpdateFailure to surface in bugsnag again so we can be aware of things like repo outages, etc.

### DIFF
--- a/app/workers/package_manager_download_worker.rb
+++ b/app/workers/package_manager_download_worker.rb
@@ -67,10 +67,10 @@ class PackageManagerDownloadWorker
     project = platform.update(name, sync_version: sync_version)
 
     # Raise/log if version was requested but not found
-    if version.present? && true # !Version.exists?(project: project, number: version)
+    if version.present? && !Version.exists?(project: project, number: version)
       Rails.logger.info("[Version Update Failure] platform=#{key} name=#{name} version=#{version}")
       if requeue_count < MAX_ATTEMPTS_TO_UPDATE_FRESH_VERSION_DATA
-        PackageManagerDownloadWorker.perform_in(2.seconds, platform_name, name, version, source, requeue_count + 1)
+        PackageManagerDownloadWorker.perform_in(5.seconds, platform_name, name, version, source, requeue_count + 1)
       else
         raise VersionUpdateFailure
       end
--- a/app/workers/package_manager_download_worker.rb
+++ b/app/workers/package_manager_download_worker.rb
@@ -1,10 +1,16 @@
 # frozen_string_literal: true
 
 class PackageManagerDownloadWorker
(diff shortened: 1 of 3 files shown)


app/workers/package_manager_download_worker.rb
```diff
@@ -61,9 +67,13 @@ def perform(platform_name, name, version = nil, source = "unknown")
     project = platform.update(name, sync_version: sync_version)
 
     # Raise/log if version was requested but not found
-    if version.present? && !Version.exists?(project: project, number: version)
+    if version.present? && true # !Version.exists?(project: project, number: version)
       Rails.logger.info("[Version Update Failure] platform=#{key} name=#{name} version=#{version}")
-      raise VersionUpdateFailure
+      if requeue_count < MAX_ATTEMPTS_TO_UPDATE_FRESH_VERSION_DATA
+        PackageManagerDownloadWorker.perform_in(2.seconds, platform_name, name, version, source, requeue_count + 1)
```
katzj, line 73: I might push this out to be more like 5 seconds but I like the approach!

katzj (approved):

app/workers/package_manager_download_worker.rb
```diff
@@ -61,9 +67,13 @@ def perform(platform_name, name, version = nil, source = "unknown")
     project = platform.update(name, sync_version: sync_version)
 
     # Raise/log if version was requested but not found
-    if version.present? && !Version.exists?(project: project, number: version)
+    if version.present? && true # !Version.exists?(project: project, number: version)
```
mikeyoung85, line 70: Do we need to add this conditional back in? With it commented out it will re queue every time right?

PR #2999, opened 2022-09-28
Decrease VersionUpdateFailure attempts to 15 and skip them for go.

tiegz:
Sometimes go modules can't be found on pkg.go.dev, e.g. https://pkg.go.dev/github.com/oraichain/wasmd . It could be because they're forks or some other reason, but the rules around pkg.go.dev indexing are elusive and changing. 

If we come across a go version update for this case, silently ignore them. If someone notices a missing package, they can manually click "Request package" on pkg.go.dev and it'll possibly fetch the package, and then we can index it on Libraries too.


app/workers/package_manager_download_worker.rb
```diff
@@ -10,7 +10,7 @@ class PackageManagerDownloadWorker
   # We were unable to fetch version, even after waiting for repo caches to refresh.
   class VersionUpdateFailure < StandardError; end
 
-  MAX_ATTEMPTS_TO_UPDATE_FRESH_VERSION_DATA = 30
```
tiegz, line 13: 30 seemed like overkill, so I'm decreasing to 15 attempts (~1.25 minutes)

jhan217 (approved):

PR #3006, opened 2022-10-03
Replace stackdriver tracing with datadog tracing

tiegz:
Now that datadog is setup, we can swap the tracing from Google::Cloud -> Datadog.


pranav:
@katzj knows better than me if we're using stackdriver for anything or if its got anything more than what dd is doing for us

katzj (approved):

Issue #3008, opened 2022-10-03
Pypi API change causing missing depndencies


DanBeard:
Thanks for taking the time to raise an issue. This template should guide you through the process of submitting a bug, enhancement or feature request. Please erase any part of this template that is not relevant to your issue. 

## Bugs
Before submitting a bug report:

- [x] Double-check that the bug is persistent,
- [x] Double-check the bug hasn't already been reported [on our issue tracker](https://github.com/issues?utf8=%E2%9C%93&q=is%3Aopen+is%3Aissue+org%3Alibrariesio), they *should* be labelled `bug` or `bugsnag`.

Description:

Pypi recently removed "releases" from their JSON API. The libraries.io pypi repository model is trying to log this value and throwing an exception (that gets eaten without logging in it's parent class base.rb). This causes all new projects to be shown with no dependencies. 

Steps to reproduce:

1. View a new pypi project release on libraries.io, via browser or JSON API
2. Observe there are no dependencies listed. Compare response from Pypi JSON api.

Random Example chosen for no particular reason besides it's a new Pypi project with dependencies:
1. View https://libraries.io/pypi/arduino-exporter/0.2.0
2. Compare to https://pypi.org/pypi/arduino-exporter/0.2.0/json  ( dependencies are listed in $["info"]["requires_dist"] )

PR #3009, opened 2022-10-03
Removed logging check that was causing dependency fetching to fail

DanBeard:

- [x] Is there a corresponding ticket for your pull request?
- [x] Have you successfully run the project with your changes locally?

Quick fix for #3008 . source_info is nil since Pypi removed `$["releases"]` from the JSON API response. Commenting out this line stops the error from being thrown and allows dependencies to be grabbed for projects in my local env. 

Unsure when the Pypi API was actually changed. Recent Pypi packages may need to be re-evaluated to ensure dependency information in libraries,io is correct.

### DIFF
--- a/app/models/package_manager/pypi.rb
+++ b/app/models/package_manager/pypi.rb
@@ -127,7 +127,7 @@ module PackageManager
       api_response = get("https://pypi.org/pypi/#{name}/#{version}/json")
       deps = api_response.dig("info", "requires_dist")
       source_info = api_response.dig("releases", version)
-      Rails.logger.warn("Pypi sdist (no deps): #{name}") unless source_info.any? { |rel| rel["packagetype"] == "bdist_wheel" }
+      #Rails.logger.warn("Pypi sdist (no deps): #{name}") unless source_info.any? { |rel| rel["packagetype"] == "bdist_wheel" }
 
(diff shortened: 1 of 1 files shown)


app/models/package_manager/pypi.rb
```diff
@@ -127,7 +127,7 @@ def self.dependencies(name, version, _mapped_project)
       api_response = get("https://pypi.org/pypi/#{name}/#{version}/json")
       deps = api_response.dig("info", "requires_dist")
       source_info = api_response.dig("releases", version)
-      Rails.logger.warn("Pypi sdist (no deps): #{name}") unless source_info.any? { |rel| rel["packagetype"] == "bdist_wheel" }
+      #Rails.logger.warn("Pypi sdist (no deps): #{name}") unless source_info.any? { |rel| rel["packagetype"] == "bdist_wheel" }
```
tiegz, line 130: This might be nicer, to maintain the warnings:

```suggestion
      Rails.logger.warn("Pypi sdist (no deps): #{name}") unless source_info&.any? { |rel| rel["packagetype"] == "bdist_wheel" }
```

DanBeard: Confirmed locally that this behaves the same as commenting it out.

PR #3010, opened 2022-10-03
sidekiq minor version update: 5.2.10

tiegz:
The Sidekiq 6.0 [upgrade guide](https://github.com/mperham/sidekiq/blob/main/6.0-Upgrade.md) recommends going to the the highest 5.* version and testing it before upgrading to 6.*.


app/controllers/application_controller.rb
```diff
@@ -99,7 +99,7 @@ def find_project
     if @project.name != params[:name]
       redirect_to(
         # Unescape since url_for automatically escapes our already-escaped project name
-        URI.unescape(
+        Addressable::URI.unescape(
```
tiegz, line 102: Unrelated, but I missed this in my recent URI.escape PR

travatomic (approved):

PR #3013, opened 2022-10-05
Remove spring and listen gems

tiegz:
When I profile RepositoryDownloadWorker, I see 30% time taken up by the `listen` gem, which was added [almost 6 years ago](https://github.com/librariesio/libraries.io/commit/6501db1ea0ed3a59c81719c6f9eb5fb7129be455) and which I don't think we need.


pranav (approved):
oh yeah if we dont do any hotreloading of code during local dev which i could see this gem being used for, then probably dont neeed it?

PR #3020, opened 2022-11-17
Add rake task to remove repository_maintenance_stats duplicates to prepare for unique index creation

papiro:
There are currently 1,301,892 repository_maintenace_stats so I think it should be okay to run this query on that many records. Both queries were adapted from the queries [at this stackoverflow post](https://stackoverflow.com/questions/53722174/most-efficient-way-to-remove-duplicates-postgres).

The first query is from the OP and the second query is from the accepted answer. I like the first query better since it is really succinct and easy to understand and according to the accepted answer is faster, but the author of the question left a comment that said he didn't know why it was sometimes leaving duplicates 🤷 . I don't see anything wrong with the query and if anyone thinks our mileage might be different then I can remove sql2 and use the first sql. Otherwise I'll remove the first sql and just use sql2.

After this is run, then a unique index will be added to ensure these duplicates don't reappear.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -109,7 +109,7 @@ end
 
 group :test do
   gem "database_cleaner"
-  gem "factory_bot_rails"
+  gem "factory_bot_rails", "~>5.2.0"
   gem "faker"
   gem "json_spec"
   gem "poltergeist", "1.17.0"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -185,10 +185,10 @@ GEM
       ffi (>= 1.15.0)
     excon (0.72.0)
     execjs (2.7.0)
-    factory_bot (5.0.2)
+    factory_bot (5.2.0)
       activesupport (>= 4.2.0)
-    factory_bot_rails (5.0.2)
-      factory_bot (~> 5.0.2)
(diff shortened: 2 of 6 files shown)


mpace965:
I don't have a strong preference on the query, but leaning towards the window query version because we've used it before.

What is the execution plan for adding this index? Is it reasonable to pause writes to RepositoryMaintenanceStats right before deduping and resume after the index is added?

papiro:
> What is the execution plan for adding this index? Is it reasonable to pause writes to RepositoryMaintenanceStats right before deduping and resume after the index is added?

oh yeah that's a good point. how do we pause writes? The write is coming from `RepositoryMaintenanceStatWorker` if that helps.

papiro:
> Is that the only place we write to the table? Could just turn that worker off for a while. Could also change the worker to read before writing, and update the most recent RepositoryMaintenanceStat for the given repo + category (if one exists and create otherwise).

Ahh yes, I looked into that actually. It already is doing a `find_or_create_by` but I read that if 2 workers run that at the same time, 2 duplicate records can be added, unless there is a unique constraint.

mpace965 (approved):

PR #3035, opened 2023-01-10
Reduce Version Syncing When Updating a Go Base Project

mikeyoung85:
* Only run a sync for base project if it doesn't exist
* If we do run a sync target only a single version
* Updated specs
* Some lint fixes


app/models/package_manager/go.rb
```diff
@@ -79,22 +79,29 @@ def self.update(name, sync_version: :all)
       # call update on base module name if the name is appended with major version
       # example: github.com/myexample/modulename/v2
       # use the returned project name in case it finds a Project via repository_url
-      update_base_module(project.name) if project.present? && project.name.match?(VERSION_MODULE_REGEX)
+
+      # pick one version to use for the base module update so that we don't sync every version
+      # when reusing the PackageManagerDownloadWorker to initialize the base name Project
+      base_sync_version = sync_version unless sync_version == :all
+      base_sync_version = base_sync_version.presence || project.versions.limit(1).first.number.presence
```
tiegz, line 86: looks like you could probably get rid of the `limit(1)`

tiegz (approved):

PR #3039, opened 2023-01-18
Maven: set "scm.url" in parent props manually, in case they're outside parent properties

tiegz:
Fixes `maven/com.indeed:proctor-tomcat-deps` on Libraries, which currently has a repository_url of `${project.parent.scm.url}`

If a parent pom has a `<scm><url /></scm>` value outside of the `<properties>`, this will add "scm.url" to the parent properties so we pass it on to Bibliothecary for interpolatuion.

(background: pom parent interpolation was added in https://github.com/librariesio/libraries.io/pull/2380)

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -96,11 +96,12 @@ GEM
     ast (2.4.2)
     autoprefixer-rails (9.5.1.1)
       execjs
-    bibliothecary (7.3.4)
+    bibliothecary (8.5.1)
       commander
       deb_control
       librariesio-gem-parser
       ox (>= 2.8.1)
+      packageurl-ruby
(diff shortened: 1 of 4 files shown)


Gemfile.lock
```diff
@@ -96,11 +96,12 @@ GEM
     ast (2.4.2)
     autoprefixer-rails (9.5.1.1)
       execjs
-    bibliothecary (7.3.4)
+    bibliothecary (8.5.1)
```
tiegz, line 99: Pulling in this change: https://github.com/librariesio/bibliothecary/pull/567

mikeyoung85 (approved):

PR #3040, opened 2023-01-24
Update Pypi.dependencies after update to package manager api

papiro:
This is fixing Pypi dependencies after this update: https://github.com/pypi/warehouse/pull/11775 which broke the existing implementation.

### DIFF
--- a/app/models/package_manager/pypi.rb
+++ b/app/models/package_manager/pypi.rb
@@ -11,6 +11,8 @@ module PackageManager
     ENTIRE_PACKAGE_CAN_BE_DEPRECATED = true
     SUPPORTS_SINGLE_VERSION_UPDATE = true
     PYPI_PRERELEASE = /(a|b|rc|dev)[0-9]+$/.freeze
+    # Adapted from https://peps.python.org/pep-0508/#names
+    PYPI_508_NAME_REGEX = /^([A-Z0-9][A-Z0-9._-]*[A-Z0-9])/i
 
     def self.package_link(db_project, version = nil)
       # NB PEP 503: "All URLs which respond with an HTML5 page MUST end with a / and the repository SHOULD redirect the URLs without a / to add a / to the end."
@@ -123,6 +125,14 @@ module PackageManager
         &.index_by { |v| v[:number] } || {}
     end
 
+    # Simply parses out the name of a PEP 508 Dependency specification: https://peps.python.org/pep-0508/
+    # Leaves the rest as-is with any leading semicolons or spaces stripped
+    def self.parse_pep_508_dep_spec(dep)
+      name, requirement = dep.split(PYPI_508_NAME_REGEX, 2).last(2).map(&:strip)
+      requirement = requirement&.sub(/^[\s;]*/, "") || ""
+      return name, requirement
+    end
+
     def self.dependencies(name, version, _mapped_project = nil)
       api_response = get("https://pypi.org/pypi/#{name}/#{version}/json")
       deps = api_response.dig("info", "requires_dist")
@@ -130,7 +140,7 @@ module PackageManager
       Rails.logger.warn("Pypi sdist (no deps): #{name}") unless source_info.any? { |rel| rel["packagetype"] == "bdist_wheel" }
 
       deps.map do |dep|
-        name, version = dep.split(/ /, 2)
+        name, version = parse_pep_508_dep_spec(dep)
         {
           project_name: name,
           requirements: version.nil? || version == ";" ? "*" : version.gsub(/\(|\)/, ""),
--- a/app/models/package_manager/pypi.rb
+++ b/app/models/package_manager/pypi.rb
@@ -12,7 +12,7 @@ module PackageManager
     SUPPORTS_SINGLE_VERSION_UPDATE = true
     PYPI_PRERELEASE = /(a|b|rc|dev)[0-9]+$/.freeze
     # Adapted from https://peps.python.org/pep-0508/#names
-    PYPI_508_NAME_REGEX = /^([A-Z0-9][A-Z0-9._-]*[A-Z0-9])/i
+    PEP_508_NAME_REGEX = /^([A-Z0-9][A-Z0-9._-]*[A-Z0-9])/i
 
     def self.package_link(db_project, version = nil)
       # NB PEP 503: "All URLs which respond with an HTML5 page MUST end with a / and the repository SHOULD redirect the URLs without a / to add a / to the end."
@@ -128,7 +128,7 @@ module PackageManager
     # Simply parses out the name of a PEP 508 Dependency specification: https://peps.python.org/pep-0508/
     # Leaves the rest as-is with any leading semicolons or spaces stripped
     def self.parse_pep_508_dep_spec(dep)
-      name, requirement = dep.split(PYPI_508_NAME_REGEX, 2).last(2).map(&:strip)
+      name, requirement = dep.split(PEP_508_NAME_REGEX, 2).last(2).map(&:strip)
       requirement = requirement&.sub(/^[\s;]*/, "") || ""
       return name, requirement
     end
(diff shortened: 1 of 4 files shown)


app/models/package_manager/pypi.rb
```diff
@@ -123,14 +123,14 @@ def self.known_versions(name)
         &.index_by { |v| v[:number] } || {}
     end
 
-    def self.dependencies(name, version, _mapped_project)
+    def self.dependencies(name, version, _mapped_project = nil)
```
mikeyoung85, line 126: Not sure we need this change. Is something calling this method with passing in the third argument?

app/models/package_manager/pypi.rb
```diff
@@ -123,14 +123,14 @@ def self.known_versions(name)
...
       api_response = get("https://pypi.org/pypi/#{name}/#{version}/json")
       deps = api_response.dig("info", "requires_dist")
-      source_info = api_response.dig("releases", version)
+      source_info = api_response.fetch("urls", [])
       Rails.logger.warn("Pypi sdist (no deps): #{name}") unless source_info.any? { |rel| rel["packagetype"] == "bdist_wheel" }
 
       deps.map do |dep|
-        name, version = dep.split
+        name, version = dep.split(/ /, 2)
```
mikeyoung85, line 133: What was the reason for this change?

app/models/package_manager/pypi.rb
papiro: Just the test.

app/models/package_manager/pypi.rb
papiro: ohhh yes my bad - I meant to link the slack discussion on this: https://tidelift.slack.com/archives/C04GTTC0UJ0/p1674508566449779?thread_ts=1674504157.951459&cid=C04GTTC0UJ0

papiro: TLDR: before we were saying definitively that a package had a certain number of dependencies, even if some of those dependencies only came with variants of the package. Now we include the information specifying whether they are only included with a variant.

mpace965: Allegedly (according to [this StackOverflow post](https://stackoverflow.com/questions/71165471/what-is-a-format-of-pypis-requires-dist-in-json-api)), these dependencies are in the [PEP 508](https://peps.python.org/pep-0508) format. I can't find an official PyPI source stating this though.

Looking at the tests in the [grammar section](https://peps.python.org/pep-0508/#complete-grammar), there are a couple of instances of valid PEP 508 dependencies where the part after a space would not specify a version, e.g.

```
name[quux, strange];python_version<'2.7' and platform_version=='2'
=> ["name[quux,", "strange];python_version<'2.7'"and platform_version=='2'"]
```

That's not the point of this change, but it's possible that the current (and this) implementation would create bad data

mpace965: That being said we should probably file something else to address it

mpace965 (approved):

app/models/package_manager/pypi.rb
papiro: oof... 

According to [this](https://peps.python.org/pep-0508/#names), we should be able to use this regex to parse the name: `^([A-Z0-9]|[A-Z0-9][A-Z0-9._-]*[A-Z0-9])`. I suppose the rest of it we could provide in `requirements`?

app/models/package_manager/pypi.rb
```diff
@@ -11,6 +11,8 @@ class Pypi < Base
     ENTIRE_PACKAGE_CAN_BE_DEPRECATED = true
     SUPPORTS_SINGLE_VERSION_UPDATE = true
     PYPI_PRERELEASE = /(a|b|rc|dev)[0-9]+$/.freeze
+    # Adapted from https://peps.python.org/pep-0508/#names
+    PYPI_508_NAME_REGEX = /^([A-Z0-9][A-Z0-9._-]*[A-Z0-9])/i
```
mpace965: I think this needs the first part of the regex too, or else we won't be able to parse single character packages like [a](https://pypi.org/project/a/)

mpace965: I think it would fail on something like `a>=3`, for example

papiro: oooh you're right!

app/models/package_manager/pypi.rb
```diff
@@ -123,14 +125,22 @@ def self.known_versions(name)
         &.index_by { |v| v[:number] } || {}
     end
 
+    # Simply parses out the name of a PEP 508 Dependency specification: https://peps.python.org/pep-0508/
+    # Leaves the rest as-is with any leading semicolons or spaces stripped
+    def self.parse_pep_508_dep_spec(dep)
+      name, requirement = dep.split(PYPI_508_NAME_REGEX, 2).last(2).map(&:strip)
```
mpace965: is `last(2)` needed here? and curious if it is, wouldn't we want the `first(2)`?

app/models/package_manager/pypi.rb
mpace965: good ol' a

app/models/package_manager/pypi.rb
```diff
@@ -123,14 +125,22 @@ def self.known_versions(name)
         &.index_by { |v| v[:number] } || {}
     end
 
-    def self.dependencies(name, version, _mapped_project)
+    # Simply parses out the name of a PEP 508 Dependency specification: https://peps.python.org/pep-0508/
+    # Leaves the rest as-is with any leading semicolons or spaces stripped
+    def self.parse_pep_508_dep_spec(dep)
+      name, requirement = dep.split(PEP_508_NAME_REGEX, 2).last(2).map(&:strip)
```
mpace965, line 131: worth noting that `requirement` is not necessarily a version still. I think it's still an improvement over what was here before though. To really be certain, I think we'll have to implement that parser and use it to see if we can pull a version out. I think that's beyond the scope of this ticket, but worth filing a separate story for.

papiro: but `requirements` is what we need: https://github.com/librariesio/libraries.io/blob/main/app/models/dependency.rb#L12

It wouldn't be possible to pull a version out of there i don't think.

mpace965: got it, that makes sense. I think what confused me is it being named `version` in `self.dependencies`. Worth renaming that to `requirement`/`requirements`

mpace965 (approved):
Not sure if the paren stripping was important, but otherwise LGTM

app/models/package_manager/pypi.rb
papiro: when we split on the name, what we get is a blank string (what's to the left of the name) and the requirements (to the right). Because we have a capture group on the name itself, apparently the `split` method doesn't count the captured matches towards the _limit_ parameter. https://apidock.com/ruby/String/split
> If limit is a positive number, at most that number of [split](https://apidock.com/ruby/String/split) substrings will be returned (captured groups will be returned as well, but are not counted towards the limit)

Without the capture group and with a limit of 2 we get `[blank string, requirements]`, but with the capture group, it seems to stick it in the middle so `[blank string, name, requirements]`. 

In stepping through and testing this again I did discover we don't need the part on the line after which defaults requirement to `""` if it's `nil`. After fixing the regex to cover single character packages it now has `[blank string, name, blank string]` for those packages.

papiro:
> Not sure if the paren stripping was important, but otherwise LGTM

I noticed it would mess with dependency specifications like this: `"os_name=='a' and (os_name=='b' or os_name=='c')"`. I looked through the git history to see why we added that and there wasn't a specific reason. I think it was just cosmetic because most dep spec requirements are simple and are yet surrounded by the parens.

app/models/package_manager/pypi.rb
mpace965: woah, that's a little odd but makes sense I guess.

PR #3041, opened 2023-01-27
Rake task to backfill missing pypi version dependencies

papiro:
This rake task takes a `limit` argument and will only query for broken versions which haven't already been fixed.

Open to suggestions on when/how to run this. The purpose of taking the `limit` argument is to avoid DOSing Pypi, but we don't know if/at what limit that would happen and don't want to find out.

The original idea was to run the `PackageManagerDownloadWorker` on each corrupted version and stagger the queuing of the runs by 1 second. I'd like to apply the same idea to the running of this rake task, but this task no longer runs `PackageManagerDownloadWorker` and takes a much more lightweight approach, basically saving 1 pypi api call per corrupt version (~600,000) and not modifying `last_synced_at` in the process.


papiro:
Need to rethink this

PR #3045, opened 2023-02-02
Backfill missing Pypi dependencies

papiro:
Rewrote the rake task for the backfill based on learnings from the past couple days.

This query doesn't find broken projects. It finds projects which are potentially broken and effectively short-lists them to be resynced (by `PackageManagerDownloadWorker`).

It will find affected projects in batches of 120, group those into 2, and then run 2 every second for 1 minute, and then repeat for the next minute, and so on. It will fix 2 projects per second until all projects are fixed.

There are currently `103999` affected projects, so it is expected to take ~14.5 hours.

The reason we are _not_ running a script which calls `Pypi#save_dependencies` (like in a [previous commit](https://github.com/librariesio/libraries.io/pull/3045/commits/9cd7633c11ce0ab284e37b55766cb49770ae41a5) is because it doesn't set an indicator which allows it to be filtered out in subsequent queries in case the backfill unexpectedly stops or needs to be stopped and restarted for any reason.

The reason we are _not_ querying for versions instead of projects and running `PackageManagerDownloadWorker` with the `version` arg is to avoid a situation where a project has multiple affected versions and one gets completed by the worker and the other fails or the task needs to be stopped and restarted. In that case the query will not pick the project back up since its `last_resync_at` timestamp got updated and it wouldn't be possible to query for such an anomaly.

7/6/2022 is the date the Pypi api changed
2/1/2022 is the day after the fix to Libraries (#3040) went live


app/models/package_manager/pypi.rb
```diff
@@ -135,7 +135,7 @@ def self.parse_pep_508_dep_spec(dep)
 
     def self.dependencies(name, version, _mapped_project = nil)
       api_response = get("https://pypi.org/pypi/#{name}/#{version}/json")
-      deps = api_response.dig("info", "requires_dist")
+      deps = api_response.dig("info", "requires_dist") || []
```
papiro, line 138: This just prevents us from logging if `requires_dist` is `null`. It was raising on the `deps.map` below, but it seems like it's Pypi's way (or _a way_) of indicating that there are no dependencies.

tiegz (approved):
nice! Getting projects instead of versions seems like a good strategy

PR #3047, opened 2023-02-06
Update GH metrics GraphQL query & VCR cassettes

johnbintz-tidelift:
vim also reformatted some of the code for me.


app/models/maintenance_stats/queries/github/issues_query.rb
```diff
@@ -3,7 +3,7 @@ module Queries
     module Github
       class IssuesQuery < BaseQuery
         ISSUES_QUERY = Rails.application.config.graphql.client.parse <<-GRAPHQL
-          query($owner: String!, $repo_name: String!, $open_pr_query: String!, $closed_pr_query: String!, $one_year: GitTimestamp!){
+          query($owner: String!, $repo_name: String!, $open_pr_query: String!, $closed_pr_query: String!, $one_year: DateTime!){
```
johnbintz-tidelift, line 6: This is the important change. The `since` parameter in the filter needed a `DateTime` object, so it was failing when receiving a `GitTimestamp`: https://docs.github.com/en/graphql/reference/input-objects#issuefilters

tiegz (approved):

spec/models/maintenance_stats/github/contributor_count_stats_spec.rb
```diff
@@ -1,49 +1,50 @@
...
   let(:client) { AuthToken.client }
   let(:query_klass) { MaintenanceStats::Queries::Github::RepositoryContributorStatsQuery.new(client) }
-  let(:query_params) { {full_name: "vuejs/vue"} }
+  let(:query_params) { { full_name: "vuejs/vue" } }
 
   let(:stat) { described_class.new(query_results) }
 
   context "with a valid repository" do
-    let(:repository) { create(:repository, full_name: 'vuejs/vue') }
+    let(:repository) { create(:repository, full_name: "vuejs/core") }
```
jhan217, line 14: 🎱 Was this change intentional?

app/models/maintenance_stats/queries/github/issues_query.rb
jhan217: 🎱 Might be worth a code comment if it's not expected/obvious.

PR #3050, opened 2023-02-28
Stop syncing of a number of maven repositories

johnbintz-tidelift:


jhan217 (approved):

app/workers/package_manager_download_worker.rb
```diff
@@ -39,12 +39,7 @@ def initialize(platform, name, version)
     inqlude: PackageManager::Inqlude,
     julia: PackageManager::Julia,
     maven: PackageManager::Maven,
-    maven_atlassian: PackageManager::Maven::Atlassian,
```
mikeyoung85, line 42: I _think_ this is only used when the worker is called with something like `maven_springlibs` and we aren't storing that value anywhere in the database. Does that sound right? I'm just trying to double check that removing these isn't going to break Project lookups for Maven projects that aren't in Central.

app/models/package_manager/maven/jboss_ea.rb
```diff
@@ -1,18 +0,0 @@
-# frozen_string_literal: true
-
-class PackageManager::Maven::JbossEa < PackageManager::Maven::Common
```
mikeyoung85, line 3: I would probably hold off on removing these files until we can take a look through the data and verify we don't have any `Versions` pointing to these repository sources

app/workers/package_manager_download_worker.rb
```diff
@@ -39,12 +39,7 @@ def initialize(platform, name, version)
...
-    maven_hortonworks: PackageManager::Maven::Hortonworks,
     maven_mavencentral: PackageManager::Maven::MavenCentral,
-    maven_springlibs: PackageManager::Maven::SpringLibs,
-    maven_jboss: PackageManager::Maven::Jboss,
```
tiegz, line 46: was there a reason for stopping JBoss? I still see about 3k updates from depper in the past month for `maven_jboss`.

docs/update-maven-repository-sources.md
```diff
@@ -0,0 +1,17 @@
...
+    * app/workers/package_manager_workload_worker.rb, `PLATFORMS`, add above class.
+* depper
+    * ingestors/maven.go
+      * add constant next to `MavenCentral`
+      * add to `GetParser()`
+    * main.go, add constant to `registerIngestors`
+* maven-updater
+  * src/m/j/i/l/m/services/, duplicate `MavenCentralUpdater.javaa and rework.
+    * The code uses reflection to find the new updater so you don't need to
+      change anything else.
```
tiegz, line 17: :thumbsup:  just FWIW, the process is similar for Conda and Packagist, IIRC

app/workers/package_manager_download_worker.rb
tiegz: AFAIK it's only a mapping from the depper-provided platforms to these subclasses in Libraries

johnbintz-tidelift:
I've talked this over with @mikeyoung85 and we're going to hold off on removing classes here until the syncing of the platforms has stopped on depper and the package download worker has stopped processing the removed sources.

tiegz (approved):

PR #3053, opened 2023-03-03
Mark NPM packages as deprecated only when all releases are marked deprecated

macowie:
NPM handles package deprecation by marking every release as deprecated. This should update libraries to match


kjmahalingam (approved):

tiegz (approved):
one caveat here is that you can deprecate a specific version of a package, and if that's the latest version then the NPM package page will list it as "Deprecated", so it looks like the whole package is deprecated but it's really not (I'm trying to repro this right now to demonstrate)

PR #3061, opened 2023-03-09
Fix error handling while pulling github contributor stats

macowie:
This method had a broken adhoc retry mechanism for a Github REST API call. Octokit can already perform retries on this call if passed a timeout value. With the default 0.5 second wait, 10 seconds allowed enough attempts to pass with the existing VCR cassettes.

### DIFF
--- a/app/models/maintenance_stats/queries/github/repository_contributor_stats_query.rb
+++ b/app/models/maintenance_stats/queries/github/repository_contributor_stats_query.rb
@@ -13,7 +13,7 @@ module MaintenanceStats
         def query(params: {})
           validate_params(params)
 
-          @client.contributor_stats(params[:full_name], retry_wait: 0.5, retry_timeout: 10)
+          @client.contributor_stats(params[:full_name], retry_timeout: 10)
         end
       end
     end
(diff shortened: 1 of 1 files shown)


app/models/maintenance_stats/queries/github/repository_contributor_stats_query.rb
```diff
@@ -5,22 +5,20 @@ module Github
...
+        def query(params: {})
           validate_params(params)
 
-          resp = @client.contributor_stats(params[:full_name])
-          if @client.last_response.status == 202
-            sleep (count + 1) * 0.5
-            return query(params: params, count: count+=1)
-          end
-          resp
+          result = @client.contributor_stats(params[:full_name], retry_timeout: TIMEOUT_SEC)
```
djpowers, line 17: Good find on the built-in parameter!

djpowers (approved):

app/models/maintenance_stats/queries/github/repository_contributor_stats_query.rb
macowie: If only I didn't spend a bunch of time fixing the existing check first 😆

PR #3063, opened 2023-03-10
Fixes for gathering github maintenance stats

macowie:
* Octokit::Error expects to be initialized with response hash

### DIFF
--- a/app/models/maintenance_stats/queries/github/repository_contributor_stats_query.rb
+++ b/app/models/maintenance_stats/queries/github/repository_contributor_stats_query.rb
@@ -13,8 +13,9 @@ module MaintenanceStats
 
         def query(params: {})
           validate_params(params)
+          full_name = params[:full_name]
 
-          result = @client.contributor_stats(params[:full_name], retry_timeout: TIMEOUT_SEC)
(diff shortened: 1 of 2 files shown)


djpowers (approved):
Change seems fine. Not sure if it's related (could be flaky test), but it looks like this spec is failing: https://github.com/librariesio/libraries.io/blob/15359040158179647291c783a0b380b678dda142/spec/models/package_manager/rubygems_spec.rb#L54

Should be able to switch to `match_array` if we aren't concerned with the ordering.

macowie:
> Change seems fine. Not sure if it's related (could be flaky test), but it looks like this spec is failing:
> 
> https://github.com/librariesio/libraries.io/blob/15359040158179647291c783a0b380b678dda142/spec/models/package_manager/rubygems_spec.rb#L54
> 
> Should be able to switch to `match_array` if we aren't concerned with the ordering.

 I'm just wondering why this is happening now. It passes consistently locally, and fails consistently in CI, and only on this PR.

It doesn't seem like version order is guaranteed without a scope, so match_array could be more appropriate anyway, I just want make sure

djpowers (approved):
Right, puzzling why it's cropping up now, and just on this branch. Looks like that appeases the CI gods though, and seems like a reasonable expectation change to make anyway.

PR #3068, opened 2023-03-16
Add project_urls.Code as a possible PyPI repository URL

johnbintz-tidelift:
## PyPI API response

```
$ curl https://pypi.org/pypi/CairoSVG/json 2>/dev/null | jq '.info.project_urls'
{
  "Code": "https://github.com/Kozea/CairoSVG/",
  "Documentation": "https://cairosvg.org/documentation/",
  "Donation": "https://opencollective.com/courtbouillon",
  "Homepage": "https://courtbouillon.org/cairosvg",
  "Issue tracker": "https://github.com/Kozea/CairoSVG/issues"
}
```

##  Live site

Screenshot_20230316_162100

## With this change

Screenshot_20230316_162125

### DIFF
--- a/spec/models/package_manager/pypi_spec.rb
+++ b/spec/models/package_manager/pypi_spec.rb
@@ -239,31 +239,44 @@ describe PackageManager::Pypi do
     end
   end
 
-  describe "#mapping" do
-    let(:result) { described_class.mapping(raw_project) }
+  describe ".select_repository_url" do
+    let(:result) { described_class.select_repository_url(raw_project) }
 
     let(:raw_project) do
       {
         "info" => {
-          "name" => "name",
-          "summary" => "summary",
-          "home_page" => "home_page",
-          "keywords" => "keywords",
-          "classifiers" => [],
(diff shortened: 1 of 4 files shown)


spec/models/package_manager/pypi_spec.rb
```diff
@@ -238,4 +238,33 @@
...
+          "keywords" => "keywords",
+          "classifiers" => [],
+          "project_urls" => project_urls,
+        },
+      }
+    end
+
+    context "respository url" do
+      context "project_urls.Code" do
+        let(:project_urls) do
```
jhan217, line 260: Worth adding a case for when there are no project_urls to make sure it doesn't error out? And/or a case above/below it in the ordering of the extracted data to make sure that the repository_url gets the correct one? e.g. if  project_urls contain Repository and Code, that the Repository value is used; if it only contains Code and Homepage, that Code is being used.

jhan217 (approved):
Thanks for the refactors + adding the test coverages!

PR #3073, opened 2023-03-28
Don't let a bad Project spoil its Repository's stats

macowie:
A Repository can be related to multiple Projects. Right now, if **any** those Projects have an erroneous or missing `repository_url`|`homepage` set, then the Project doesn't have a github|bitbucket_owner_name set, and we don't gather stats on that repository and trash existing ones. Given not the name or anything from Projects are actually needed or used when gathering repository stats, I've replaced that check. 

If there's a reason at the Project level we'd want this integrity check or to skip/reset gathering stats can find a better place to do it.


johnbintz-tidelift (approved):
This seems fine to me!

djpowers (approved):

PR #3077, opened 2023-04-05
Create one-off task to sync potentially missing NuGet packages

johnbintz-tidelift:
For an unknown period of time, NuGet packages were not being updated due to a deprecated endpoint: https://github.com/librariesio/libraries.io/commit/45781c2f17eb83db19e52e3068a4521ecbe950a4

This PR adds a Rake task that fills in potentially missing packages that had updates for the 6 months prior to, and one month after, the referenced commit. Only missing packages will be created.


lib/tasks/one_off.rake
```diff
@@ -84,6 +84,20 @@ namespace :one_off do
...
+  # this can be run repeatedly until each package has been checked
+  desc "Backfill NuGet packages that might have been missed while using a deprecated API"
+  task backfill_missing_nuget_packages_2023_03: :environment do
+    packages = File.readlines("lib/tasks/input/2023-03-nuget_packages_to_check.txt").map(&:strip)
 
+    packages.each do |package_name|
+      puts "Checking #{package_name}..."
+      next if Project.find_by(platform: "NuGet", name: package_name)
+
+      puts "#{package_name} not found, updating..."
```
jhan217, line 98: It looks like this would produce ~58k * 2 lines of logs, would it be worth consolidating that into a summary count, or log only the packages that got updated?

jhan217 (commented):
Is the plan to run this in a production console? It looks like the session it would need to be open for at least 16 hours (~58029 package names * 1 second delay = 16.11 hours). Would it make sense to enqueue a bunch of async Sidekiq workers that do say, 1000 packages at a time instead?

Also, I wasn't able to easily check the contents of the input file on Github -- do you know if there are any packages with the same name but different capitalizations/cases? IIRC there was an issue where we have duplicate Nuget packages in libraries, each with a different capitalization scheme, so it would be good if this backfill could avoid exacerbating that issue.

johnbintz-tidelift:
I clicked merge too fast. I'll be opening another PR to address your concerns @jhan217. 

* The majority of these packages should be in Libraries already. I do plan on running this on a production console. If it starts to take too long, we can think about moving to Sidekiq.
* Since I'm going to be babysitting this while it runs, I do not mind seeing some logging as it runs.
* I got all of these package names from NuGet itself. There are 140 duplicates if I do case-insensitive comparisons. I'll change the package lookup to use a case insensitive check to avoid making duplicates.

jhan217:
Note that the comparisons should only be done on the lookup, the package names that are going in/getting created needs to have the correct case as that's what will show up in the libraries UI.

PR #3079, opened 2023-04-06
Severly limit number of missing packages to check

johnbintz-tidelift:
I used the Libraries.io API locally to filter packages so that we don't have to process as many entries on the server itself. This should massively cut down on the time needed to perform this check.

* Add more verbose logging since we have fewer entries now


kjmahalingam:
Could you describe on what condition you filtered the entries locally, just for context? Otherwise makes sense and will approve!

johnbintz-tidelift:
I requested information on multiple packages at a time from Libraries. The endpoint I was using returns either package data if it exists, or `nil` if it does not. I tracked which packages had data returned for them and which did not, and put the ones that did not into the new, shorter list.

jhan217 (approved):

PR #3089, opened 2023-04-12
Ignore GitHub hooks where the "sender" field is nil

tiegz:
Occasionally we receive GH hook events like "push" where the "sender" field is missing, e.g. for PR merges it seems. The docs [say](https://docs.github.com/en/webhooks-and-events/webhooks/webhook-events-and-payloads#push) that "sender" is optional for these events, and we don't update the Repository unless we find the "sender"'s GH token anyway, so we can ignore these. 

Fixes https://app.bugsnag.com/tidelift/libraries-dot-io/errors/63e7fef4e6b45a000baf7d20

### DIFF
--- a/app/services/github_hook_handler.rb
+++ b/app/services/github_hook_handler.rb
@@ -14,7 +14,7 @@ class GithubHookHandler
     when "issue_comment", "issues", "pull_request"
       return nil # noop
     when "push"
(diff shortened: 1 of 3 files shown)


kjmahalingam (approved):
LGTM!

PR #3092, opened 2023-04-13
Add detection for when a NuGet package is unlisted and mark it deprecated

macowie:
https://app.shortcut.com/tidelift/story/30844/recognize-when-nuget-packages-are-deprecated

Rolling back the changes when unlisted, sounds like we want to handle those a different way, but this keeps the fleshed out tests and cassettes for coming back to this.


app/models/package_manager/nu_get.rb
```diff
@@ -29,14 +29,27 @@ def self.install_instructions(db_project, version = nil)
...
+
+      if is_deprecated
+        message = catalog_entry.dig("deprecation", "message") || "unlisted"
+        # alternate_package is not currently stored in DB but I wanted to record it because it seems useful
+        alternate_package = catalog_entry.dig("deprecation", "alternatePackage", "id")
+
+        {
+          is_deprecated: true,
+          message: message,
+          alternate_package: alternate_package,
```
johnbintz-tidelift, line 44: Do any of the packages in the tests below have an alternative package? If so, can you update the tests to show that this data gathering is working?

macowie: Not sure, but that would be good to add!

kjmahalingam (approved):

PR #3093, opened 2023-04-14
Add IP to trusted_proxies so we get the real IP in logs

tiegz:
* adds libraries' IP to the trusted proxy list, so that `remote_ip()` filters it out from`X-Forwarded-For`
* also log `ip()`, to make sure it's correct (since that's the same value that rack-attack is using)

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -282,6 +282,7 @@ class ApplicationController < ActionController::Base
     payload[:rescued_error] = @rescued_error if @rescued_error
     payload[:current_user] = @current_user.id if @current_user
     payload[:remote_ip] = request.remote_ip
+    payload[:ip] = request.ip # TODO: remove this if these are always the same
     if @current_api_key
       payload[:api_key] = {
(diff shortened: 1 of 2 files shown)


papiro (approved):

havocp (approved):
I don't really know the details here but as long as you follow up on any issues 👍

tiegz:
I should've added some more background here:

* `remote_ip()` is [Rails' algorithm for calculating the real client IP](https://github.com/rails/rails/blob/main/actionpack/lib/action_dispatch/middleware/remote_ip.rb#L103-L121)
* `ip()` is [rack's algorithm for calculating the real client IP](https://github.com/rack/rack/blob/77cf0579ddd53da6e1a449a26850464284a75aec/lib/rack/request.rb#L414)

Since load balancer setups are different, they do their best job but sometimes you have to just tell it your load balancer IP, otherwise they might return your load balancer IP.

Re: rack-attack, it only has the `ip()` method available, so I just wanted to make sure it's getting the right IP too. (this has been [historically challenging](https://github.com/rack/rack-attack/issues/145)).

PR #3105, opened 2023-04-26
Fix retrieval of project status

kjmahalingam:
[Story](https://app.shortcut.com/tidelift/story/32776/fix-libraries-io-retrieval-of-project-status)

[Related PR for infra](https://github.com/tidelift/infra/pull/385)

This PR fixes `Project#check_status` and the `projects:check_status` rake task (which is being used in a new cron job in the related infra PR). We already regularly check removed projects to see if they've been reintroduced [here](https://github.com/tidelift/infra/blob/main/libraries-kube/cronjobs/libraries-projects-check-removed-status.yaml). This change is to do the opposite and check whether projects that exist have been removed.

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -618,20 +618,16 @@ class Project < ApplicationRecord
       update_attribute(:status, "Removed")
     elsif platform.downcase != "packagist" && [400, 404, 410].include?(response.response_code)
       update_attribute(:status, "Removed")
-    elsif platform.downcase == "clojars" && response.response_code == 404
-      update_attribute(:status, "Removed")
-    elsif platform.downcase == "pypi" && response.response_code == 404
-      update_attribute(:status, "Removed")
     elsif can_have_entire_package_deprecated?
       result = platform_class.deprecation_info(self)
       if result[:is_deprecated]
         update_attribute(:status, "Deprecated")
         update_attribute(:deprecation_reason, result[:message])
       else # in case package was accidentally marked as deprecated (their logic or ours), mark it as not deprecated
-        update_attribute(:status, '')
+        update_attribute(:status, nil)
         update_attribute(:deprecation_reason, nil)
       end
-    elsif deprecated_or_removed
+    elsif deprecated_or_removed # if package is already deprecated or removed and isn't anymore, remove status
       update_attribute(:status, nil)
     end
   end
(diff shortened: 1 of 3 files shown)


app/models/project.rb
```diff
@@ -618,20 +618,16 @@ def check_status(deprecated_or_removed = false)
       update_attribute(:status, "Removed")
     elsif platform.downcase != "packagist" && [400, 404, 410].include?(response.response_code)
       update_attribute(:status, "Removed")
-    elsif platform.downcase == "clojars" && response.response_code == 404
```
kjmahalingam, line 621: These conditions were unreachable, because they would have already been triggered by this:
`elsif platform.downcase != "packagist" && [400, 404, 410].include?(response.response_code)`

app/models/project.rb
```diff
@@ -618,20 +618,16 @@ def check_status(deprecated_or_removed = false)
...
-    elsif platform.downcase == "pypi" && response.response_code == 404
-      update_attribute(:status, "Removed")
     elsif can_have_entire_package_deprecated?
       result = platform_class.deprecation_info(self)
       if result[:is_deprecated]
         update_attribute(:status, "Deprecated")
         update_attribute(:deprecation_reason, result[:message])
       else # in case package was accidentally marked as deprecated (their logic or ours), mark it as not deprecated
-        update_attribute(:status, '')
+        update_attribute(:status, nil)
```
kjmahalingam, line 627: It felt incorrect to me that we had two cases of unsetting the status, one where it was an empty string and the other where it was nil. I changed this case to nil as well, and updated the specs accordingly. If this is incorrect then let me know and I can revert this change.

lib/tasks/projects.rake
```diff
@@ -23,7 +23,7 @@ namespace :projects do
   desc 'Check status of projects'
   task check_status: :environment do
     exit if ENV['READ_ONLY'].present?
-    ['npm', 'rubygems', 'packagist', 'nuget', 'cpan', 'clojars', 'cocoapods',
+    ['pypi', 'npm', 'rubygems', 'packagist', 'cpan', 'clojars', 'cocoapods',
```
kjmahalingam, line 26: Changed the list of package managers to match those in `projects:check_removed_status`, which was updated recently.

macowie (commented):
This looks good to me, just might need to be wary about the status of null vs. empty string. The scopes/queries that look at check null but not blanks. Still in favor of the change, just looking out for consequences

app/models/project.rb
tiegz: looks like the "" statuses are a very small number, so we could probably update those to nil after deploy?

tiegz (approved):

djpowers (approved):

PR #3111, opened 2023-05-09
Chore/unpin rubocop run autofix

djpowers:
This includes a few changes to use the latest version of RuboCop, and applies those fixes:
- updates app to use latest version of RuboCop (this was only a minor version bump)
- changes the RuboCop target Ruby version to match that of the app
- disables one style rule that resulted in a syntax error when autofixing
- safe autocorrects remaining errors
- adds an [ignore revs file](https://git-scm.com/docs/git-blame#Documentation/git-blame.txt---ignore-revs-fileltfilegt) so blame commands can ignore this refactor if desired
- repeats the previous three steps with autocorrecting all remaining errors

This should bring more consistency with other repos, and also fixes a snag I ran into where the 1.0 version of RuboCop doesn't support the `-a` flag, so running from the command line or plugins that run the command on your behalf don't work.


macowie (commented):
On main:

- 642 files inspected, 4461 offenses detected, 4352 offenses auto-correctable

This branch:

- 642 files inspected, 81 offenses detected

Well that's a bit more manageable 😅 

Still skimming/spot checking the autofixes but looks ok so far!

PR #3115, opened 2023-05-17
Remove ignored deprecated_or_removed parameter from CheckStatusWorker.perform

kjmahalingam:
What I believe is the appropriate fix for the patch introduced [here](https://github.com/librariesio/libraries.io/commit/3593fd23d715dd80fe80be656be5d469fe3612c7).

~Removed ignored parameter from method signature~, and removes it from other calls of this method. I did not see any other calls that were being made incorrectly in searching through the code.

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -229,7 +229,7 @@ class Project < ApplicationRecord
     return unless platform_class_exists?
 
     sync_classes.each { |sync_class| PackageManagerDownloadWorker.perform_async(sync_class.name, name, nil, "project", 0, force_sync_dependencies) }
-    CheckStatusWorker.perform_async(id, status == "Removed" || status == "Deprecated")
+    CheckStatusWorker.perform_async(id)
   end
 
   def sync_classes
--- a/app/workers/check_status_worker.rb
+++ b/app/workers/check_status_worker.rb
@@ -4,7 +4,7 @@ class CheckStatusWorker
   include Sidekiq::Worker
   sidekiq_options queue: :status, unique: :until_executed
 
-  def perform(project_id, _ignored = nil)
+  def perform(project_id)
     Project.find_by_id(project_id).try(:check_status)
   end
 end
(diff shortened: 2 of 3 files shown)


lib/tasks/projects.rake
```diff
@@ -77,7 +77,7 @@ namespace :projects do
     # Check if removed/deprecated projects are still deprecated/removed
     PLATFORMS_FOR_STATUS_CHECKS.each do |platform|
       Project.platform(platform).removed_or_deprecated.select("id").find_each do |project|
-        CheckStatusWorker.perform_async(project.id, true)
+        CheckStatusWorker.perform_async(project.id)
```
kjmahalingam, line 80: Removing the flag here should be fine given that we're already filtered down to projects that are removed or deprecated.

app/models/project.rb
```diff
@@ -229,7 +229,7 @@ def async_sync(force_sync_dependencies: false)
     return unless platform_class_exists?
 
     sync_classes.each { |sync_class| PackageManagerDownloadWorker.perform_async(sync_class.name, name, nil, "project", 0, force_sync_dependencies) }
-    CheckStatusWorker.perform_async(id, status == "Removed" || status == "Deprecated")
+    CheckStatusWorker.perform_async(id)
```
kjmahalingam, line 232: Removing the flag here should be appropriate, since we were just calculating the flag conditionally, and now we don't need to do that anymore.

app/workers/check_status_worker.rb
```diff
@@ -4,7 +4,7 @@ class CheckStatusWorker
   include Sidekiq::Worker
   sidekiq_options queue: :status, unique: :until_executed
 
-  def perform(project_id, _ignored = nil)
+  def perform(project_id)
```
katzj, line 7: I'd suggest leaving the ignored parameter in until all callers don't send it for a bit. There's otherwise a race condition of jobs being in sidekiq that were previously enqueued and hitting a worker running new code that then throws an error

jhan217 (approved):

PR #3117, opened 2023-05-19
Update check_status rake task to use status_checked_at

kjmahalingam:
This updated the `projects:check_status` rake task use the new `Project.status_checked_at` field in order to properly backfill existing records whose status hasn't been checked in a while, and prepare continuous prioritization of what to check in the future.

### DIFF
--- a/lib/tasks/projects.rake
+++ b/lib/tasks/projects.rake
@@ -28,24 +28,28 @@ namespace :projects do
   task :check_status, %i[max_num_of_projects_to_check batch_size] => :environment do |_task, args|
     exit if ENV["READ_ONLY"].present?
 
-    # default should spread out around 50k package jobs per platform over ~ 83 minutes
-    max_num_of_projects_to_check = args.max_num_of_projects_to_check.nil? ? 50000 : args.max_num_of_projects_to_check.to_i
-    batch_size = args.batch_size.nil? ? 5000 : args.batch_size.to_i
+    max_num_of_projects_to_check = args.max_num_of_projects_to_check.nil? ? 150000 : args.max_num_of_projects_to_check.to_i
+    batch_size = args.batch_size.nil? ? 10000 : args.batch_size.to_i
 
-    PLATFORMS_FOR_STATUS_CHECKS.each do |platform|
-      project_ids_to_check = Project.platform(platform).not_removed.where("projects.updated_at < ?", 1.week.ago).order(updated_at: :desc).select("id")
-      enqueued_projects_sum = 0
-      project_ids_to_check.limit(max_num_of_projects_to_check).in_batches(of: batch_size) do |project_ids_batch|
-        project_ids_batch.pluck(:id).each_with_index do |project_id, i|
-          CheckStatusWorker.perform_in(i, project_id)
-        end
+    # we want all nil status_checked_at first, and this ordering should achieve that
+    project_ids_to_check = Project
+      .where(platform: PLATFORMS_FOR_STATUS_CHECKS)
+      .where("projects.status_checked_at < ?", 1.week.ago)
+      .order(status_checked_at: :desc)
+      .limit(max_num_of_projects_to_check)
+      .select("id")
 
-        enqueued_projects_sum += project_ids_batch.count
-        log_string = "#{enqueued_projects_sum} of up to #{max_num_of_projects_to_check} #{platform} jobs were enqueued"
-
-        Rails.logger.info("logger: #{log_string}")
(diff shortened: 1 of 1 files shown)


lib/tasks/projects.rake
```diff
@@ -28,24 +28,28 @@ namespace :projects do
   task :check_status, %i[max_num_of_projects_to_check batch_size] => :environment do |_task, args|
     exit if ENV["READ_ONLY"].present?
 
-    # default should spread out around 50k package jobs per platform over ~ 83 minutes
-    max_num_of_projects_to_check = args.max_num_of_projects_to_check.nil? ? 50000 : args.max_num_of_projects_to_check.to_i
-    batch_size = args.batch_size.nil? ? 5000 : args.batch_size.to_i
-
-    PLATFORMS_FOR_STATUS_CHECKS.each do |platform|
```
kjmahalingam, line 35: With the new prioritization method of looking at dates, it makes more sense to just look at all platform that are allowed simultaneously, rather than iterating over them.

lib/tasks/projects.rake
```diff
@@ -71,25 +75,6 @@ namespace :projects do
     end
   end
 
-  desc "Check to see if projects are still removed/deprecated"
-  task check_removed_status: :environment do
```
kjmahalingam, line 75: Since we are no longer filtering for `not_removed` in the `check_status` task, we can remove this task entirely.

lib/tasks/projects.rake
```diff
@@ -28,24 +28,27 @@ namespace :projects do
...
-        project_ids_batch.pluck(:id).each_with_index do |project_id, i|
-          CheckStatusWorker.perform_in(i, project_id)
-        end
+    max_num_of_projects_to_check = args.max_num_of_projects_to_check.nil? ? 150000 : args.max_num_of_projects_to_check.to_i
+    batch_size = args.batch_size.nil? ? 10000 : args.batch_size.to_i
+
+    # we want all nil status_checked_at first, and this ordering should achieve that
+    project_ids_to_check = Project
+      .where(platform: PLATFORMS_FOR_STATUS_CHECKS)
+      .order(status_checked_at: :desc)
```
kjmahalingam, line 37: I just realized that I think this ordering might be incorrect. I think we want to actually have the `status_checked_at` in `asc` order to get the oldest ones. That then means there needs to be some additional logic to still prioritize `nil` values first. I will fix that.

kjmahalingam: Should be fixed!

lib/tasks/projects.rake
```diff
@@ -28,24 +28,26 @@ namespace :projects do
...
-      enqueued_projects_sum = 0
-      project_ids_to_check.limit(max_num_of_projects_to_check).in_batches(of: batch_size) do |project_ids_batch|
-        project_ids_batch.pluck(:id).each_with_index do |project_id, i|
-          CheckStatusWorker.perform_in(i, project_id)
-        end
+    max_num_of_projects_to_check = args.max_num_of_projects_to_check.nil? ? 150000 : args.max_num_of_projects_to_check.to_i
+    batch_size = args.batch_size.nil? ? 10000 : args.batch_size.to_i
+
+    project_ids_to_check = Project
+      .where(platform: PLATFORMS_FOR_STATUS_CHECKS)
```
jhan217, line 35: Thoughts about keeping the filter for looking at things older than 1 week per existing clause on line 36?

lib/tasks/projects.rake
```diff
@@ -28,24 +28,26 @@ namespace :projects do
...
-        enqueued_projects_sum += project_ids_batch.count
-        log_string = "#{enqueued_projects_sum} of up to #{max_num_of_projects_to_check} #{platform} jobs were enqueued"
+      enqueued_projects_sum += project_ids_batch.count
+      log_string = "#{enqueued_projects_sum} of up to #{max_num_of_projects_to_check} jobs were enqueued"
 
-        Rails.logger.info("logger: #{log_string}")
-        puts "puts: #{log_string}"
-      end
+      Rails.logger.info("logger: #{log_string}")
+      puts "puts: #{log_string}"
```
jhan217, line 50: The `puts` is probably not needed anymore, as the Rails logger should be configured properly.

lib/tasks/projects.rake
kjmahalingam: I can do that, though given the volume of items to check, I don't *think* that really changes anything in terms of what's processed. But it won't really hurt either way, so I'll put it back!

jhan217 (approved):

PR #3118, opened 2023-05-22
Fix platform checking for projects in task, add specs

kjmahalingam:


lib/tasks/projects.rake
```diff
@@ -28,6 +26,10 @@ namespace :projects do
   task :check_status, %i[max_num_of_projects_to_check batch_size] => :environment do |_task, args|
     exit if ENV["READ_ONLY"].present?
 
+    PLATFORMS_FOR_STATUS_CHECKS = %w[pypi npm rubygems packagist cpan clojars cocoapods hackage cran pub elm dub]
+      .map { |platform| PackageManager::Base.format_name(platform) }
```
kjmahalingam, line 30: We needed to add case sensitivity checking here, since that was the issue filtering out all records before.

lib/tasks/projects.rake
```diff
@@ -40,14 +42,16 @@ namespace :projects do
 
     enqueued_projects_sum = 0
     project_ids_to_check.in_batches(of: batch_size) do |project_ids_batch|
+      enqueued_projects_sum += project_ids_batch.count
```
kjmahalingam, line 45: I noticed the batch count was reset to 0 after being iterated through, so the logs were inaccurate. I have moved it here so the count is correct.

lib/tasks/projects.rake
```diff
@@ -40,14 +42,16 @@ namespace :projects do
...
       project_ids_batch.pluck(:id).each_with_index do |project_id, i|
         CheckStatusWorker.perform_in(i, project_id)
       end
 
-      enqueued_projects_sum += project_ids_batch.count
       log_string = "#{enqueued_projects_sum} of up to #{max_num_of_projects_to_check} jobs were enqueued"
 
-      Rails.logger.info(log_string)
+      Rails.logger.info("logger: #{log_string}")
+      puts("puts: #{log_string}")
```
kjmahalingam, line 54: Adding back the `puts` should be unnecessary but not harmful, and we weren't sure about whether the logging was working correctly, so this will help get more clarity.

macowie: I've noticed this logger issue on this application before, I made a ticket to follow up. This is all good for now.

macowie (approved):
Lavender Green Tea Matcha

macowie (approved):
Lavender Green Tea Matcha

PR #3122, opened 2023-05-30
Add optional field to dependencies for maven, add optional to DependencySerializer

kjmahalingam:
Uses the new `optional` field (added [here](https://github.com/librariesio/bibliothecary/pull/572) and [here](https://github.com/librariesio/libraries.io/pull/3121)) to add this to the `Dependency.optional` for Maven records.

Adds the `Dependency.optional` field to `DependencySerializer` so it is present in API responses.


macowie (approved):
lgtm

PR #3125, opened 2023-05-31
Backfill "kind" of Pypi dependencies

papiro:
~~This task is meant to batch approx 500k affected versions into 4 updates per second (4 Pypi api-calls per second) over the course of 1.75 days~~

This task would still take 1.75 days, but only because the main query which scans dependencies needs to be done in batches to avoid timing out. The actual number of affected versions is way less than I originally thought. I thought 1/10 versions needed to be updated but it's actually more like 1/1000. So the actual number of Pypi api-calls won't be very much, ~5k spaced over 1.75 days.

### DIFF
--- a/app/models/package_manager/pypi.rb
+++ b/app/models/package_manager/pypi.rb
@@ -144,21 +144,40 @@ module PackageManager
       [name, requirement]
     end
 
+    def self.parse_requirement_extras(requirements)
+      parsed = requirements
+                 .split(/(?:or )?extra\s*==/)
+                 .map { |part| part.delete("'\"\\ ;") }
+
+      (requirement, *extras) = parsed
+
+      [requirement, extras]
+    end
+
     def self.dependencies(name, version, _mapped_project = nil)
       api_response = get("https://pypi.org/pypi/#{name}/#{version}/json")
       deps = api_response.dig("info", "requires_dist") || []
       source_info = api_response.fetch("urls", [])
       Rails.logger.warn("Pypi sdist (no deps): #{name}") unless source_info.any? { |rel| rel["packagetype"] == "bdist_wheel" }
 
-      deps.map do |dep|
+      deps.flat_map do |dep|
         dep_name, requirements = parse_pep_508_dep_spec(dep)
-        {
+        requirement, extras = parse_requirement_extras(requirements)
+
+        mapped_dep = {
           project_name: dep_name,
-          requirements: requirements.blank? ? "*" : requirements,
-          kind: "runtime",
+          requirements: requirement.blank? ? "*" : requirement,
           optional: false,
           platform: self.name.demodulize,
         }
+
+        if extras.present?
+          extras.map do |extra|
+            mapped_dep.merge(kind: extra)
+          end
+        else
+          mapped_dep.merge(kind: "runtime")
+        end
       end
     end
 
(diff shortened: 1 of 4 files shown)


app/models/package_manager/pypi.rb
```diff
@@ -13,6 +13,11 @@ class Pypi < Base
     PYPI_PRERELEASE = /(a|b|rc|dev)[0-9]+$/.freeze
     # Adapted from https://peps.python.org/pep-0508/#names
     PEP_508_NAME_REGEX = /^([A-Z0-9][A-Z0-9._-]*[A-Z0-9]|[A-Z0-9])/i.freeze
+    PEP_508_ENVIRONMENT_MARKERS = "python_version|python_full_version|" +
+      "os_name|sys_platform|platform_release|" +
+      "platform_system|platform_version|" +
+      "platform_machine|platform_python_implementation|" +
+      "implementation_name|implementation_version|extra"
```
mpace965, line 20:
```suggestion
    PEP_508_ENVIRONMENT_MARKERS = %w[
      python_version python_full_version os_name
      sys_platform platform_release platform_system
      platform_version platform_machine platform_python_implementation
      implementation_name implementation_version extra
    ].freeze
    
    PEP_508_ENVIRONMENT_MARKER_REGEX = PEP_508_ENVIRONMENT_MARKERS.join("|")
```

saves you from splitting them back out later

lib/tasks/one_off.rake
```diff
@@ -118,9 +118,59 @@ namespace :one_off do
...
+      .split("|")
+      .map { |em| "%#{em}%" }
+      .join(", ")
+
+    pypi_versions.in_batches(of: batch_size).each_with_index do |batch_versions, batch_versions_index|
+      affected_versions = Dependency
+        .where(version: batch_versions)
+        .joins(version: :project)
+        .where("dependencies.requirements LIKE any (array[#{environment_markers}])")
+        .select("versions.number as project_version, projects.name as project_name")
```
mpace965, line 156: How's the performance on this query? Looks like there is an index on `dependencies.version_id`, so that's good. We could potentially offload everything in this block to a new worker

lib/tasks/one_off.rake
```diff
@@ -118,9 +118,59 @@ namespace :one_off do
...
+    # There's ~5 million Pypi versions
+    # Based on a sampling of 5000 Pypi versions, ~1/10 contain environment markers
+    # 4 Pypi API calls per second = 240/min
+    # leaving a buffer of 40 for any batches which contain more, 200/min is conservative.
+    # 200/min = 12000/hour = 288k/day
+    # At that rate it will take ~1.75 days
+
+    pypi_versions = Version.where(
+      project_id: Project.where(platform: "Pypi")
+    )
```
mikeyoung85, line 138: Do you know what SQL is getting generated for this? Is it using an `IN` clause with 5 million IDs?

lib/tasks/one_off.rake
papiro: I actually had this query completely wrong. It was duplicating versions, once for each dependency with environment markers. I think the performance of this query on a batch is ok. I made it so we can pass in batch size and the start so we can tweak the running of it as needed. I think putting in a `sleep` would have essentially the same effect as moving it to its own worker no? Moving it to its own worker seems like a lot of work for a one-off so I'd like to avoid it if I can.

PR #3128, opened 2023-06-06
Fix environment marker parsing and increase batch-size for backfill

papiro:


spec/tasks/one_off_spec.rb
```diff
@@ -29,7 +29,7 @@
     end
   end
 
-  describe "backfill kind of Pypi dependencies" do
+  xdescribe "backfill kind of Pypi dependencies" do
```
mpace965, line 32: are we intending to skip this?

app/models/package_manager/pypi.rb
```diff
@@ -154,7 +154,7 @@ def self.parse_pep_508_dep_spec(dep)
 
     def self.parse_environment_markers(requirement)
       version_regex = "(?<version>[^;]+)"
-      environment_markers_regex = "(?<environment_markers>#{PEP_508_ENVIRONMENT_MARKER_REGEX}\s*==.+)"
+      environment_markers_regex = "(?<environment_markers>(?:#{PEP_508_ENVIRONMENT_MARKER_REGEX}).+)"
```
mpace965, line 157: what is the purpose of the non-capturing group here? I think maybe the `.+` is just matching everything?

spec/tasks/one_off_spec.rb
papiro: yeah, I left it in the code because it's useful to have to make any tweaks to the rake task, but I think it's wasteful to run it as part of the test suite on every commit into main.

app/models/package_manager/pypi.rb
papiro: Ohh, I added the non-capture notation because I need the parentheses there for separation purposes, but didn't want it to be interpreted as a group. If I remove it though, all the tests pass. I think I was just being explicit about the intention of the parentheses. Should I remove it?

With regards to splitting on `;`, I had hoped that I could do that (i think at one point my solution was basically that), but there are requirements which come out like `extra == \"pipfile-deprecated-finder\" or extra == \"requirements-deprecated-finder\"`, which have no semicolon. But I think I see a way we could make it work. We are parsing out the leading semicolon on versionless requirements here: https://github.com/librariesio/libraries.io/blob/main/app/models/package_manager/pypi.rb#L151 I can see if we can just not do that and simply split on semicolon to get the kind. Do you see any drawbacks with simplifying the approach as such? I can't think of any.

mpace965: I think the implementation you proposed would be preferable. Maybe it could all be rolled into `. parse_pep_508_dep_spec` and return `[name, version, requirement]`. The `requirement` it returns right now is a bit of a misnomer by our current understanding.

mpace965 (approved):

app/models/package_manager/pypi.rb
mpace965: Thanks for this. Regex can be so terse. I find this implementation much easier to read!

mpace965 (approved):

PR #3140, opened 2023-07-05
Remove Parentheses from PyPI API Dependency Requirements When Parsing Results

mikeyoung85:
Dependency version requirements being returned from the PyPI API often have parentheses surrounding them which breaks the tree resolver when trying to find the latest version that fulfills each requirement. This PR would strip the parentheses from the requirements text before saving the dependency rows. We can run a backfill to fix old data or leave the fix in `dependency_checks.rb` to remove them when calling `semantic_requirements` on a `Dependency`.

### DIFF
--- a/app/models/concerns/dependency_checks.rb
+++ b/app/models/concerns/dependency_checks.rb
@@ -23,7 +23,7 @@ module DependencyChecks
       numbers = requirements.split("<= v")
       ">=#{numbers[0].strip} #{numbers[1].strip}"
     when "pypi"
-      requirements&.remove(/[()]/)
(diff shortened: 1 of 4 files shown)


mpace965 (commented):
This does seem consistent with the [PEP 508](https://peps.python.org/pep-0508/#grammar) spec (see the `versionspec` rule, which thankfully only allows a single level of nesting around a version specifier)

Is it feasible to add a regression test for this scenario in the tree resolver specs?

tiegz (approved):

mpace965 (approved):

PR #3142, opened 2023-07-06
Add Indexes to dependency and repository_dependency on created_at date

mikeyoung85:
This PR is to help out the [Projects Dependencies Link](https://github.com/librariesio/libraries.io/blob/d68a5ffb6f032d1ab9f91c3c7bf35ce45a32d973/lib/tasks/projects.rake#L17-L22) daily job which has been failing to complete for a while now. The main reason this job fails is that it is impossible to query the `dependencies` table because of its large size and it is querying an unindexed column. The PR adds an index for `created_at` and scopes the values down to remove the time portion and index it as a Date so that values are grouped up a bit in the index and updates the jobs to query by Date and instead of DateTime.

This migration is something that should be run manually outside of our normal deployment since it will take some time for the database to add the indexes.

### DIFF
--- a/db/migrate/20230705195409_add_created_at_index_to_dependencies.rb
+++ b/db/migrate/20230705195409_add_created_at_index_to_dependencies.rb
@@ -2,23 +2,23 @@ class AddCreatedAtIndexToDependencies < ActiveRecord::Migration[5.2]
   disable_ddl_transaction!
 
   def up
-    unless index_name_exists?(:dependencies, "index_dependencies_on_created_at_date")
-      add_index :dependencies, "(created_at::date)", name: "index_dependencies_on_created_at_date", algorithm: :concurrently
+    unless index_name_exists?(:dependencies, "index_dependencies_on_project_created_at_date")
+      add_index :dependencies, "project_id, (created_at::date)", name: "index_dependencies_on_project_created_at_date", algorithm: :concurrently
     end
 
-    unless index_name_exists?(:repository_dependencies, "index_repository_dependencies_on_created_at_date")
-      add_index :repository_dependencies, "(created_at::date)", name: "index_repository_dependencies_on_created_at_date", algorithm: :concurrently
+    unless index_name_exists?(:repository_dependencies, "index_repository_dependencies_on_project_created_at_date")
+      add_index :repository_dependencies, "project_id, (created_at::date)", name: "index_repository_dependencies_on_project_created_at_date", algorithm: :concurrently
     end
 
   end
 
   def down
-    if index_name_exists?(:dependencies, "index_dependencies_on_created_at_date")
-      remove_index :dependencies, "(created_at::date)"
+    if index_name_exists?(:dependencies, "index_dependencies_on_project_created_at_date")
+      remove_index :dependencies, name: "index_dependencies_on_project_created_at_date"
     end
 
-    if index_name_exists?(:repository_dependencies, "index_repository_dependencies_on_created_at_date")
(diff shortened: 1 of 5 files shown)


db/migrate/20230705195409_add_created_at_index_to_dependencies.rb
```diff
@@ -0,0 +1,24 @@
+class AddCreatedAtIndexToDependencies < ActiveRecord::Migration[5.2]
+  disable_ddl_transaction!
+
+  def up
+    unless index_name_exists?(:dependencies, "index_dependencies_on_created_at_date")
+      add_index :dependencies, "(created_at::date)", name: "index_dependencies_on_created_at_date", algorithm: :concurrently
```
mikeyoung85, line 6: The other option I thought of for indexing that might suit the query is to index by `project_id, created_at::date` but my local testing showed this index worked better, but I don't have the same volume of data so the plan may be different on the prod Libraries database.

tiegz: just out of curiosity, have you tried this locally and tested EXPLAIN? I'm curious if the psql query planner would pick the right index, based on the date type-coercion.

tiegz: nevermind, just found some docs that say it'll work as long as the queried value is cast in the same way 👍🏻

tiegz (approved):

db/migrate/20230705195409_add_created_at_index_to_dependencies.rb
mikeyoung85: With project_id, created_at index

```=> EXPLAIN for: SELECT "dependencies".* FROM "dependencies" WHERE (created_at::date > date('2023-07-05 19:18:21.170924')) AND "dependencies"."project_id" IS NULL AND (dependencies.project_name <> '') /*application:Libraries*/
                                                 QUERY PLAN
------------------------------------------------------------------------------------------------------------
 Bitmap Heap Scan on dependencies  (cost=4011.42..24251.83 rows=292154 width=68)
   Recheck Cond: ((project_id IS NULL) AND ((created_at)::date > '2023-07-05'::date))
   Filter: ((project_name)::text <> ''::text)
   ->  Bitmap Index Scan on index_dependencies_on_created_at_date  (cost=0.00..3938.38 rows=292195 width=0)
         Index Cond: ((project_id IS NULL) AND ((created_at)::date > '2023-07-05'::date))
```

With created_at only index

```
=> EXPLAIN for: SELECT "dependencies".* FROM "dependencies" WHERE (created_at::date > date('2023-07-05 19:15:22.701206')) AND "dependencies"."project_id" IS NULL AND (dependencies.project_name <> '') /*application:Libraries*/
                                                    QUERY PLAN
------------------------------------------------------------------------------------------------------------------
 Index Scan using index_dependencies_on_created_at_date on dependencies  (cost=0.43..3927.07 rows=80741 width=68)
   Index Cond: ((created_at)::date > '2023-07-05'::date)
   Filter: ((project_id IS NULL) AND ((project_name)::text <> ''::text))
```

mikeyoung85: I yolo'd this index creation on production and the `project_id, created_at` index seems to at least be somewhat usable. The `created_at` on its own still results in queries timing out. I've updated the branch with this new index.

papiro (approved):

PR #3145, opened 2023-07-10
Update the list of package managers to check the status of

jhan217:
Updating the list of package managers we check the statuses of. Confirmed on a console that they all have a url string coming out of `check_status_url`.


kjmahalingam (approved):
🎉

PR #3149, opened 2023-07-13
Retrieve PyPI RSS feed for missing pubdates; refactor code

johnbintz-tidelift:
Feature-wise, this adds the ability to retrieve a PyPI package's RSS feed to gather publishing dates when the JSON ASPI has no information on them. It also adds an integration test to ensure version retrieval works as intended.

Structure-wise, this breaks apart the PyPI version processing code into a few classes which makes testing the individual pieces a lot easier. It also breaks out the shared data retrieval code into a module for use outside of PackageManager::Base and its subclasses.

### DIFF
--- a/app/models/package_manager/base.rb
+++ b/app/models/package_manager/base.rb
@@ -2,6 +2,8 @@
 
 module PackageManager
   class Base
+    extend CanRequest
+
     COLOR = "#fff"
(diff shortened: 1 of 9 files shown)


johnbintz-tidelift:
Closing to restructure code.

PR #3151, opened 2023-07-17
Use PyPI RSS feed to fill in missing published dates

johnbintz-tidelift:
This breaks out much of the data retrieval code for the PyPI package manager, centralizes version processing in a set of objects, and adds the ability to load published dates from a package's RSS feed.


macowie (approved):

djpowers (approved):

app/models/package_manager/pypi/json_api_project.rb
```diff
@@ -0,0 +1,109 @@
...
+          homepage_url
+        )
+      end
+
+      def releases
+        JsonApiProjectReleases.new(
+          @data["releases"].map do |version_number, details|
+            JsonApiProjectRelease.new(
+              version_number: version_number,
+              published_at: self.class.release_data_published_at(details)
```
jhan217, line 89: Could `release_data_published_at` be an instance method? I only see it being used here

app/models/package_manager/pypi/version_processor.rb
```diff
@@ -0,0 +1,53 @@
...
+        @project_releases.map do |project_release|
+          next @known_versions[project_release.version_number] if @known_versions.key?(project_release.version_number)
+
+          version_number = project_release.version_number
+          original_license = json_api_single_release_for_version(version_number).license
+          rss_api_release = rss_api_releases_hash[version_number]
+          published_at = project_release.published_at || rss_api_release&.published_at
+
+          {
+            number: project_release.version_number,
```
jhan217, line 20: Could this just be `version_number`?

jhan217 (approved):
LGTM -- thanks for modularizing all of this, hopefully it will be easier to build on in the future! :)

app/models/package_manager/pypi/json_api_project.rb
johnbintz-tidelift: I would want to make it a private method and then add tests to `#releases` to ensure all the private method's cases are handled.

johnbintz-tidelift: Scratch that, this is going to need more refactoring to be structured correctly.

jhan217 (approved):

PR #3152, opened 2023-07-17
Add support for detecting removed ('retracted') Go versions

macowie:


kjmahalingam (approved):
This looks good to me, although I'm not very familiar with Golang stuff, so it might be best for someone else who's more familiar to also take a look.

app/models/package_manager/go/go_mod_parser.rb
```diff
@@ -0,0 +1,82 @@
...
+          end_range = parse_version(right)
+
+          results << [start_range, end_range] if start_range && end_range
+        else
+          single_version = parse_version(line)
+          results << single_version if single_version
+        end
+      end
+    end
+  end
```
johnbintz-tidelift, line 81: Since there's no reason to expose these to the rest of the app, these should become private methods.

app/models/package_manager/go/go_mod_parser.rb
```diff
@@ -0,0 +1,82 @@
...
+    # Scans contents for retract directives parses their contents
+    #
+    # @return [Array<String, Array<String, String>>] an array consisting of:
+    #   a version number (String) or
+    #   a range represented by a two element array: [range_start_number, range_end_number]
+    def retracted_version_ranges
+      return @retracted_version_ranges if @retracted_version_ranges
+
+      retract_specs = []
+      retract_specs.concat stripped_contents.scan(RETRACT_WITH_PARENS).flatten
```
johnbintz-tidelift, line 24: Do these need to be `flatten`ed individually, or can they all be `concat`ted into the array and `flatten`ed together?

app/models/package_manager/go/go_mod_parser.rb
```diff
@@ -0,0 +1,82 @@
...
+
+      @retracted_version_ranges = retract_specs.flat_map do |retract_spec|
+        parse_retract(retract_spec)
+      end
+    end
+
+    def canonical_module_name
+      module_line = stripped_contents.lines.find { |line| line.match(MODULE_REGEX) }
+
+      module_line&.scan(MODULE_REGEX)&.dig(0, 0)
```
johnbintz-tidelift, line 35: Can you add a comment showing what this `scan` is looking for in an example string?

app/models/package_manager/go/go_mod_parser.rb
```diff
@@ -0,0 +1,82 @@
...
+      retract_specs.concat stripped_contents.scan(RETRACT_WITH_PARENS).flatten
+      retract_specs.concat stripped_contents.scan(RETRACT_WITHOUT_PARENS).flatten
+
+      @retracted_version_ranges = retract_specs.flat_map do |retract_spec|
+        parse_retract(retract_spec)
+      end
+    end
+
+    def canonical_module_name
+      module_line = stripped_contents.lines.find { |line| line.match(MODULE_REGEX) }
```
johnbintz-tidelift, line 33: If you `return` early here if `module_line` is `nil`, you can skip needing all of the existence operators.

app/models/package_manager/go/go_mod.rb
```diff
@@ -0,0 +1,25 @@
...
+        if version_or_range.is_a?(Array)
+          begin_range, end_range = version_or_range
+          SemanticRange.gte?(given_version, begin_range) &&
+            SemanticRange.lte?(given_version, end_range)
+        else
+          SemanticRange.satisfies?(given_version, version_or_range)
+        end
+      end
+    end
+  end
```
johnbintz-tidelift, line 24: This entire class is acting on a single `GoModParser` object, calling the methods on that object for everything `retracted?` does as well as passing along delegated values from `GoModParser`. WDYT about moving this method to `GoModParser` and removing `GoMod` entirely?

app/models/package_manager/go.rb
```diff
@@ -299,5 +267,36 @@ def self.canonical_module_name(name)
...
+
+      version = latest_version_number(name) if version.nil?
+      return if version.nil?
+
+      contents = get_raw("#{PROXY_BASE_URL}/#{encode_for_proxy(name)}/@v/#{encode_for_proxy(version)}.mod")
+
+      Rails.logger.info "[Unable to fetch go.mod contents] name=#{name}" if contents.blank?
+
+      contents
+    end
```
johnbintz-tidelift, line 300: If this is only being used by `fetch_mod` there's no need for them to be separate methods. `fetch_mod` can do the actual fetch, then return the `GoMod` object.

spec/models/package_manager/go/go_mod_spec.rb
```diff
@@ -0,0 +1,130 @@
...
+
+require "rails_helper"
+
+RSpec.describe PackageManager::Go::GoMod do
+  subject(:go_mod) { described_class.new(mod_contents) }
+  let(:mod_contents) { "" }
+
+  describe "#retracted_version_ranges" do
+    context "when empty" do
+      let(:mod_contents) { "" }
```
johnbintz-tidelift, line 11: There's no need to set this value again as it's the same as the value on line 7.

spec/models/package_manager/go_spec.rb
```diff
@@ -337,4 +349,49 @@
       end
     end
   end
+
+  describe "::deprecate_versions" do
```
johnbintz-tidelift, line 353:
```suggestion
  describe ".deprecate_versions" do
```

spec/models/package_manager/go_spec.rb
```diff
@@ -337,4 +349,49 @@
...
+      project.versions.create!(number: "1.0.1")
+    end
+
+    it "should mark missing versions as Removed" do
+      described_class.deprecate_versions(project, [{ number: "1.0.0", published_at: nil, original_license: nil }])
+      expect(project.reload.versions.pluck(:number, :status)).to match_array([["1.0.0", nil], ["1.0.1", "Removed"]])
+    end
+  end
+
+  describe "::canonical_module_name" do
```
johnbintz-tidelift, line 365:
```suggestion
  describe ".canonical_module_name" do
```

spec/models/package_manager/go_spec.rb
```diff
@@ -337,4 +349,49 @@
...
+    it "maps only major revision versions to module" do
+      VCR.use_cassette("go/pkg_go_dev") do
+        result = described_class.canonical_module_name(package_name)
+        expect(result).to eq(package_name)
+      end
+    end
+  end
+
+  describe "::dependencies" do
```
johnbintz-tidelift, line 374:
```suggestion
  describe ".dependencies" do
```

app/models/package_manager/go/go_mod_parser.rb
```diff
@@ -0,0 +1,82 @@
...
+      @stripped_contents ||= mod_contents
+        .to_s
+        .lines
+        .map { |line| line.gsub(/\/\/.*/, "").strip }
+        .reject(&:blank?)
+        .join("\r\n")
+    end
+
+    def parse_version(substring)
+      substring.scan(/(v\d\.\d\.\d(\w|\+|\.|-)*)/).dig(0, 0)
```
johnbintz-tidelift, line 61: Can you add an example string here as well?

app/models/package_manager/go/go_mod.rb
kjmahalingam: I refactored it all as you suggested, except I put it all into `GoMod` instead of `GoModParser` (but that's a simple renaming).

app/models/package_manager/go/go_mod_parser.rb
kjmahalingam: Couldn't make these private unless we make some other changes, as these methods are used in `app/models/package_manager/go.rb` so they are currently not private. We could make those changes if we want, or otherwise leave as-is.

app/models/package_manager/go/go_mod_parser.rb
kjmahalingam: As I wasn't the original author I'm not deeply familiar, so I wasn't comfortable writing a comment for this.

app/models/package_manager/go/go_mod_parser.rb
johnbintz-tidelift: As-is is fine.

johnbintz-tidelift (approved):

PR #3154, opened 2023-07-18
Clean up projects#sync_by_platform rake task

kjmahalingam:
This adds some additional logging between batches, and removes a long `sleep` from the iterations, because this task was taking very long to run.

### DIFF
--- a/lib/tasks/projects.rake
+++ b/lib/tasks/projects.rake
@@ -313,17 +313,14 @@ namespace :projects do
     end
 
     batch_size = 500
-    batch_wait = 5
 
     result_count = 0
 
     Project.platform(platform).find_in_batches(batch_size: batch_size).each do |projects|
-      if commit
-        projects.each(&:manual_sync)
-        sleep batch_wait
-      end
+      projects.each(&:manual_sync) if commit
 
       result_count += projects.size
+      Rails.logger.info("#{result_count} projects queued to resync so far.")
     end
 
     stats = <<~STATS


lib/tasks/projects.rake
```diff
@@ -313,17 +313,14 @@ namespace :projects do
...
-    batch_wait = 5
 
     result_count = 0
 
     Project.platform(platform).find_in_batches(batch_size: batch_size).each do |projects|
-      if commit
-        projects.each(&:manual_sync)
-        sleep batch_wait
-      end
+      projects.each(&:manual_sync) if commit
```
kjmahalingam, line 320: This queues up some sidekiq workers, but notably one of the is `PackageManagerDownloadWorker` which is marked as a critical priority queue sidekiq job. Considering that this we're on the scale of millions of records to update with this backfill, do we need to do something to address that? Is there maybe some way to change the priority?

Alternatively, we don't *technically* need to do a full manual sync. All we need to do is run the `add_versions` and `deprecate_versions` for each of these projects. So we *could* make a new sidekiq worker with a lower priority whose only job is to do that.

kjmahalingam: Additionally, we thought about putting some of this into a transaction and doing a rollback if it was a dry run, but since the actual DB interactions occur in a sidekiq worker, using a transaction here doesn't really do much of significance, so I left that out.

lib/tasks/projects.rake
```diff
@@ -313,17 +313,14 @@ namespace :projects do
...
 
     batch_size = 500
-    batch_wait = 5
 
     result_count = 0
 
     Project.platform(platform).find_in_batches(batch_size: batch_size).each do |projects|
-      if commit
-        projects.each(&:manual_sync)
-        sleep batch_wait
```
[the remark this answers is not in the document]
jhan217: If line 322 is just adding jobs to the queue, I don't think we need to stagger that since it's just putting stuff onto the Sidekiq queue, which gets run when resources become available.

lib/tasks/projects.rake
jhan217: I wouldn't want to put millions of records onto the critical queue if it's not actually time sensitive to get all millions of records refreshed ASAP. If we don't need to do a full manual sync, it could be worth trying to scope things down to just the thing we want.

macowie (approved):
I'm good with not staggering the batches here, I'd like to follow up and update `sync_from_list` to match

kjmahalingam:
Not going to immediately run this again, as I will do some additional research and experimentation beforehand, but will merge this in for additional logging and removing the sleep.

PR #3156, opened 2023-07-18
feat: [sc-34933] Libraries can detect yanked Rubygems versions we don't know about

djpowers:
This allows us to pull down yanked versions from Rubygems, which are not returned as part of the versions API endpoint.

The functionality is opt-in to avoid making unnecessary HTTP calls. The including `parse_html: true` will check the web associated gem's version pages for any that have been yanked.

Examples:
```ruby
PackageManager::Rubygems.versions({ "name" => "flowbyte-yanked" }, "flowbyte-yanked")
=> [{:number=>"1.0.0", :published_at=>"2019-05-24T14:36:49.910Z", :original_license=>["MIT"]}]
```
```ruby
PackageManager::Rubygems.versions({ "name" => "flowbyte-yanked" }, "flowbyte-yanked", parse_html: true)
=> [{:number=>"1.0.0", :published_at=>"2019-05-24T14:36:49.910Z", :original_license=>["MIT"]}, {:number=>"1.0.1", :published_at=>"2019-05-24T00:00:00-04:00", :original_license=>"", :yanked=>true}]
```


app/models/package_manager/rubygems.rb
```diff
@@ -108,5 +107,53 @@ def self.download_registry_users(name)
     def self.registry_user_url(login)
       "https://rubygems.org/profiles/#{login}"
     end
+
+    def self.parse_html_versions(raw_project)
```
kjmahalingam, line 111: Am I correct that this method is only used to collect yanked versions, and we'll never retrieve anything not yanked from this? We'll get everything that isn't yanked from `parse_json_versions`? If so, it might be nice to have some comments and/or renamings to make that clear.

kjmahalingam (approved):
Looks good! Minor nitpick in comments.

app/models/package_manager/rubygems.rb
djpowers: Correct, since the only reason to use the HTML option would be to collect yanked versions. I updated the method name to clarify this.

PR #3157, opened 2023-07-19
Query for all Dependency IDs Instead of Using Batching SQL Queries

mikeyoung85:
The query without using the batching SQL query parameters, namely `order by id asc where id > ? limit 1000`, seems to run pretty quickly and should avoid timing out. This change may result in the container hitting some memory issues, but that seems more manageable than trying to speed up the queries on these tables some more.

### DIFF
--- a/lib/tasks/projects.rake
+++ b/lib/tasks/projects.rake
@@ -17,12 +17,12 @@ namespace :projects do
   desc "Link dependencies to projects"
   task link_dependencies: :environment do
     exit if ENV["READ_ONLY"].present?
-    ids = Dependency.where("created_at::date > date(?)", 1.day.ago).without_project_id.pluck(:id)
+    ids = Dependency.where("created_at::date >= date(?)", 1.day.ago).without_project_id.pluck(:id)
     ids.in_groups_of(1000) do |group|
       Dependency.where(id: group).each(&:update_project_id)
     end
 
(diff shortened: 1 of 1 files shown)


lib/tasks/projects.rake
```diff
@@ -17,8 +17,23 @@ namespace :projects do
   desc "Link dependencies to projects"
   task link_dependencies: :environment do
     exit if ENV["READ_ONLY"].present?
-    Dependency.where("created_at::date > date(?)", 1.day.ago).without_project_id.find_each(&:update_project_id)
-    RepositoryDependency.where("created_at::date > date(?)", 1.day.ago).without_project_id.find_each(&:update_project_id)
+    ids = Dependency.where("created_at::date >= date(?)", 1.day.ago).without_project_id.pluck(:id)
+    Rails.logger.info("Found #{ids.count} Dependency IDs")
+
+    total_slices = (ids.count / 1000.to_f).ceil
+    ids.each_slice(1000).with_index do |group, index|
```
papiro, line 24: Is it that `ids` doesn't materialize until it gets to here, and then it materializes with the `limit 1000`?

mikeyoung85: No the IDs will be materialized right away from the query so there will be an array that could contain thousands of integers. So far that doesn't seem to be too big of an issue. The slice size here will determine how many `Dependency` and `RepositoryDependency` objects get loaded though, so tuning this up/down should have a bigger impact.

PR #3161, opened 2023-07-19
Try to find and display matching project if exact match not found

johnbintz-tidelift:


macowie (approved):
Could also avoid the platform ILIKEs with

```
Project.platform(platform).where("name ILIKE ?", name).first
```

Must have missed this one when I was updating

macowie (approved):

PR #3164, opened 2023-07-21
feat: [sc-31451] Extract PackageManager VersionUpdater to separate classes

johnbintz-tidelift:
As I was working through how to update PyPI versions, I made an attempt to extract more of the PackageManager code to separate classes, as I had already started doing with PyPI.

### DIFF
--- a/app/models/package_manager/base/incoming_version.rb
+++ b/app/models/package_manager/base/incoming_version.rb
@@ -1,0 +1,34 @@
+module PackageManager
+  class Base
+    class IncomingVersion
+      attr_reader :number
+
+      def initialize(
+        number:,
+        published_at:,
+        runtime_dependencies_count:,
+        original_license:,
+        repository_sources:,
+        status:
+      )
+        @number = number
+        @published_at = published_at
+        @runtime_dependencies_count = runtime_dependencies_count
+        @original_license = original_license
+        @repository_sources = repository_sources
+        @status = status
+      end
+
+      def to_h
+        {
+          number: @number,
+          published_at: @published_at,
+          runtime_dependencies_count: @runtime_dependencies_count,
+          original_license: @original_license,
+          repository_sources: @repository_sources,
+          status: @status,
+        }.reject { |_k, v| v.nil? }
+      end
+    end
+  end
+end
--- a/app/models/package_manager/base/version_updater.rb
+++ b/app/models/package_manager/base/version_updater.rb
@@ -3,30 +3,34 @@ module PackageManager
     class VersionUpdater
       def initialize(
         project:,
-        incoming_version:,
-        repository_source:
+        api_version_to_upsert:,
+        new_repository_source:
       )
         @project = project
-        @incoming_version = incoming_version
-        @repository_source = repository_source
+        @api_version_to_upsert = api_version_to_upsert
+        @new_repository_source = new_repository_source
       end
 
-      def execute
+      def upsert_version_for_project!
         db_project_version.skip_save_project = true
-        db_project_version.assign_attributes(@incoming_version.to_h)
+        db_project_version.assign_attributes(@api_version_to_upsert.to_version_model_attributes)
 
-        if @repository_source
+        if @new_repository_source
           existing_repository_sources = Set.new(db_project_version.repository_sources)
-          existing_repository_sources << @repository_source
+          existing_repository_sources << @new_repository_source
           db_project_version.repository_sources = existing_repository_sources.to_a
         end
 
         db_project_version.save!
+      # It's possible for this same code to be running, at the same time, on
+      # the same version, where both are trying to add a new version at the
+      # same time. These rescues catch those possiblities so that we can be
+      # alerted when they're happening.
       rescue ActiveRecord::RecordNotUnique => e
         # Until all package managers support version-specific updates, we'll have this race condition
         # of 2+ jobs trying to add versions at the same time.
         if e.message =~ /PG::UniqueViolation/
-          Rails.logger.info "[DUPLICATE VERSION 1] platform=#{@project.platform} name=#{@project.name} version=#{@incoming_version.number}"
+          Rails.logger.info "[DUPLICATE VERSION 1] platform=#{@project.platform} name=#{@project.name} version=#{@api_version_to_upsert.version_number}"
         else
           raise e
         end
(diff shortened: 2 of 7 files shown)


kjmahalingam (approved):
Awesome!

app/models/package_manager/base/incoming_version.rb
```diff
@@ -0,0 +1,34 @@
...
+        repository_sources:,
+        status:
+      )
+        @number = number
+        @published_at = published_at
+        @runtime_dependencies_count = runtime_dependencies_count
+        @original_license = original_license
+        @repository_sources = repository_sources
+        @status = status
+      end
```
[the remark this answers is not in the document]
johnbintz-tidelift: This request is beyond the scope.

app/models/package_manager/base/incoming_version.rb
```diff
@@ -0,0 +1,34 @@
+module PackageManager
+  class Base
+    class IncomingVersion
+      attr_reader :number
+
+      def initialize(
+        number:,
```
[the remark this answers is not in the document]
johnbintz-tidelift: The `Version` model throughout Libraries uses the field `number`, so I'm going to keep the word "number" in the names and add "version" to them.

app/models/package_manager/base/version_updater.rb
```diff
@@ -0,0 +1,50 @@
...
+        project:,
+        incoming_version:,
+        repository_source:
+      )
+        @project = project
+        @incoming_version = incoming_version
+        @repository_source = repository_source
+      end
+
+      def execute
```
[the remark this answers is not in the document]
johnbintz-tidelift: Can you be specific as to how this code is different from other existing uses?

spec/models/package_manager/base/version_updater_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+    let!(:db_project_version) { db_project.versions.create(number: version_number, published_at: nil, repository_sources: ["a"]) }
+    let(:incoming_version) { instance_double(PackageManager::Base::IncomingVersion, number: version_number, to_h: { published_at: published_at }) }
+
+    let(:project_name) { "name" }
+    let(:version_number) { "1.0.0" }
+    let(:published_at) { Time.zone.now }
+
+    let(:version_updater) { described_class.new(project: db_project, incoming_version: incoming_version, repository_source: "b") }
+
+    it "updates the version" do
```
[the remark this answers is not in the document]
johnbintz-tidelift: I don't know how to test for this. This seems to be an issue with running this code in a multi-process environment where two copies of the app are in this same block of code, at different points in the code, and one beats the other to updating the database.

spec/models/package_manager/base/version_updater_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+
+      let(:error_class) { ActiveRecord::RecordNotUnique }
+      let(:exception_details) { "PG::UniqueViolation" }
+
+      context "with activerecord not unique error" do
+        context "with postgresql violation" do
+          it "logs a message" do
+            expect(logger).to receive(:info).with(/DUPLICATE VERSION 1/)
+
+            version_updater.upsert_version_for_project!
```
jhan217, line 60: There's a subtle difference between expecting something to happen before invoking it, I don't remember what the edge case was but asserting the expectation after running it was the better option, in addition to it making a bit more sense when reading.

```suggestion
          it "logs a message" do
            version_updater.upsert_version_for_project!
            expect(logger).to have_received(:info).with(/DUPLICATE VERSION 1/)
```

spec/models/package_manager/base/version_updater_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+
+      let(:logger) { instance_double(ActiveSupport::Logger) }
+
+      before do
+        allow(db_project.versions).to receive(:find_or_initialize_by).with(number: version_number).and_return(db_project_version_stub)
+        allow(db_project_version_stub).to receive(:save!).and_raise(error_class, exception_details)
+
+        allow(Rails).to receive(:logger).and_return(logger)
+        allow(logger).to receive(:info)
+      end
```
jhan217, line 50: I think you can just do 
```
allow(Rails.logger).to receive(:info)
```
and then use that as the object to assert expectations on
```
expect(Rails.logger).to have_received(:info).with(...)
```
instead of spoofing the logger and stubbing its methods out.

spec/models/package_manager/base/version_updater_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+
+        before do
+          allow(db_project_version_stub).to receive(:errors).and_return(errors)
+        end
+
+        context "with already taken" do
+          it "logs a message" do
+            expect(logger).to receive(:info).with(/DUPLICATE VERSION 2/)
+
+            version_updater.upsert_version_for_project!
```
jhan217, line 96:
```suggestion
          it "logs a message" do
            version_updater.upsert_version_for_project!
            expect(logger).to have_received(:info).with(/DUPLICATE VERSION 2/)
```

spec/models/package_manager/base/version_updater_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+          )
+        end
+
+        let(:message_text) { "Number has already been taken" }
+
+        before do
+          allow(db_project_version_stub).to receive(:errors).and_return(errors)
+        end
+
+        context "with already taken" do
```
jhan217, line 92:
```suggestion
        context "when the version_number is already in the database" do
```

spec/models/package_manager/base/version_updater_spec.rb
```diff
@@ -0,0 +1,110 @@
...
+        allow(db_project_version_stub).to receive(:save!).and_raise(error_class, exception_details)
+
+        allow(Rails).to receive(:logger).and_return(logger)
+        allow(logger).to receive(:info)
+      end
+
+      let(:error_class) { ActiveRecord::RecordNotUnique }
+      let(:exception_details) { "PG::UniqueViolation" }
+
+      context "with activerecord not unique error" do
```
jhan217, line 55:
```suggestion
      context "when the version record already exists in the database" do
```

jhan217 (approved):
Looks good, just some minor cleanup in the spex.

spec/models/package_manager/base/version_updater_spec.rb
johnbintz-tidelift: This doesn't work, I already tried it. I needed to create a logger and have `Rails` use that.

PR #3170, opened 2023-07-27
Bump updated_ats when updating versions to Removed status

jhan217:
Follow up to https://github.com/librariesio/libraries.io/pull/3169#discussion_r1275050493

This should bump the updated_at timestamps on a version only when the version's status changes to 'Removed'.

I only updated the Go spec since it was representative of the other package managers in that block; the method should probably be either overwritten by their respective package managers or spex should exist for the `base.deprecate_versions` method, but I didn't want to get into a big refactor on this.

### DIFF
--- a/app/models/package_manager/base.rb
+++ b/app/models/package_manager/base.rb
@@ -167,6 +167,7 @@ module PackageManager
 
     def self.deprecate_versions(db_project, version_hashes)
       removed_status = "Removed"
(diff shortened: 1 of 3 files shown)


kjmahalingam (approved):
Looks good! And in addition to updating the timestamp, thanks for adding the check so we don't perform an update if it's already removed.

djpowers (approved):

PR #3174, opened 2023-07-28
Update Libraries Pypi parser to handle different "homepage" variations

djpowers:
Libraries will display a "Homepage" link for each package where we have such data available.
Libraries example: https://libraries.io/pypi/PyICU
Pypi source: https://pypi.org/project/PyICU/ (note the "Homepage" link under "Project links")

In the JSON API response we look for `home_page` or  `project_urls` => `Homepage` to populate this field in Libraries.

However, Pypi is not consistent with this field, and different projects will refer to this field in different ways. I've also observed:
- `project_urls` => `Home`
- `project_urls` => `homepage`
- `project_urls` => `HomePage`

Examples from Pypi where we are missing homepage data on Libraries:
- https://pypi.org/project/jupyter-server-fileid/
- https://pypi.org/project/nlbq/
- https://pypi.org/project/objverify/
- https://pypi.org/project/attacus/
- https://pypi.org/project/coolapi/
- https://pypi.org/project/songwhip/
- https://pypi.org/project/smartdoor/
- https://pypi.org/project/bit8/

This changes ensures we parse out the few additional fields that are also used on PyPI, so checking these ones in addition should allow us to fill in more homepage data.

### DIFF
--- a/app/models/package_manager/pypi/json_api_project.rb
+++ b/app/models/package_manager/pypi/json_api_project.rb
@@ -68,7 +68,10 @@ module PackageManager
 
       def homepage_url
         @data.dig("info", "home_page").presence ||
-          @data.dig("info", "project_urls", "Homepage")
+          @data.dig("info", "project_urls", "Homepage") ||
+          @data.dig("info", "project_urls", "Home") ||
+          @data.dig("info", "project_urls", "homepage") ||
+          @data.dig("info", "project_urls", "HomePage")
       end
 
       def preferred_repository_url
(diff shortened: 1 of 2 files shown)


macowie (approved):

jhan217 (approved):

app/models/package_manager/pypi/json_api_project.rb
```diff
@@ -68,7 +68,10 @@ def repository_url
 
       def homepage_url
         @data.dig("info", "home_page").presence ||
```
jhan217, line 70: 🎱 Is it worth logging if we can't find a homepage so we can go back and see if there are additional variations of "Homepage" we can pull from?

PR #3175, opened 2023-07-28
Adds "lifted" badge to project pages, and add rake task to resync projects' lifted status

tiegz:
* adds `Project#lifted` boolean column
* adds `Project.fetch_lifted_projects` to get a list of all the lifted projects
* adds `projects:update_lifting_statuses` rake task to resync that list ^ with the Libraries db
* adds "this is lifted" badge/copy to project pages that are lifted

### Example non-lifted

Screenshot 2023-07-28 at 4 02 36 PM

### Example lifted
Screenshot 2023-07-28 at 4 00 50 PM


papiro (approved):

app/controllers/api/projects_controller.rb
```diff
@@ -2,11 +2,25 @@
...
+  ALLOWED_UPDATE_API_KEY_IDS = [1].freeze
 
   def show
     render(json: @project, show_updated_at: internal_api_key?)
   end
 
+  def update
+    # currently this endpoint only updates the "lifted" attribute. No need for callbacks/validations yet.
+    project_params = params.require(:project).permit(:lifted)
+    @project.update_columns(lifted: project_params[:lifted])
```
tiegz, line 17: `update_columns` still uses AR's builtin type-casting here, so anything that looks like a "false" value will get saved as false (or true, otherwise).

db/migrate/20230728192141_add_lifted_to_project.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddLiftedToProject < ActiveRecord::Migration[5.2]
+  def change
+    add_column :projects, :lifted, :boolean, default: false
```
mikeyoung85, line 5: Strong migrations recommends this be a two step operation: https://github.com/ankane/strong_migrations#adding-a-column-with-a-default-value

Although I'm not sure if you can have a boolean column without it defaulting to something...

```
class AddLiftedToProject < ActiveRecord::Migration[5.2]
  def up
    add_column :projects, :lifted, :boolean
    change_column_default :projects, :lifted, false
  end

  def down
    remove_column :projects, :lifted
  end
end
```

mikeyoung85 (approved):

db/migrate/20230728192141_add_lifted_to_project.rb
tiegz: thanks, just pushed up a fix

db/migrate/20230731220834_backfill_add_lifted_to_project.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+class BackfillAddLiftedToProject < ActiveRecord::Migration[5.2]
+  disable_ddl_transaction!
+
+  def up
+    Project.in_batches do |relation|
+      relation.update_all lifted: false
+      sleep(0.01)
+    end
```
tiegz, line 10: should take 1-2 minutes, by my estimate

config/application.rb
```diff
@@ -82,5 +82,6 @@ class Application < Rails::Application
     config.bitbucket_secret = ENV["BITBUCKET_SECRET"]
     config.ga_analytics_id = ENV["GA_ANALYTICS_ID"]
     config.gtm_id = ENV["GTM_ID"]
+    config.tidelift_api_key = ENV["TIDELIFT_API_KEY"]
```
tiegz, line 85: I'm setting to a new org key called "Libraries Lifting API Key" in "tidelift-internal" (which has beta API access)

app/models/project.rb
```diff
@@ -200,8 +201,23 @@ def self.total
...
-  def test
-    puts "hello"
+  def self.fetch_lifted_projects
+    raise "Tidelift API key not set." if Rails.configuration.tidelift_api_key.blank?
+
+    current_page = 1
+    results = []
+    while !current_page.nil? && current_page < 1000 # failsafe
+      resp = Typhoeus.get("https://api.tidelift.com/external-api/packages?lifted=true&page=#{current_page}&per_page=100", headers: { "Authorization" => "Bearer #{Rails.configuration.tidelift_api_key}" })
+      break if resp.response_code != 200
```
mikeyoung85, line 211: This break is going to cut off the results if any of the page requests fail. The rake task calling this is marking any currently lifted Projects that are missing from the list as unlifted. This break is going to cause some projects to be labeled as unlifted until the rake task runs again and is able to get through all pages.

app/models/project.rb
```diff
@@ -200,8 +201,23 @@ def self.total
...
+
+      json = JSON.parse(resp.body)
+      current_page = json["next_page"]
+      results += json["results"]
+    end
```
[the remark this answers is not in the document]
tiegz: agreed, this doesn't seem like the right place for it. I think I can use the `app/services` folder to put something together

app/services/fetch_lifted_projects.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+# Fetches a list of all Projects that are lifted, via the Tidelift API.
+class FetchLiftedProjects
```
tiegz, line 4: cc @mpace965

app/services/fetch_lifted_projects.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  def run
+    raise MissingApiKey if Rails.configuration.tidelift_api_key.blank?
+
+    current_page = 1
+    results = []
+    while !current_page.nil? && current_page < 1000 # failsafe
+      resp = Typhoeus.get("https://api.tidelift.com/external-api/packages?lifted=true&page=#{current_page}&per_page=100", headers: { "Authorization" => "Bearer #{Rails.configuration.tidelift_api_key}" })
+
+      raise BadResponse.new(resp.response_code, current_page) if resp.response_code != 200
```
tiegz, line 21: @mikeyoung85 WDYT about this? It'll fail entirely if it gets a bad response, which will halt the rake task. If this happens then we can investigate, instead of accidentally marking Projects as "unlifted".

mikeyoung85: That works for me

app/services/fetch_lifted_projects.rb
```diff
@@ -0,0 +1,32 @@
...
+
+      json = JSON.parse(resp.body)
+      current_page = json["next_page"]
+      results += json["results"]
+    end
+
+    results
+      .map { |result| Project.find_best(result["platform"], result["name"]) }
+      .compact
```
mikeyoung85, line 30: WDYT of passing back the IDs here instead of all the Projects found from the response? I'm wondering if we are going to run into memory issues holding onto every `Project` object when the rake task only cares about the IDs.

tiegz: yeah, I bet memory-wise it'll be fine, but can't hurt to be more efficient anyways 👍🏻

lib/tasks/projects.rake
```diff
@@ -10,6 +10,27 @@ namespace :projects do
     Project.not_removed.where(last_synced_at: nil).order("updated_at ASC").limit(500).each(&:async_sync)
   end
 
+  desc "Update lifting statuses of projects"
+  task update_lifting_statuses: :environment do
+    exit if ENV["READ_ONLY"].present?
+
+    lifted_project_ids = Project.fetch_lifted_projects.map(&:id)
```
mpace965, line 17:
```suggestion
    FetchLiftedProjects.new.run.map(&:id)
```

or without the map if you switch the impl to return ids

PR #3178, opened 2023-08-01
Allow pypi project versions to change statuses

jhan217:
This PR allows changes to the status field on existing Pypi project version.

For Pypi projects, the existing code short circuits updating the version if Libraries already knows about the version. I suspect this is due to not wanting to make multiple external API calls per project version on lookup of a project's versions, especially for the individual version's license & published_ats. This makes sense, as it would get extremely expensive for projects that have tens of thousands of releases, and we don't expect versions to change that much after they get published (e.g. once a version is published, we don't expect its published date to change). Doing this set of lookups/updates only when we see a new version that we haven't seen before mitigates that concern.

However, this existing short circuit prevents us from updating project versions if something changes about it after publication -- such as when a specific version gets yanked. This PR changes that behavior slightly to allow updates to the status if we see that the version has been yanked, which we get from the main project API call. If the version is marked with `yanked: true`, we will change the version.status to 'Removed'; if it is not, then we either keep the existing status or set the status to nil (the "unyanking" case). Setting the version.status is done in the Pypi specific version processor, and the update happens in the `add_version` call, which is more of an upsert at this point. This means the `deprecate_version` method is effectively not doing anything for Pypi projects, but I've left a note there that it may need to get updated to support the "libraries knows about the version, but pypi doesn't (probably because it was hard deleted on the package manager side)" scenario.

### DIFF
--- a/app/models/package_manager/base.rb
+++ b/app/models/package_manager/base.rb
@@ -177,16 +177,8 @@ module PackageManager
           .where.not(number: version_hashes.pluck(:number))
           .where("status != ? or status is null", removed_status)
           .update_all(status: removed_status, updated_at: current_time)
-      when "pypi"
-        removed_versions = version_hashes.filter { |x| x[:yanked] == true }
-
-        if removed_versions.any?
-          db_project
-            .versions
-            .where(number: removed_versions.pluck(:number).compact)
-            .where("status != ? or status is null", removed_status)
-            .update_all(status: removed_status, updated_at: current_time)
-        end
+        # yanked pypi versions are marked as such in the api, and the majority of them are handled upstream of here
+        # TODO: if libraries knows about the version but pypi does not, also mark the version differences as deprecated
       end
     end
 
--- a/app/models/package_manager/base/api_version_to_upsert.rb
+++ b/app/models/package_manager/base/api_version_to_upsert.rb
@@ -22,14 +22,17 @@ module PackageManager
       end
 
       def to_version_model_attributes
-        {
+        non_nillable_attributes = {
           number: @version_number,
           published_at: @published_at,
           runtime_dependencies_count: @runtime_dependencies_count,
           original_license: @original_license,
           repository_sources: @repository_sources,
-          status: @status,
         }.compact
+
+        # If the status is currently "Removed" and a change comes in to update that status to nil/"Active",
+        # Update the version's status
+        non_nillable_attributes.merge({ status: @status })
       end
     end
   end
--- a/app/models/package_manager/pypi/version_processor.rb
+++ b/app/models/package_manager/pypi/version_processor.rb
@@ -11,20 +11,31 @@ module PackageManager
 
       def execute
         @project_releases.map do |project_release|
-          next @known_versions[project_release.version_number] if @known_versions.key?(project_release.version_number)
+          if @known_versions.key?(project_release.version_number)
+            known_attributes = @known_versions[project_release.version_number]
+            status = if project_release&.yanked?
+                       "Removed"
+                     elsif known_attributes[:status] == "Removed"
+                       nil
+                     else
+                       known_attributes[:status]
+                     end
+            known_attributes[:status] = status
+            known_attributes
+          else
+            version_number = project_release.version_number
+            original_license = json_api_single_release_for_version(version_number).license
+            rss_api_release = rss_api_releases_hash[version_number]
+            published_at = project_release.published_at || rss_api_release&.published_at
+            status = project_release&.yanked? ? "Removed" : nil
 
-          version_number = project_release.version_number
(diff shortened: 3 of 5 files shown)


app/models/package_manager/base/api_version_to_upsert.rb
```diff
@@ -22,14 +22,17 @@ def initialize(
...
+        non_nillable_attributes = {
           number: @version_number,
           published_at: @published_at,
           runtime_dependencies_count: @runtime_dependencies_count,
           original_license: @original_license,
           repository_sources: @repository_sources,
-          status: @status,
         }.compact
+
+        # If the status is currently "Removed" and a change comes in to update that status to nil/"Active",
```
jhan217, line 33: I'm not sure if this is something that needs to take the PackageManager into consideration, since it could potentially be unintentionally setting a bunch of statuses to nil if the other PackageManagers are relying on this cleanup to avoid updating stuff that don't need to be updated, vs passing in exactly what needs to be updated. But at least from `Base#update`, if the status gets set to nil here, it'll get set back to 'Removed' right after this in the `deprecate_version` call.

This adds another reason to change the idea of "version is active" from status: nil to status: 'active' -- we can't tell if the nil is because we haven't looked at the version status so it's 'active' by default, or if it's because it's actually active based off of some other piece of information we get about the version, or if it's a bug somewhere upstream. But that will be a bigger change that touches all the package managers, and out of scope for this PR.

app/models/package_manager/pypi/version_processor.rb
```diff
@@ -11,20 +11,31 @@ def initialize(project_releases:, project_name:, known_versions:)
...
-          published_at = project_release.published_at || rss_api_release&.published_at
-          yanked = project_release&.yanked? || false
-
-          {
-            number: version_number,
-            published_at: published_at,
-            original_license: original_license,
-            yanked: yanked,
-          }
+          if @known_versions.key?(project_release.version_number)
```
jhan217, line 14: This is basically turning the postfix if into a full if/else statement, with the case if we already know the version only potentially changing the status field.

app/models/package_manager/pypi/version_processor.rb
```diff
@@ -11,20 +11,31 @@ def initialize(project_releases:, project_name:, known_versions:)
...
+            known_attributes = @known_versions[project_release.version_number]
+            status = if project_release&.yanked?
+                       "Removed"
+                     elsif known_attributes[:status] == "Removed"
+                       nil
+                     else
+                       known_attributes[:status]
```
jhan217, line 21: This is to set the status to Removed if we get a positive indication from pypi that the status is removed, unyank the version if needed, and to not unset the existing status in the upsert if it isn't yanked.

app/models/package_manager/base.rb
```diff
@@ -177,16 +177,8 @@ def self.deprecate_versions(db_project, version_hashes)
...
-        removed_versions = version_hashes.filter { |x| x[:yanked] == true }
-
-        if removed_versions.any?
-          db_project
-            .versions
-            .where(number: removed_versions.pluck(:number).compact)
-            .where("status != ? or status is null", removed_status)
-            .update_all(status: removed_status, updated_at: current_time)
-        end
+        # yanked pypi versions are marked as such in the api, and the majority of them are handled upstream of here
```
jhan217, line 180: This change basically takes advantage of the fact that the `add_version` call on line 133 does both an add (for versions we haven't seen before) or an update (for versions that we know about already). The latter means all the data that gets passed in needs to be current state of the version row, so the status change can be passed in here.

app/models/package_manager/pypi/version_processor.rb
```diff
@@ -11,20 +11,31 @@ def initialize(project_releases:, project_name:, known_versions:)
 
       def execute
         @project_releases.map do |project_release|
```
jhan217, line 13: The case we're missing here is if libraries knows about a version but the api doesn't (i.e. the version number is not in `@project_releases`)

app/models/package_manager/base/api_version_to_upsert.rb
macowie: Agreed on the nil status problem. I think this change is reasonably safe because the pre-refactor version of `add_version` also always overwrote the status field

```
      version_hash = version_hash.symbolize_keys.slice(
        :number,
        :published_at,
        :runtime_dependencies_count,
        :original_license,
        :repository_sources,
        :status
      )

      existing = db_project.versions.find_or_initialize_by(number: version_hash[:number])
      existing.skip_save_project = true
      existing.assign_attributes(version_hash)
```

macowie (approved):

mikeyoung85 (approved):

johnbintz-tidelift (approved):

PR #3184, opened 2023-08-10
update banner message

bee64:
update banner message to be one libraries message instead of two upstream messages

Screenshot 2023-08-10 at 3 07 48 PM


kjmahalingam (approved):
🎉

PR #3186, opened 2023-08-18
chore: Upgrade Libraries to Rails 6.0

macowie:
The goal of this PR is to take the first step towards getting the application updated to a supported version of Rails. To the degree that it is doable (and/or convenient), hold off on configuration changes and other updates until after we get there. 

The elasticsearch gem versions we were using are still compatible, so I didn't touch them. I had a look at the code that interacts with them and don't see anything that should be affected by the upgrade. My local indices continued to work and in all the tests and my spot checking seems to be fine. As long as we have backups of production's indices, I don't see much risk here.

I did a fair amount of manual testing to look for any unexpected bugs and make sure the user facing things were working and one template glitch aside, it seemed good to me. Others are more than welcome to bang on it, though.


app/views/shared/_flash.html.erb
```diff
@@ -51,6 +51,6 @@
     class="flash-banner text-center hidden-xs"
     role="alert"
   >
-    <%= render :file => tidelift_flash_partial %>
+    <%== render :file => tidelift_flash_partial %>
```
tiegz, line 54: was this one a typo/debugging change?

pranav (approved):

config/initializers/active_record.rb
```diff
@@ -11,7 +11,8 @@ def with_statement_timeout(seconds)
       execute("SET statement_timeout = '#{seconds * 1000}';")
       yield(self)
     ensure
-      execute("SET statement_timeout = '#{ActiveRecord::Base.configurations.dig(Rails.env, 'variables', 'statement_timeout')}';")
+      default_timeout = ActiveRecord::Base.configurations.find_db_config(Rails.env).config.dig("variables", "statement_timeout")
```
tiegz, line 14: looks good, but just a headsup that `config` will be gone in Rails 7.0:

`DEPRECATION WARNING: DatabaseConfig#config will be removed in 7.0.0 in favor of DatabaseConfig#configuration_hash which returns a hash with symbol keys (called from <main> at bin/rails:6)`

tiegz (approved):

app/views/shared/_flash.html.erb
macowie: I couldn't track this back to a change in Rails, so this could be not the quite right fix, but after the upgrade I was getting the escaped html from this using `render :file`.  AFAICT this was all still correct usage

tiegz: interesting! The erubi docs confirm that it's the escaped version of `<%=`. TIL!

PR #3187, opened 2023-08-28
chore: [sc-36386] Upgrade Libraries to Rails 6.1

macowie:
[5.2 to 6.0 Upgrade PR](https://github.com/librariesio/libraries.io/pull/3186)
[Rails Upgrade Guide](https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-6-0-to-rails-6-1)

Like the prior, this PR is focused on getting the framework updated without too many unnecessary changes while taking care to handle actual deprecation warnings and such. This is a smaller jump than last time, and AFAICT that's been a smooth rollout, so 🤞


config/environments/development.rb
```diff
@@ -53,13 +61,18 @@
...
 
   # Use an evented file watcher to asynchronously detect changes in source code,
   # routes, locales, etc. This feature depends on the listen gem.
   # config.file_watcher = ActiveSupport::EventedFileUpdateChecker
 
-  config.host = "localhost"
+  # Uncomment if you wish to allow Action Cable access from any origin.
+  # config.action_cable.disable_request_forgery_protection = true
 
+  config.host = "localhost"
```
macowie, line 76: Tried to get cleaner diffs when updating and move options that weren't in the default config to the bottom

config/routes.rb
```diff
@@ -229,5 +229,5 @@
   get "/:platform/:name.about", to: "projects#about", as: :about_project, constraints: { name: /.*/ }
   get "/:platform/:name.ABOUT", to: "projects#about", constraints: { name: /.*/ }
   get "/:platform/:name", to: "projects#show", as: :project, constraints: { name: /.*/ }, defaults: { format: "html" }
-  get "/:id", to: "platforms#show", as: :platform
+  get "/:id", to: "platforms#show", as: :platform, constraints: { format: "html" }
```
macowie, line 232: Not strictly upgrade related, but something that emerged from the last upgrade. Because this is a root level catch-all route, we end up running a lot of strange/spam/probe requests through this action. 
We don't render any other formats on this action, and if it's supposed to support platform names with a dot in them, it doesn't currently anyway. So this should just be turning CSRF/UnknownFormat errors into 404s as they should be.

tiegz (approved):
WFM locally!

PR #3203, opened 2023-09-05
chore: Update omniauth

macowie:
We had been on an old version of Omniauth with some outstanding alerts, along with some aged plugins. This updates everything to latest, and switches to a [different stale bitbucket integration](https://github.com/fnando/omniauth-atlassian-bitbucket/blob/main/lib/omniauth/strategies/bitbucket.rb), but one that actually uses oauth2 now.

This does introduce a new (silenced) warning about allowing oauth requests by GET, and [this is in fact insecure](https://github.com/omniauth/omniauth/wiki/Resolving-CVE-2015-9284). But it's also something the app has been doing the whole time. So, while it's something to resolve, it shouldn't block other fixes.


macowie:
Filed a followup story to update this insecure setting, it's a little involved and I'd rather at least get the patches out for now

pranav (approved):

PR #3204, opened 2023-09-06
Add rake task to retrieve maven versions from ignored sources

kjmahalingam:


lib/tasks/one_off.rake
```diff
@@ -257,4 +257,43 @@ namespace :one_off do
...
     puts "Versions corrected: #{corrected_count}"
   end
+
+  desc "Retrieve Maven versions from ignored sources"
+  task :retrieve_maven_versions_from_ignored_sources, %i[output_file] => :environment do |_t, args|
+    maven_versions = Version.where(project_id: Project.where(platform: "Maven"))
+    # We want to retrieve the versions where the repository_sources are not solely ["Maven"] or ["Google"]
+    ignored_maven_versions = maven_versions.where.not("repository_sources = ?", ["Maven"].to_json).where.not("repository_sources = ?", ["Google"].to_json)
+    # We also want to retrieve the versions where repository_sources is nil, which aren't caught by the above
+    no_source_maven_versions = maven_versions.where("repository_sources IS NULL")
```
kjmahalingam, line 267: I'm open to condensing these, although I had some errors with nil-checking as I constructed some queries with `.or`s. Since we only intend for this to be a one-off I wasn't super concerned with it and left them separate. Also allows us to log info about these different cases separately.

kjmahalingam (commented):
After this is run and we get the export, I intend to delete this task and spec, as they will no longer be needed.

lib/tasks/one_off.rake
```diff
@@ -257,4 +257,43 @@ namespace :one_off do
...
+
+    puts "Count of ignored-source Maven versions to retrieve: #{ignored_maven_versions.count}"
+    puts "Count of no-source Maven versions to retrieve: #{no_source_maven_versions.count}"
+
+    output_file = args[:output_file] || File.join(__dir__, "output", "ignored_maven_versions.csv")
```
johnbintz-tidelift, line 272: Just to be safe, making sure the target directory exists would be good:

```suggestion
    output_file = args[:output_file] || File.join(__dir__, "output", "ignored_maven_versions.csv")
    FileUtils.mkdir_p File.dirname(output_file)
```

lib/tasks/one_off.rake
```diff
@@ -257,4 +257,43 @@ namespace :one_off do
...
+    puts "Outputting data to #{output_file}"
+
+    output_csv = CSV.new(File.open(output_file, "w"))
+    output_csv << %w[platform name version]
+
+    puts "Outputting data for ignored-source Maven versions"
+    ignored_maven_versions.in_batches do |batch|
+      puts "Processing batch"
+      batch.each do |version|
+        output_csv << [version.project.platform, version.project.name, version.number]
```
johnbintz-tidelift, line 283: Can you also include what the `repository_sources` for this version are, so we're able to double-check things using the CSV file?

lib/tasks/one_off.rake
```diff
@@ -257,4 +257,43 @@ namespace :one_off do
     puts "Total packages checked: #{input_tsv_file.count}"
     puts "Versions corrected: #{corrected_count}"
   end
+
+  desc "Retrieve Maven versions from ignored sources"
+  task :retrieve_maven_versions_from_ignored_sources, %i[output_file] => :environment do |_t, args|
+    maven_versions = Version.where(project_id: Project.where(platform: "Maven"))
+    # We want to retrieve the versions where the repository_sources are not solely ["Maven"] or ["Google"]
+    ignored_maven_versions = maven_versions.where.not("repository_sources = ?", ["Maven"].to_json).where.not("repository_sources = ?", ["Google"].to_json)
```
johnbintz-tidelift, line 265: Is this using an `OR` in the query? If not, does this produce an `OR`?

```suggestion
    ignored_maven_versions = maven_versions.where.not("repository_sources = ?", ["Maven"].to_json).or(maven_versions.where.not("repository_sources = ?", ["Google"].to_json))
```

lib/tasks/one_off.rake
johnbintz-tidelift: I say leave them uncondensed.

lib/tasks/one_off.rake
kjmahalingam: We discussed this and determined that changing this query changes the results to something we do not desire, so leaving it as-is.

johnbintz-tidelift (approved):

PR #3206, opened 2023-09-14
feat: [sc-31317] [libraries] Update NuGet package manager to use canonical names

macowie:
Story details: https://app.shortcut.com/tidelift/story/31317


app/domain/structured_log.rb
```diff
@@ -0,0 +1,71 @@
+# frozen_string_literal: true
+
+module StructuredLog
+  # See https://github.com/tidelift/design-docs/blob/main/2022-02-backend-logging-conventions.md for conventions
```
jhan217, line 4: Not sure if we should be linking to private repos from public repos.

app/domain/structured_log.rb
```diff
@@ -0,0 +1,71 @@
...
+  def self.capture(name, data_hash)
+    rails_logger(name, data_hash)
+  rescue ArgumentError => e
+    Rails.logger.error(e.message)
+    raise e
+  rescue StandardError => e
+    Rails.logger.error("Error capturing structured log for metric=#{name} - #{e.message}")
+  end
+
+  def self.rails_logger(name, data_hash)
```
jhan217, line 39: Perhaps something like `def self.log_info` or so?  To me at least, `rails_logger` implies a logging object is the output of this method. Including `info` in there somewhere can make it more descriptive & differentiate it from warning/error logs as well.

app/models/package_manager/nu_get.rb
```diff
@@ -226,5 +232,19 @@ def self.parse_requirements(range)
...
+      nuget_href = "https://www.nuget.org/packages/#{name}"
+      page = get_html(nuget_href)
+      og_url_element = page.css("meta[property='og:url']").first
+
+      unless og_url_element
+        StructuredLog.capture("FETCH_CANONICAL_NAME_FAILED", { platform: "nuget", name: name })
+        return false
+      end
+
+      og_url_element.attributes.fetch("content").text.sub("https://nuget.org/packages/", "").sub(/\/$/, "")
```
jhan217, line 247: What happens if the upstream changes where they put this field / html structure?

spec/models/package_manager/nu_get_spec.rb
```diff
@@ -221,4 +222,91 @@ def stub_releases(releases_data = {})
       end
     end
   end
+
+  describe "::fetch_canonical_nuget_name" do
+    subject(:result) do
+      VCR.use_cassette(cassette) { described_class.fetch_canonical_nuget_name(name) }
```
jhan217, line 228: Would it be possible to do an example without a cassette, such as using a formatted string/snippet with the minimum amount of data needed? That way, it'll be easier for the next person to quickly identify what it is to look for, vs looking through the code + cassette to figure out what the expected structure is.

app/models/package_manager/nu_get.rb
```diff
@@ -73,11 +73,17 @@ def self.project_names
     end
 
     def self.project(name)
+      canonical_name = fetch_canonical_nuget_name(name)
+
+      if name != canonical_name
```
jhan217, line 78: Is it expected for the input name to always be the canonical name?

jhan217:
For the future - is it necessary for the logging and code functionality change to go out in the same PR?

app/models/package_manager/nu_get.rb
macowie: I want it to fail loudly or sound an alert when there's a spike in fetch errors. I can make the failure more explicit here though

jhan217: HTML parsing is pretty fraught with errors. The only bad case I think we can autodetect is if we somehow manage to extract some sort of empty string / nil without erroring, which I think we don't ever want to map to anything, so failing loudly seems like a good idea.

jhan217:
Posthumous 👍

PR #3209, opened 2023-09-21
SearchController#index: short-circuit when no search term is provided

mpace965:
This prevents expensive elasticsearch queries that have been the target of scraping efforts.

Screenshot 2023-09-22 at 11 35 42 AM


katzj (approved):

loganlifts:
No concerns with shipping this approach, but I'm curious, should we do anything at all if a search term isn't provided into the input? It seems reasonable to me to just not allow the search button / form submission to work.

PR #3214, opened 2023-09-29
fix: [sc-37352] Packages are incorrectly being marked as deprecated

johnbintz-tidelift:
This implements for NuGet the same class structure approach we used in PyPI.
This made it easier to find and fix, and write tests for, an issue with
package deprecation where the incorrect release was being used. As with
the PyPI changes, this removes more uses of raw Hashes for data storage
and transfer around the app.


app/models/package_manager/nu_get.rb
```diff
@@ -28,14 +28,15 @@ def self.install_instructions(db_project, version = nil)
     end
 
     def self.deprecation_info(db_project)
-      info = latest_remote_version(db_project.name)
-      deprecation = info&.dig("items")&.first&.dig("items")&.first&.dig("catalogEntry", "deprecation")
```
johnbintz-tidelift, line 32: This is where the wrong release was being used. We can't guarantee the order by which NuGet returns the releases, so we can't assume the first one found is both the latest, and the correct place to get deprecation info.

With the changes in this PR:

* we now sort all items by published date
* we use the most recent release to determine if the entire package is deprecated, which is how we handle some other ecosystems

This can likely become more precise as we run into more issues, but for now it fixes the immediate issue, and makes it easier to make changes later on if needed.

macowie: Might be missing it, but where do we ensure the releases are sorted for the latest to be the last item? I see the rocketship on releases to sort by published_at but not where we do the sorting?

johnbintz-tidelift:
Gonna have some changes coming to this shortly, hold off on reviewing any more until I push another commit.

app/models/package_manager/nu_get.rb
johnbintz-tidelift: I forgot to add the sort and a test for it. They've been added now.

macowie (approved):
Looks great!

PR #3219, opened 2023-10-05
Limit Number of Times We Try to Find a Working Auth Token

mikeyoung85:


app/models/auth_token.rb
```diff
@@ -82,12 +82,14 @@ def headers(_context)
     GraphQL::Client.new(schema: Rails.application.config.graphql.schema, execute: http_adapter)
   end
 
-  def self.find_token(api_version)
-    return @auth_token if @auth_token&.high_rate_limit?(api_version)
+  def self.find_token(api_version, retries: 0)
+    auth_token = authorized.order(Arel.sql("RANDOM()")).limit(1).first
```
tiegz, line 86: any reason to keep the `limit(1)` alongside the `first`?

tiegz (approved):

app/models/auth_token.rb
mikeyoung85: It's a bad habit at this point. I keep forgetting that rails magically figures out this query when using `.first`. I'll make the update here.

PR #3221, opened 2023-10-06
[SC-34666] Restructure and rename version removal code in PackageManager::Base

johnbintz-tidelift:
The goals of this PR are to:

* remove the switch on an object property within a method to determine behavior (`Base.deprecate_versions`) and instead let the package manager classes inject the necessary behavior into the release removal action.
* use the correct terminology. We aren't deprecating these releases -- they are Removed as the package managers in question no longer acknowledge them. Deprecation of releases/projects is handled in a separate codepath.
* reduce the number of untyped Hashes that are being passed around. The `ApiVersionToUpsert` class can be turned into `ApiVersion` and reused for release removal actions.
* move release removal behavior to a single executor class, rather than burying it in a static method on an inherited class. This matches how we are adding new versions to Projects.


spec/models/package_manager/base/missing_version_remover_spec.rb
```diff
@@ -0,0 +1,109 @@
...
+    let(:version_already_removed_status) { "Removed" }
+    let!(:version_already_removed) { db_project.versions.create(number: "3.0.0", status: version_already_removed_status, updated_at: original_time) }
+
+    let(:version_in_another_state_status) { "Deprecated" }
+    let!(:version_in_another_state) { db_project.versions.create(number: "4.0.0", status: version_in_another_state_status, updated_at: original_time) }
+
+    let(:original_time) { Time.zone.parse("2010-01-01 10:00:00") }
+    let(:removal_time) { Time.zone.parse("2020-01-01 10:00:00") }
+
+    let(:versions_to_remove) { ["1.0.0"] }
```
macowie, line 25: Should these be `:versions_to_keep`?

macowie (approved):

PR #3223, opened 2023-10-06
Enable removed release support for Maven Central

johnbintz-tidelift:
Goal:

* Handle removed releases from Maven Central, using a real package with removed releases as an example.


macowie (approved):
🚀

macowie (approved):

PR #3224, opened 2023-10-10
feat: [sc-31318] NuGet project verification worker: fix queueing adjust and rate of work

macowie:
Story details: https://app.shortcut.com/tidelift/story/31318

Fix spacing out the batches, and slow the rate some more to be safe & considerate.

### DIFF
--- a/lib/tasks/projects.rake
+++ b/lib/tasks/projects.rake
@@ -297,20 +297,28 @@ namespace :projects do
 
   desc "Verify NuGet Projects"
   task :verify_nuget_projects, [:count] => :environment do |_task, args|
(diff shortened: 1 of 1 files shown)


jhan217 (approved):
Would it be worth putting in something to let the user pick up where things started breaking partway through a run instead of going through the entire dataset each time?

macowie:
> Would it be worth putting in something to let the user pick up where things started breaking partway through a run instead of going through the entire dataset each time?

Wasn't sure how best to go about this, but this seems to work and could be a good pattern going forward.

jhan217 (approved):

PR #3227, opened 2023-10-17
fix: [sc-37574] Add task to scan and clean out expired authtokens

macowie:
Story details: https://app.shortcut.com/tidelift/story/37574

First part of cleaning AuthTokens up. Toss out the expired ones so we can get to the good ones.

I wrote this as a one-off, but it's probably worth scheduling too, monthly maybe.

### DIFF
--- a/lib/tasks/auth_tokens.rake
+++ b/lib/tasks/auth_tokens.rake
@@ -1,0 +1,33 @@
+# frozen_string_literal: true
+
+namespace :auth_tokens do
+  desc "Verify each authorized AuthToken that it's still authorized"
+  task :reverify_authorized, %i[count start] => :environment do |_task, args|
+    exit if ENV["READ_ONLY"].present?
+    args.with_defaults(count: 500, start: nil)
+    last_id = "none"
+
+    begin
+      AuthToken
+        .authorized
+        .in_batches(of: args.count, start: args.start)
+        .each do |token_batch|
+          token_batch.each do |token|
+            token.update_attributes(
+              login: token.github_client.user[:login],
+              authorized: token.still_authorized?
+            )
+
+            last_id = token.id
+          end
+
+          # Don't necessarily have to worry about rate limit #still_authorized? /should/ still work
+          # because a rate limit means we auth'ed successfully. But we can still pace ourselves.
+          sleep 1
+        end
+    rescue StandardError, Interrupt => e
+      puts "\n\n### Last AuthToken id processed: #{last_id} \n\n\n"
+      raise e
+    end
+  end
+end
--- a/lib/tasks/auth_tokens.rake
+++ b/lib/tasks/auth_tokens.rake
@@ -14,15 +14,17 @@ namespace :auth_tokens do
     begin
       AuthToken
         .authorized
-        .find_in_batches(of: args.batch_size, start: args.start).each |token_batch|
+        .find_in_batches(batch_size: args.batch_size, start: args.start).each do |token_batch|
           token_batch.each do |token|
             result = token.still_authorized?
-            fresh_login = token.github_client.user[:login] if result
 
-            token.update(
-              login: (fresh_login || token.login),
-              authorized: result
-            )
+            if result == true
+              token.login = token.github_client.user[:login]
+            else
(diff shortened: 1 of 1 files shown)


lib/tasks/auth_tokens.rake
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+namespace :auth_tokens do
+  desc "Verify each authorized AuthToken that it's still authorized"
+  task :reverify_authorized, %i[count start] => :environment do |_task, args|
+    exit if ENV["READ_ONLY"].present?
```
johnbintz-tidelift, line 6: When is `READ_ONLY` set?

lib/tasks/auth_tokens.rake
```diff
@@ -0,0 +1,47 @@
...
+    args.with_defaults(count: 500, start: nil)
+    last_id = "none"
+
+    begin
+      AuthToken
+        .authorized
```
johnbintz-tidelift, line 12: WDYT about sorting this list in either one of two ways:

* alphabetically by `id` so that, if we have to reprocess the list, the `last_id` in the error message will allow us to restart from there if desired
* oldest-to-newest by `updated_at` so that the oldest `AuthTokens` are processed first

lib/tasks/auth_tokens.rake
```diff
@@ -0,0 +1,47 @@
...
+
+            last_id = token.id
+          end
+
+          # Don't necessarily have to worry about rate limit #still_authorized? /should/ still work
+          # because a rate limit means we auth'ed successfully. But we can still pace ourselves.
+          sleep 1
+        end
+    rescue StandardError, Interrupt => e
+      puts "\n\n### Last AuthToken id processed: #{last_id} \n\n\n"
```
johnbintz-tidelift, line 43: Would it be possible to also print the number of auth tokens processed so far so that the `start` arg can be used?

lib/tasks/auth_tokens.rake
johnbintz-tidelift: Sounds good to me!

lib/tasks/auth_tokens.rake
macowie: This is a good question! The application seems to (or did) support some kind of [read only mode](https://github.com/librariesio/libraries.io/blob/268d5397c91d8c598b2af9fad0d9dd972bb4ed64/app/controllers/application_controller.rb#L281-L283) but it does not seem to be documented, nor referenced in any of our infra. Happy to stop cargo culting this line if we don't know we need it

lib/tasks/auth_tokens.rake
```diff
@@ -0,0 +1,47 @@
...
+    last_id = "none"
+
+    begin
+      AuthToken
+        .authorized
+        .in_batches(of: args.count, start: args.start)
+        .each do |token_batch|
+          token_batch.each do |token|
+            result = token.still_authorized?
+            fresh_login = token.github_client.user[:login]
```
mikeyoung85, line 17: This is going to throw some errors if the token is not authorized anymore. You can see some of the rescues happening in the [still_authorized?](https://github.com/librariesio/libraries.io/blob/bd03a18a2fd6a11292d0096a47613ea362e0c19f/app/models/auth_token.rb#L46) method. If we want the login, we should only fetch it if `result == true`.

lib/tasks/auth_tokens.rake
```diff
@@ -0,0 +1,47 @@
...
+    exit if ENV["READ_ONLY"].present?
+    args.with_defaults(count: 500, start: nil)
+    last_id = "none"
+
+    begin
+      AuthToken
+        .authorized
+        .in_batches(of: args.count, start: args.start)
+        .each do |token_batch|
+          token_batch.each do |token|
```
mikeyoung85, line 15: You could get rid of the double `.each` by using [find_each](https://apidock.com/rails/ActiveRecord/Batches/find_each) instead of `in_batches`. It should accept the same parameters and is basically syntactic sugar for what you have here.

mikeyoung85 (approved):

PR #3228, opened 2023-10-18
fix: [sc-37574] Resolve Libraries to Github API connection errors

macowie:
Story details: https://app.shortcut.com/tidelift/story/37574

It turns out we are not passing the token anymore, and they'd really like to see one.

### DIFF
--- a/app/models/auth_token.rb
+++ b/app/models/auth_token.rb
@@ -67,13 +67,10 @@ class AuthToken < ApplicationRecord
   def self.new_v4_client(token)
     token ||= AuthToken.token
     http_adapter = GraphQL::Client::HTTP.new("https://api.github.com/graphql") do
-      @token = token
-
+      @@token = token # rubocop: disable Style/ClassVars
       # rubocop: disable Lint/NestedMethodDefinition:
       def headers(_context)
-        {
-          "Authorization" => "Bearer #{@token}",
-        }
+        { "Authorization" => "bearer #{@@token}" }
       end
       # rubocop: enable Lint/NestedMethodDefinition:
     end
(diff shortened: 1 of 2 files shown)


app/models/auth_token.rb
```diff
@@ -67,13 +67,10 @@ def self.new_client(token, options = {})
...
     http_adapter = GraphQL::Client::HTTP.new("https://api.github.com/graphql") do
-      @token = token
-
+      @@token = token # rubocop: disable Style/ClassVars
       # rubocop: disable Lint/NestedMethodDefinition:
       def headers(_context)
-        {
-          "Authorization" => "Bearer #{@token}",
-        }
+        { "Authorization" => "bearer #{@@token}" }
```
mikeyoung85, line 73: Since this is a class variable now won't this mean that all v4 clients end up using the last token that was defined in here? What is the scope of this block and do we need to keep the token around after defining this `headers` method? For example could this just be `{ "Authorization" => "bearer #{token}" }`?

johnbintz-tidelift, line 73: One way to avoid needing a class variable is taking advantage of the fact that `HTTP.new` creates an object that can have other methods attached to it. Something like this may be possible:

```ruby
http_adapter = GraphQL::Client::HTTP.new("https://api.github.com/graphql") do
  attr_accessor :token

  def headers(_context)
    { "Authorization" => "Bearer #{token}" }
  end
end

http_adapter.token = token
```

macowie: 💡  Way too simple for me to see after looking at it as long as I was

tiegz (approved):

PR #3233, opened 2023-11-04
Fetch graphql rate limits from same request as rest limits

macowie:
Octokit hides the [full details](https://docs.github.com/en/rest/rate-limit/rate-limit?apiVersion=2022-11-28) we can get from this endpoint, but we can inspect the full response and check a token's rate limit for both REST and GraphQL from the same request.

This means we don't have to involve both APIs just for checking if a token can be used, and cut down some code paths.

### DIFF
--- a/app/models/auth_token.rb
+++ b/app/models/auth_token.rb
@@ -54,8 +54,12 @@ class AuthToken < ApplicationRecord
     remaining > LOW_RATE_LIMIT_REMAINING_THRESHOLD
   rescue Octokit::Unauthorized, Octokit::AccountSuspended
     false
-  rescue StandardError
-    StructuredLog.capture("FAILED_READING_GITHUB_RATE_LIMITS", { api_version: api_version, auth_token_id: id, remaining: remaining })
+  rescue StandardError => e
+    StructuredLog.capture(
+      "FAILED_READING_GITHUB_RATE_LIMITS",
+      { api_version: api_version, auth_token_id: id, remaining: remaining, error_class: e, error_message: e.message }
+    )
+
     false
   end
 
--- a/app/models/auth_token.rb
+++ b/app/models/auth_token.rb
@@ -15,6 +15,8 @@ class AuthToken < ApplicationRecord
   validates_presence_of :token
   scope :authorized, -> { where(authorized: [true, nil]) }
 
+  LOW_RATE_LIMIT_REMAINING_THRESHOLD = 500
+
   def self.client(options = {})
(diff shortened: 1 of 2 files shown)


app/models/auth_token.rb
```diff
@@ -33,12 +35,28 @@ def self.create_multiple(array_of_tokens)
...
+  def safe_to_use?(api_version)
+    resource_limits = fetch_resource_limits
+    remaining = resource_limits.fetch(api_version)
+
+    remaining > LOW_RATE_LIMIT_REMAINING_THRESHOLD
   rescue Octokit::Unauthorized, Octokit::AccountSuspended
     false
+  rescue StandardError
+    StructuredLog.capture("FAILED_READING_GITHUB_RATE_LIMITS", { api_version: api_version, auth_token_id: id, remaining: remaining })
+    false
```
macowie, line 59: It might be more appropriate to just let the exception raise at this point, but I do want to see when/how often it actually comes up

app/models/auth_token.rb
```diff
@@ -33,12 +35,28 @@ def self.create_multiple(array_of_tokens)
...
+
+  def safe_to_use?(api_version)
+    resource_limits = fetch_resource_limits
+    remaining = resource_limits.fetch(api_version)
+
+    remaining > LOW_RATE_LIMIT_REMAINING_THRESHOLD
   rescue Octokit::Unauthorized, Octokit::AccountSuspended
     false
+  rescue StandardError
+    StructuredLog.capture("FAILED_READING_GITHUB_RATE_LIMITS", { api_version: api_version, auth_token_id: id, remaining: remaining })
```
tiegz, line 58: is it worth logging the error class here too?

app/models/auth_token.rb
```diff
@@ -84,34 +106,11 @@ def headers(_context)
 
   def self.find_token(api_version, retries: 0)
     auth_token = authorized.order(Arel.sql("RANDOM()")).first
-    return auth_token if auth_token.high_rate_limit?(api_version)
+    return auth_token if auth_token.safe_to_use?(api_version)
```
jhan217, line 109: If it's not safe to use, should we just stop instead of retrying a few more times?

mikeyoung85 (approved):

app/models/auth_token.rb
macowie: If we don't pass this check, it'll fetch a new random record next time around

tiegz (approved):

PR #3234, opened 2023-11-06
Add one off rake task to delete ignored Maven versions and resync aff…

kjmahalingam:
The intent is for this rake task and its specs to be deleted after it is run.

This will find all Maven versions that are not from only Maven or Google as a source, delete them, resync the packages associated with these versions, then delete the package if no versions remain.

### DIFF
--- a/lib/tasks/one_off.rake
+++ b/lib/tasks/one_off.rake
@@ -257,4 +257,46 @@ namespace :one_off do
     puts "Total packages checked: #{input_tsv_file.count}"
     puts "Versions corrected: #{corrected_count}"
   end
+
+  desc "Delete ignored Maven versions and resync packages"
+  task :delete_ignored_maven_versions_and_resync_packages, %i[commit] => :environment do |_t, args|
+    commit = args.commit.present? && args.commit == "yes"
+
+    puts "DRY RUN" unless commit
+
+    # Retrieve the projects where at least one version's repository_sources is not solely ["Maven"] or ["Google"]
+    affected_projects = Project
+      .joins(:versions)
+      .where(platform: "Maven")
+      .where(%(versions.repository_sources != '["Maven"]' AND versions.repository_sources != '["Google"]' OR repository_sources IS NULL))
+      .distinct
+
+    puts "Count of projects with versions to re-process: #{affected_projects.count}"
+
+    batch_size = 50
+    processed_count = 0
+
+    puts "Processing...."
+    affected_projects.in_batches(of: batch_size).each do |batch|
+      batch.each do |project|
+        affected_versions = project.versions.where(%(repository_sources != '["Maven"]' AND repository_sources != '["Google"]' OR repository_sources IS NULL))
+
+        puts "Destroying #{affected_versions.count} versions for #{project.platform}/#{project.name}."
+        affected_versions.destroy_all if commit
+
+        # Use versions count manually because it might be unsafe to trust
+        # cached project.versions_count value immediately after doing deletions
+        if project.versions.count > 0
+          puts "Versions remaining for #{project.platform}/#{project.name}: Trying manual sync."
+          project.try(:manual_sync) if commit
+        else
+          puts "No versions remaining for #{project.platform}/#{project.name}: Destroying project."
+          project.destroy! if commit
+        end
+      end
+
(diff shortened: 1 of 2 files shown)


spec/tasks/one_off_spec.rb
```diff
@@ -63,4 +63,57 @@
       expect(maven2_version_dependency.platform).to eq("Maven")
     end
   end
+
+  describe "delete_ignored_maven_versions_and_resync_packages" do
```
kjmahalingam, line 67: I did not add expectations to ensure the manual_sync is actually run because that behavior is pretty standard, and this spec covers the destructive behavior. If we want to add any checks for that though we can do that.

lib/tasks/one_off.rake
```diff
@@ -257,4 +257,45 @@ namespace :one_off do
...
+
+    puts "Count of projects with versions to re-process: #{affected_projects.count}"
+
+    batch_size = 50
+    processed_count = 0
+
+    puts "Processing...."
+    affected_projects.in_batches(of: batch_size).each do |batch|
+      batch.each do |project|
+        affected_versions = project.versions.where(%(repository_sources != '["Maven"]' AND repository_sources != '["Google"]' OR repository_sources IS NULL))
```
jhan217, line 282: Is it `(repository_sources != '["Maven"]' AND repository_sources != '["Google"]') OR repository_sources IS NULL` or `repository_sources != '["Maven"]' AND (repository_sources != '["Google"]' OR repository_sources IS NULL)`?

kjmahalingam: It is the former, which I believe is the default SQL behavior. I can make that explicit for readability if we want!

spec/tasks/one_off_spec.rb
```diff
@@ -63,4 +63,57 @@
...
+      let(:google_source_version) { create(:version, project: project, number: "2.0.0", repository_sources: ["Google"]) }
+      let(:no_source_version) { create(:version, project: project, number: "3.0.0", repository_sources: nil) }
+      let(:ignored_source_version) { create(:version, project: project, number: "4.0.0", repository_sources: ["Other"]) }
+
+      it "deletes ignored versions but not the project or its other versions" do
+        Rake::Task["one_off:delete_ignored_maven_versions_and_resync_packages"].invoke("yes")
+
+        expect(Project.all).to match_array([project])
+        expect(Version.all).to match_array([maven_source_version, google_source_version])
+      end
```
macowie, line 104: One thing here, this is testing if these records get deleted by making sure they don't exist, but where do we ensure they ever existed? This test still passes even without running the task

macowie (changes requested):

spec/tasks/one_off_spec.rb
kjmahalingam: Hm good point. I usually don't add an expectation that spec setup worked correctly, and only add expectations for the actual behavior we're testing. But you're right, so maybe it's worth doing that for these specs.

kjmahalingam: Alternatively could do some `let!` in the setup as well, though not sure I prefer that. Will investigate a good approach and fix either way, good catch.

jhan217: Can you use the `change` matcher?

macowie: Wrapping it in a `change` expectation would be nice. Using `let!` should work fine as well. I wouldn't say you necessarily need to assert that they exist, just to realize that's the assumption being made (and to make sure to start with a test that fails)

kjmahalingam: I made the relevant changes here, although if there's a better way to write the specs feel free to let me know. Not sure, for example, if there's a good way to confirm a `.not_change` without having to write expectations before calling the task, to confirm initial state.

This said, these specs won't hang around after the task is run, so it's not a huge deal to fix as long as we're confident it works.

macowie: Thanks for including specs in the first place. The task itself looked correct I just wanted to back up and make sure we didn't commit a classic goof

For future ref you shooooould be able to write `expect { some_op }.to not_change(thing.prop1).from("some initial value")` -- ie just omit the `.to()` matcher.

macowie (approved):

PR #3236, opened 2023-11-07
Split up Maven cleanup one-off into two tasks

kjmahalingam:
In the original rake task, the `project.manual_sync` was an asynchronous process, meaning that it could be potentially unsafe to delete a `project` if it had no versions without the `manual_sync` finishing, since that could potentially create new versions again. This splits the work into two rake tasks; one to delete the versions, then a second to delete Maven projects without versions after the async workers are finished.

### DIFF
--- a/lib/tasks/one_off.rake
+++ b/lib/tasks/one_off.rake
@@ -259,8 +259,9 @@ namespace :one_off do
   end
 
   desc "Delete ignored Maven versions and resync packages"
-  task :delete_ignored_maven_versions_and_resync_packages, %i[commit] => :environment do |_t, args|
-    commit = args.commit.present? && args.commit == "yes"
+  task :delete_ignored_maven_versions_and_resync_packages, %i[max commit] => :environment do |_t, args|
+    max = args[:max].presence.to_i == 0 ? nil : args[:max].to_i
+    commit = args[:commit].present? && args[:commit] == "yes"
 
     puts "DRY RUN" unless commit
 
@@ -270,6 +271,7 @@ namespace :one_off do
       .where(platform: "Maven")
       .where(%((versions.repository_sources != '["Maven"]' AND versions.repository_sources != '["Google"]') OR repository_sources IS NULL))
       .distinct
+    affected_projects = affected_projects.limit(max) if max.present?
 
     puts "Count of projects with versions to re-process: #{affected_projects.count}"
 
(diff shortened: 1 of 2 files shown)


lib/tasks/one_off.rake
```diff
@@ -259,8 +259,9 @@ namespace :one_off do
   end
 
   desc "Delete ignored Maven versions and resync packages"
-  task :delete_ignored_maven_versions_and_resync_packages, %i[commit] => :environment do |_t, args|
-    commit = args.commit.present? && args.commit == "yes"
+  task :delete_ignored_maven_versions_and_resync_packages, %i[max commit] => :environment do |_t, args|
+    max = args[:max].presence.to_i == 0 ? nil : args[:max].to_i
```
kjmahalingam, line 263: There are two optional arguments. To handle the case where we want the `max` missing but the `commit` present, I'm just treating the first one here so that a non-numerical value will mean it defaults to not being used.

lib/tasks/one_off.rake
```diff
@@ -285,17 +287,30 @@ namespace :one_off do
...
+  desc "Delete Maven packages without versions"
+  task :delete_maven_packages_without_versions, %i[commit] => :environment do |_t, args|
+    commit = args.commit.present? && args.commit == "yes"
+
+    puts "DRY RUN" unless commit
+
+    batch_size = 10000
+    processed_count = 0
+
+    affected_projects = Project.where(platform: "Maven").without_versions
```
kjmahalingam, line 307: This can technically catch Maven projects that weren't affected by the `delete_ignored_maven_versions_and_resync_packages` task, but happen to have no versions anyway, and delete them. I am assuming this is likely fine, but if it isn't then we might need to be smarter about that.

spec/tasks/one_off_spec.rb
```diff
@@ -104,22 +100,69 @@
...
-            .and change(Project, :all).from([project]).to([])
+        end
+      end
+    end
+  end
+
+  describe "delete_maven_packages_without_versions" do
+    after(:each) do
+      Rake::Task["one_off:delete_maven_packages_without_versions"].reenable
+    end
```
macowie, line 124: Was it complaining about running a task twice in one test?

kjmahalingam: This was pre-emptive; I needed to add a similar `after` to the specs for `delete_ignored_maven_versions_and_resync_packages` because if I ran individual tests they would pass, but if I ran the full block some would fail because it thought the task had already been run. I had assumed similar would be the case for this task since there are multiple tests in the same block.

lib/tasks/one_off.rake
```diff
@@ -285,17 +287,30 @@ namespace :one_off do
...
+    puts "Count of projects without versions to destroy: #{affected_projects.count}"
+    puts "Processing...."
+    affected_projects.in_batches(of: batch_size).each do |batch|
+      batch.destroy_all if commit
+      processed_count += [batch_size, batch.size].min
```
macowie, line 312: When would the `processed_count` be the desired batch size and not the actual `batch.size`?

kjmahalingam: Good call, never would happen! Will patch that.

lib/tasks/one_off.rake
```diff
@@ -285,17 +287,30 @@ namespace :one_off do
...
+    puts "DRY RUN" unless commit
+
+    batch_size = 10000
+    processed_count = 0
+
+    affected_projects = Project.where(platform: "Maven").without_versions
+    puts "Count of projects without versions to destroy: #{affected_projects.count}"
+    puts "Processing...."
+    affected_projects.in_batches(of: batch_size).each do |batch|
+      batch.destroy_all if commit
```
johnbintz-tidelift, line 311: WDYT about printing the projects to be deleted if `commit` is false?

macowie (approved):

macowie (approved):

PR #3238, opened 2023-11-09
Update Maven cleanup rake task to update and not delete versions with ignored and non-ignored sources

kjmahalingam:
Previously, if there was a version with multiple repository sources, and one was ignored and another wasn't, this rake task would delete the version. Now, it doesn't do that, and instead just deletes the ignored repository sources.

### DIFF
--- a/lib/tasks/one_off.rake
+++ b/lib/tasks/one_off.rake
@@ -279,17 +279,34 @@ namespace :one_off do
     processed_count = 0
 
     puts "Processing...."
-    affected_projects.in_batches(of: batch_size).each do |batch|
-      batch.each do |project|
-        affected_versions = project.versions.where(%((repository_sources != '["Maven"]' AND repository_sources != '["Google"]') OR repository_sources IS NULL))
+    affected_projects.in_batches(of: batch_size).each do |projects_batch|
+      projects_batch.each do |project|
+        no_source_versions = project.versions.where("repository_sources IS NULL")
+        ignored_source_versions = project.versions.where(%((repository_sources != '["Maven"]' AND repository_sources != '["Google"]')))
 
-        puts "Destroying #{affected_versions.count} versions for #{project.platform}/#{project.name}."
-        affected_versions.destroy_all if commit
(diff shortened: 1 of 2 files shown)


lib/tasks/one_off.rake
```diff
@@ -279,17 +279,34 @@ namespace :one_off do
...
+        if commit
+          # If no repository_sources, then destroy it
+          no_source_versions.destroy_all
+
+          # If there are repository_sources,
+          # - if any are Maven or Google, remove repository_sources that aren't Maven or Google
+          # - if none are Maven or Google, destroy version
+          ignored_source_versions.in_batches do |versions_batch|
+            versions_batch.each do |version|
+              if version.repository_sources.include?("Maven") || version.repository_sources.include?("Google")
```
jhan217, line 297: Do the repository_sources use a consistent casing or are there variations of 'maven' that we'd want to keep?

lib/tasks/one_off.rake
```diff
@@ -279,17 +279,34 @@ namespace :one_off do
...
     puts "Processing...."
-    affected_projects.in_batches(of: batch_size).each do |batch|
-      batch.each do |project|
-        affected_versions = project.versions.where(%((repository_sources != '["Maven"]' AND repository_sources != '["Google"]') OR repository_sources IS NULL))
-
-        puts "Destroying #{affected_versions.count} versions for #{project.platform}/#{project.name}."
-        affected_versions.destroy_all if commit
+    affected_projects.in_batches(of: batch_size).each do |projects_batch|
+      projects_batch.each do |project|
+        no_source_versions = project.versions.where("repository_sources IS NULL")
```
jhan217, line 284: Are there any with empty arrays?

jhan217 (approved):

lib/tasks/one_off.rake
kjmahalingam: They should be consistent, given my understanding!

jhan217 (approved):

PR #3243, opened 2023-11-13
Add way to get versions_from_html as opposed to searching through Maven metadata

kjmahalingam:
For Maven Central, search through HTML index directory for versions instead of using Maven metadata. This is required because Maven Central metadata is not guaranteed to include all versions.

### DIFF
--- a/app/models/package_manager/maven.rb
+++ b/app/models/package_manager/maven.rb
@@ -160,8 +160,6 @@ module PackageManager
       if raw_project && raw_project[:versions]
         raw_project[:versions]
       else
-        # TODO: Use the commented below instead when specs are fixed
-        # retrieve_versions(versions_from_html(name), name)
         xml_metadata = maven_metadata(name)
         xml_versions = Nokogiri::XML(xml_metadata).css("version").map(&:text)
         retrieve_versions(xml_versions.filter { |item| !item.ends_with?("-SNAPSHOT") }, name)
@@ -255,12 +253,6 @@ module PackageManager
         .css("versioning > latest, versioning > release, metadata > version")
         .map(&:text)
         .first
-    end
-
-    def self.versions_from_html(name)
-      get_html(MavenUrl.from_name(name, repository_base, NAME_DELIMITER).base).css("a").filter_map do |a|
-        a.text.chomp("/") if a.text.ends_with?("/") && a.text != "../"
-      end
     end
 
     def self.db_platform
--- a/app/models/package_manager/maven.rb
+++ b/app/models/package_manager/maven.rb
@@ -160,6 +160,8 @@ module PackageManager
       if raw_project && raw_project[:versions]
         raw_project[:versions]
       else
+        # TODO: Use the commented below instead when specs are fixed
+        # retrieve_versions(versions_from_html(name), name)
         xml_metadata = maven_metadata(name)
         xml_versions = Nokogiri::XML(xml_metadata).css("version").map(&:text)
         retrieve_versions(xml_versions.filter { |item| !item.ends_with?("-SNAPSHOT") }, name)
@@ -253,6 +255,12 @@ module PackageManager
         .css("versioning > latest, versioning > release, metadata > version")
         .map(&:text)
         .first
+    end
+
+    def self.versions_from_html(name)
+      get_html(MavenUrl.from_name(name, repository_base, NAME_DELIMITER).base).css("a").filter_map do |a|
+        a.text.chomp("/") if a.text.ends_with?("/") && a.text != "../"
+      end
     end
 
     def self.db_platform
--- a/app/models/package_manager/maven.rb
+++ b/app/models/package_manager/maven.rb
@@ -192,8 +192,9 @@ module PackageManager
       versions
         .map do |version|
           one_version_for_name(version, name)
-      rescue Ox::Error, POMNotFound
-        next
+        rescue Ox::Error, POMNotFound
+          StructuredLog.capture("MAVEN_RETRIEVE_VERSION_FAILURE", { version: version, name: name, message: "Version POM not found or valid" })
+          next
         end
         .compact
     end
--- a/app/models/package_manager/maven/maven_central.rb
+++ b/app/models/package_manager/maven/maven_central.rb
@@ -21,6 +21,8 @@ class PackageManager::Maven::MavenCentral < PackageManager::Maven::Common
     PackageManager::Base::MissingVersionRemover
   end
 
+  # maven-metadata.xml for Maven Central does not appear to be guaranteed to contain all relevant versions for a package
+  # So instead, if needed, we will retrieve the versions from the raw HTML index page
   def self.versions(raw_project, name)
     if raw_project && raw_project[:versions]
       raw_project[:versions]
--- a/app/models/package_manager/maven/maven_central.rb
+++ b/app/models/package_manager/maven/maven_central.rb
@@ -20,4 +20,18 @@ class PackageManager::Maven::MavenCentral < PackageManager::Maven::Common
   def self.missing_version_remover
     PackageManager::Base::MissingVersionRemover
   end
+
+  def self.versions(raw_project, name)
+    if raw_project && raw_project[:versions]
+      raw_project[:versions]
+    else
+      retrieve_versions(versions_from_html(name), name)
+    end
+  end
+
+  def self.versions_from_html(name)
+    get_html(MavenUrl.from_name(name, repository_base, NAME_DELIMITER).base).css("a").filter_map do |a|
+      a.text.chomp("/") if a.text.ends_with?("/") && a.text != "../"
+    end
+  end
 end
--- a/spec/models/package_manager/maven/maven_central_spec.rb
+++ b/spec/models/package_manager/maven/maven_central_spec.rb
@@ -29,4 +29,13 @@ describe PackageManager::Maven::MavenCentral do
       end
     end
   end
+
+  describe ".versions_from_html" do
+    it "retrieves versions from Maven Central index HTML" do
+      VCR.use_cassette("maven-central/lambda-tracer") do
+        # Matching versions as of November 2023
+        expect(described_class.versions_from_html("com.appdynamics:lambda-tracer")).to match_array(%w[1.1.1363 1.2.1390 20.03.1391 20.11.1400])
+      end
+    end
+  end
 end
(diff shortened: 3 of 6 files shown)


app/models/package_manager/maven/maven_central.rb
```diff
@@ -20,4 +20,20 @@ def self.one_version(raw_project, version_string)
   def self.missing_version_remover
     PackageManager::Base::MissingVersionRemover
   end
+
+  # maven-metadata.xml for Maven Central does not appear to be guaranteed to contain all relevant versions for a package
```
jhan217, line 24: What assumptions are being made here & can we validate/invalidate those assumptions (e.g. log when those assumptions are no longer true)?

app/models/package_manager/maven/maven_central.rb
```diff
@@ -20,4 +20,20 @@ def self.one_version(raw_project, version_string)
...
+  def self.versions(raw_project, name)
+    if raw_project && raw_project[:versions]
+      raw_project[:versions]
+    else
+      retrieve_versions(versions_from_html(name), name)
+    end
+  end
+
+  def self.versions_from_html(name)
+    get_html(MavenUrl.from_name(name, repository_base, NAME_DELIMITER).base).css("a").filter_map do |a|
```
jhan217, line 35: Should we navigate into the folder & persist the timestamp from an artifact?

app/models/package_manager/maven/maven_central.rb
```diff
@@ -20,4 +20,20 @@ def self.one_version(raw_project, version_string)
...
+      a.text.chomp("/") if a.text.ends_with?("/") && a.text != "../"
```
jhan217, line 36: Do we know how reliable the folder structure maps to actual versions/artifacts (i.e. are all folders versions and all versions folders?)?

app/models/package_manager/maven/maven_central.rb
```diff
@@ -20,4 +20,20 @@ def self.one_version(raw_project, version_string)
...
     PackageManager::Base::MissingVersionRemover
   end
+
+  # maven-metadata.xml for Maven Central does not appear to be guaranteed to contain all relevant versions for a package
+  # So instead, if needed, we will retrieve the versions from the raw HTML index page
+  def self.versions(raw_project, name)
+    if raw_project && raw_project[:versions]
+      raw_project[:versions]
+    else
+      retrieve_versions(versions_from_html(name), name)
```
jhan217, line 30: What happens if the repository is temporarily unavailable?

spec/models/package_manager/maven/maven_central_spec.rb
```diff
@@ -13,7 +13,7 @@
       # a version that was removed on Maven Central that we still know about
       let!(:remotely_removed_version) { project.versions.create(number: "1.0.2-1361", status: nil) }
 
-      # a version that, as of October 2023, was still available on Maven Central
+      # a version that, as of November 2023, was still available on Maven Central
```
jhan217, line 16: Is there a case for when we have both the .xml file & the file structure but with differing info to ensure we're getting only data from one or the other?

Any reason to not add unit tests specifically for the new methods?

app/models/package_manager/maven/maven_central.rb
kjmahalingam: If the repository is temporarily unavailable then I think this will just not retrieve the versions; which was already the case with the previous more generic method this was using seen here: https://github.com/librariesio/libraries.io/blob/main/app/models/package_manager/maven.rb#L159-L167

app/models/package_manager/maven/maven_central.rb
kjmahalingam: Technically no, we do not reliably know this. But this should not be an issue because in `retrieve_versions` we handle cases where we do not find a version `.pom` at the correct address, so any invalid matches should be handled. https://github.com/librariesio/libraries.io/blob/main/app/models/package_manager/maven.rb#L195-L198

app/models/package_manager/maven/maven_central.rb
kjmahalingam: What are you referring to in this situation, exactly? Are you referring to how we know that the metadata isn't guaranteed to contain all versions? If so, we have determined that to be true from some instances of this that we've noticed already, where a metadata file did not contain all the versions listed in the index.

If you are referring to the assumptions being made in terms of the HTML format we're parsing, then yes, perhaps there are some things we could do to log if we don't get what we expect.

spec/models/package_manager/maven/maven_central_spec.rb
kjmahalingam: There isn't a case for this, no, but I don't think that's needed; the VCR exists for the response that we expect the test to trigger, and for Maven Central that is this index HTML response. Elsewhere, the VCR for Google Maven uses the generic method, which reads maven-metadata.xml, and that verifies that.

We could add unit tests specifically for the new methods, I just hadn't done so yet. I can add some.

kjmahalingam: I added new tests to verify that the network responses for both Maven Central and Google Maven are pulling versions from the correct sources.

app/models/package_manager/maven/maven_central.rb
kjmahalingam: Additional thought: we shouldn't need to persist a timestamp, because in `retrieve_versions` we will still be collecting `.pom` files which will have the info we need for each version we collect.

app/models/package_manager/maven/maven_central.rb
jhan217: Stepping back to look at the overall Maven version data landscape - This method is adding another source of where we get version data. This means we can now get version data from at least three places:
1.) Version data from the maven directory
2.) Version data from the maven .xml file
3.) Version data stored in libraries already

We have discovered that 2 has data issues, and we are currently trying to rectify data issues around 3. We have looked at 1 and have come to the conclusion that the data seems more accurate than than 2 - but to my knowledge, we've only checked a small % of all Maven packages, which may or may not be a representative sample. Is there anything we can check or cross reference to reinforce or disprove the conclusion that a) 1 is always better than 2 and b) 1 is the most correct source of maven version data? Could we do something to let us know if things seem to be trending in the wrong direction / let us know that this data source is decaying and we should look for another source?

app/models/package_manager/maven/maven_central.rb
jhan217: Isn't the published_at an important data point for versions though?

app/models/package_manager/maven/maven_central.rb
jhan217: Can we log something if what is expected to be a real version based off the string is not actually a real version so we can improve the regex matching in the future?

app/models/package_manager/maven/maven_central.rb
kjmahalingam: Yeah I think you're correct, but we are getting all the same data that we previously would at a later step in the process, from the `.pom`s. This change in the process does not interfere with that.

https://github.com/librariesio/libraries.io/pull/3243/files#diff-a54c889fd056c7288db14f084d2ef1a242e3a96fa6d5b8ee6daa6116673f2615R30

https://github.com/librariesio/libraries.io/blob/main/app/models/package_manager/maven.rb#L189-L199

https://github.com/librariesio/libraries.io/blob/main/app/models/package_manager/maven.rb#L170

app/models/package_manager/maven/maven_central.rb
jhan217: Ok, as long as we are not marking everything as removed or something when there is a temporary server issue.

kjmahalingam: I will investigate this a bit, because this new methodology will cause the same behavior in a temporary server issue as it would previously. But agreed that if it would mark everything as removed in that case, we probably should fix that too.

app/models/package_manager/maven/maven_central.rb
kjmahalingam: Yup; I think that we can trust that this data source is likely correct for us to use in the short term, but agreed it would be good to add some logging about assumptions we're making so that we can reassess in the future if needed. I can do that.

kjmahalingam: An additional note to make: This is not exactly replacing source (2) above with source (1). Functionally, we already depend on the directory structure from source (1) in order to get the data from the POMs. So really this new directory structure parsing isn't new, as we already depended on these directories existing in the previous implementation. All this PR is doing is removing source (2) in favor of source (1) which we already use.

app/models/package_manager/cran.rb
```diff
@@ -25,7 +25,7 @@ def self.check_status_url(db_project)
     end
 
     def self.project_names
-      html = get_html("https://cran.r-project.org/web/packages/available_packages_by_date.html", request: {timeout: 5})
+      html = get_html("https://cran.r-project.org/web/packages/available_packages_by_date.html", request: { timeout: 5 })
```
kjmahalingam, line 28: The build was failing because of these lint errors after I rebased, since it seems like `main` was failing on them. These changes are just for that fix and are totally unrelated.

app/models/package_manager/maven/maven_central.rb
kjmahalingam: We discussed this further and decided that these logs for the diff should be unnecessary. Since we already use the directory structure to find the POMs we need, and we are using the same directory structure here, we are necessarily going to find equal or more versions using this approach than with using the metadata.

app/models/package_manager/maven/maven_central.rb
kjmahalingam: I have added a log here if a version is not found, in case the directory structure includes more directories we don't expect.

app/models/package_manager/maven.rb
```diff
@@ -192,8 +192,9 @@ def self.retrieve_versions(versions, name)
       versions
         .map do |version|
           one_version_for_name(version, name)
-      rescue Ox::Error, POMNotFound
-        next
+        rescue Ox::Error, POMNotFound
+          StructuredLog.capture("MAVEN_RETRIEVE_VERSION_FAILURE", { version: version, name: name, message: "Version POM not found or valid" })
```
jhan217, line 196: 🎱 Might be nice to log the url / directory that we attempted to look at here to make it easier for debugging.

jhan217 (approved):

app/models/package_manager/maven.rb
kjmahalingam: In order to do this, we'd either have to reconstruct the `MavenUrl` that we generate [here](https://github.com/librariesio/libraries.io/blob/main/app/models/package_manager/maven.rb#L202) which would risk them both getting out of sync, or we would have to pass more arguments between methods, which I think is not necessary. I think that for debugging have this information will likely be sufficient, as it will inform the arguments passed into the method to make the request, so personally I'm inclined to leave this log as it is.

PR #3244, opened 2023-11-13
Fix Packagist dependencies update.

tiegz:
fixes this error, which is getting silenced into logs only, and which has happened about 1 million times in the past 2 weeks: `"ERROR Error while trying to get dependencies for Packagist/PACKAGE/NAME@1.0.0: no implicit conversion of String into Integer"`.

### DIFF
--- a/app/models/package_manager/packagist.rb
+++ b/app/models/package_manager/packagist.rb
@@ -105,7 +105,7 @@ module PackageManager
     end
 
     def self.dependencies(_name, version, mapped_project)
-      vers = mapped_project[:versions][version]
+      vers = mapped_project[:versions].find { |v| v["version"] == version }
       return [] if vers.nil?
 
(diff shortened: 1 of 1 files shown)


app/models/package_manager/packagist.rb
```diff
@@ -105,7 +105,7 @@ def self.acceptable_versions(versions)
     end
 
     def self.dependencies(_name, version, mapped_project)
-      vers = mapped_project[:versions][version]
+      vers = mapped_project[:versions].find { |v| v["version"] == version }
```
jhan217, line 108: Is this packagist specific?

tiegz: yeah, this is specifically for the Packagist project response

macowie (approved):

PR #3246, opened 2023-11-13
Log how often we update projects without/with changes.

tiegz:
log every time a project would be sending webhooks, to see how often we send webhooks when a project hasn't vs has changed.

(the project's changes itself aren't the full story, as there might be version, dependency, etc changes as well)


macowie (approved):

tiegz:
giving up on this one for now, because the effort to check if a project has changed or not will be substantial.

PR #3247, opened 2023-11-14
Add some loggers for SidekiqUniqueJob events

tiegz:


config/initializers/sidekiq.rb
```diff
@@ -24,6 +24,8 @@ class SidekiqQuietRetryError < StandardError; end
     chain.add SidekiqUniqueJobs::Middleware::Server
   end
 
+  SidekiqUniqueJobs::Server.configure(config)
```
tiegz, line 27: @macowie not sure if this is the issue, but I noticed this was missing from the 7.1 directions: https://github.com/mhenrixon/sidekiq-unique-jobs#add-the-middleware

macowie (approved):

PR #3248, opened 2023-11-14
Fix sidekiq unique jobs options

macowie:
Note to self: open PR for their upgrading guide


tiegz (approved):
funny detail to leave out of an upgrading guide 😆

PR #3252, opened 2023-11-14
Adds some useful debugging for initial missing deps bug.

tiegz:
there's an occasional bug we have where a version is ingested (via PackageManagerDownloadWorker) but the dependencies aren't saved on the initial download.

It's not clear why this is happening, so this PR adds some debugging code to monitor these a bit.

### DIFF
--- a/app/models/package_manager/base.rb
+++ b/app/models/package_manager/base.rb
@@ -260,7 +260,7 @@ module PackageManager
       db_versions = db_versions.where(number: sync_version) unless sync_version == :all
 
       if db_versions.empty?
-        StructuredLog.capture("SAVE_DEPENDENCIES_FAILURE", { platform: db_platform, name: name, version: version, message: "no versions found" })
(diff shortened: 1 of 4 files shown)


kjmahalingam (approved):

macowie (approved):

app/models/package_manager/base.rb
```diff
@@ -258,6 +258,11 @@ def self.save_dependencies(mapped_project, sync_version: :all, force_sync_depend
       db_project = Project.find_by(name: name, platform: db_platform)
       db_versions = db_project.versions.includes(:dependencies)
       db_versions = db_versions.where(number: sync_version) unless sync_version == :all
+
+      if db_versions.empty?
+        StructuredLog.capture("SAVE_DEPENDENCIES_FAILURE", { platform: db_platform, name: name, version: version, message: "no versions found" })
```
macowie, line 263: Oh, there's no `version` in this method

macowie (changes requested):

macowie (approved):

PR #3253, opened 2023-11-15
Set a reasonable 5-second timeout for CRAN requests.

tiegz:
CRAN is offline for maintenance today, so set a default timeout for HTML requests (the default seems to be a few minutes).

### DIFF
--- a/app/models/package_manager/cran.rb
+++ b/app/models/package_manager/cran.rb
@@ -25,7 +25,7 @@ module PackageManager
     end
 
     def self.project_names
-      html = get_html("https://cran.r-project.org/web/packages/available_packages_by_date.html")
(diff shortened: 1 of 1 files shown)


app/models/package_manager/cran.rb
```diff
@@ -25,7 +25,7 @@ def self.check_status_url(db_project)
     end
 
     def self.project_names
-      html = get_html("https://cran.r-project.org/web/packages/available_packages_by_date.html")
+      html = get_html("https://cran.r-project.org/web/packages/available_packages_by_date.html", request: {timeout: 5})
```
pranav, line 28: might wanna set this globally somwhere

pranav (approved):

PR #3254, opened 2023-11-15
PackageManager::CPAN.dependencies(): fix this by passing the CPAN API "_source" instead of "field"

tiegz:
we need to select "dependency" as a returned field from their ES API, but "dependency" is an Object itself so ES complains that it's not a leaf node. Using `_source` lets us return that field specifically.

[docs on _source vs field](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-fields.html)
[docs on metacpan's API](https://github.com/metacpan/metacpan-api/blob/master/docs/API-docs.md)

### DIFF
--- a/app/models/package_manager/cpan.rb
+++ b/app/models/package_manager/cpan.rb
@@ -46,19 +46,19 @@ module PackageManager
     end
 
     def self.versions(raw_project, _name)
-      versions = get("https://fastapi.metacpan.org/v1/release/_search?q=distribution:#{raw_project['distribution']}&size=5000&fields=version,date")["hits"]["hits"]
+      versions = get("https://fastapi.metacpan.org/v1/release/_search?q=distribution:#{raw_project['distribution']}&size=5000&_source=version,date")["hits"]["hits"]
       versions.map do |version|
         {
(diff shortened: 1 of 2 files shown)


app/models/package_manager/cpan.rb
```diff
@@ -46,19 +46,19 @@ def self.mapping(raw_project)
...
+          published_at: version["_source"]["date"],
         }
       end
     end
 
     def self.dependencies(_name, version, mapped_project)
-      versions = mapped_project[:versions]
-      version_data = versions.find { |v| v["fields"]["version"] == version }
-      version_data["fields"]["dependency"].select { |dep| dep["relationship"] == "requires" }.map do |dep|
+      versions = get("https://fastapi.metacpan.org/v1/release/_search?q=distribution:#{mapped_project[:name]}&size=5000&_source=version,dependency")["hits"]["hits"]
```
tiegz, line 59: we're also passing in `mapped_project` here, which only has `number` and `published_at`, so we have to make an extra request to get the dependencies.

kjmahalingam (approved):

PR #3255, opened 2023-11-16
Implements Version#dependencies_count

tiegz:
we need a "have dependencies been fetched yet?" flag on the Version model. To that end:

* adds `Version#dependencies_count`, which will be `nil` if we haven't fetched its deps yet
  * (note that we have an existing `#runtime_dependencies_count` column)
* adds `Version#set_dependencies_count` method 
* call `Version#set_dependencies_count` after we've fetched a Version's dependencies (whether or not there are zero)
* adds `one_off:backfill_version_dependencies_count` task to backfill this column in batches

### DIFF
--- a/spec/models/package_manager/pypi_spec.rb
+++ b/spec/models/package_manager/pypi_spec.rb
@@ -228,7 +228,6 @@ describe PackageManager::Pypi do
         expect(version.dependencies.count).to be 6
       end
 
-
       it "updates the dependencies counters" do
         expect(version.dependencies.count).to be 1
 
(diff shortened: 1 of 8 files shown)


spec/models/package_manager/pypi_spec.rb
```diff
@@ -228,6 +228,18 @@
         expect(version.dependencies.count).to be 6
       end
 
+      it "updates the dependencies counters" do
+        expect(version.dependencies.count).to be 1
+
+        expect do
+          VCR.use_cassette("pypi_dependencies_requests") do
+            described_class.save_dependencies(mapped_project, sync_version: version_number, force_sync_dependencies: true)
```
mpace965, line 236: worth a spec for the 0 dependencies case?

mpace965 (approved):

PR #3256, opened 2023-11-16
Api::Projects#dependencies_bulk: return "dependencies: nil" when deps haven't been fetched yet

tiegz:
when a release's dependencies haven't been fetched yet, return `dependencies: nil` instead of `dependencies: []`, as a signal that the client can try again later.

(followup to https://github.com/librariesio/libraries.io/pull/3255 )


havocp (approved):
Ha! I _knew_ there was gonna be some nil-vs-empty confusion here and I just couldn't find it in a quick look. Nice spotting.

PR #3257, opened 2023-11-16
Adds a worker to backfill Version#dependencies_count

tiegz:
the backfill rake task `one_off:backfill_version_dependencies_count` (for https://github.com/librariesio/libraries.io/pull/3255) is taking too long to babysit, so this introduces a worker that does a ~10 second batch of updates, and then re-spawns itself if needed.


tiegz:
we have an alternate way to backfill these in https://github.com/librariesio/libraries.io/pull/3258

PR #3260, opened 2023-11-21
Add login prompt to API docs page

macowie:
image


app/views/api/docs/_login_prompt.html.erb
```diff
@@ -0,0 +1,12 @@
+<div class="alert alert-info" role="alert">
+  <h4 class="alert-heading">
+    Want to use Libraries.io for Data-API capabilities?
+  </h3>
+
+  <p>
+    API access is free to all registered users. All you need to do is create an account to
+    get access to these capabilities! (its free!)
```
bee64, line 8:
```suggestion
    get access to these capabilities! (It's free!)
```

as much as i love to type in all lowercase i don't think it's the usual libraries style :p

bee64 (commented):
lgtm, with one copy nit

app/views/api/docs/_login_prompt.html.erb
macowie: fair enough :p

PR #3261, opened 2023-11-22
Catch some empty payloads and better log errors

macowie:
Reduces some of the Bugsnag noise and make it easier to catch issues with a particular service

### DIFF
--- a/app/models/repository_host/bitbucket.rb
+++ b/app/models/repository_host/bitbucket.rb
@@ -2,6 +2,7 @@
 
 module RepositoryHost
   class Bitbucket < Base
+    class EmptyResponseError < RuntimeError; end
     IGNORABLE_EXCEPTIONS = [
       BitBucket::Error::NotFound,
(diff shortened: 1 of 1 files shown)


app/models/repository_host/bitbucket.rb
```diff
@@ -201,5 +229,16 @@ def self.api_client(token = nil)
...
       self.class.api_client(token)
     end
+
+    def log_error(repo_name:, request:, error: nil)
+      log_info = {
+        external_service: "bitbucket",
+        request_label: request,
+        repository_full_name: repo_name,
+      }
+      log_info.merge!(error_message: error.message, error_name: error.name) if error
```
kjmahalingam, line 239: If this method is for logging an error and it's only called when we're rescuing, is there a reason `error` would ever be `nil`?

kjmahalingam (approved):

tiegz (approved):

PR #3264, opened 2023-11-29
Fix up the repository dashboard

katzj:
This should fix https://app.bugsnag.com/tidelift/libraries-dot-io/errors/64c8081643b15e0008352b41?filters[error.status]=open&filters[event.since]=30d&pivot_tab=event

### DIFF
--- a/app/models/concerns/github_identity.rb
+++ b/app/models/concerns/github_identity.rb
@@ -23,11 +23,11 @@ module GithubIdentity
 
   def github_settings_url
     if private_repo_token.present?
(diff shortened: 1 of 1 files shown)


app/models/concerns/github_identity.rb
```diff
@@ -23,11 +23,11 @@ def hidden=(val)
 
   def github_settings_url
     if private_repo_token.present?
```
tiegz, line 25: could also simplify with `key = if private_repo_token.present?` instead of per-line

tiegz (approved):

PR #3266, opened 2023-12-01
Combine Project.find_exact and find_lower into one SQL query.

tiegz:
the `.or()` ActiveRelation method lets us combine two conditions. This is one step towards fixing N+1s while [updating project dependencies](https://github.com/librariesio/libraries.io/blob/30e729a77e9799a368cb3376bf72c6a34c71744e/app/models/package_manager/base.rb#L290)

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -466,23 +466,15 @@ class Project < ApplicationRecord
   end
 
   def self.find_best!(platform, name, includes = [])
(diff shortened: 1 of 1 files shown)


tiegz:
actually, we probably want to avert the "lower()" query if the first query works. I'm going to switch directions toward bulk loading these instead.

PR #3270, opened 2023-12-03
Follow redirects for repos that are moved on GitLab

tiegz:
fixes bugsnag 655efe17f52efa0007687c7e

### DIFF
--- a/app/models/repository_owner/gitlab.rb
+++ b/app/models/repository_owner/gitlab.rb
@@ -163,6 +163,7 @@ module RepositoryOwner
     def get_json(url)
       r = Typhoeus::Request.new(url,
                                 method: :get,
+                                followlocation: true,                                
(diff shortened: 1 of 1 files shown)


app/models/repository_owner/gitlab.rb
```diff
@@ -163,6 +163,7 @@ def self.create_org(org_hash)
     def get_json(url)
       r = Typhoeus::Request.new(url,
                                 method: :get,
+                                followlocation: true,
```
tiegz, line 166: Typhoeus doesn't let you set a limit, but assuming this is safe because the url is always GitLab

macowie: WDYT about using `PackageManager::ApiService.request_and_parse_json` here?

macowie (approved):

PR #3272, opened 2023-12-04
 Fix a "stack level too deep" bug with repository owner org updates.

tiegz:
fixes bugsnag 65527440cf4e030007888c55

this is the same fix as in https://github.com/librariesio/libraries.io/pull/3271 , but for RepositoryOrganisations instead of RepositoryUsers.

I found a case with an org that had moved their login to a new account, and the old account got a new login. After this fix, the account with the stale login should be deleted, which would then get re-added by Repository#download_owner at some point. These records haven't been updated in 5 years, so fixing them like this is better than having the stale data around.


macowie (approved):
🎉

PR #3273, opened 2023-12-05
Remove unnecessary/broken validations on Repository{User,Organisation}

tiegz:
this removes 4 uniqueness validations on RepositoryUser and RepositoryOrganisation. The two "host_type/login" validations were broken because they don't work for the case where a login changes case (e.g. "FOO" -> "foo"). Another reason is that unique indices already exist for these column pairs (uuid/host_type) (lower(host_type)/lower(login)), so we don't need extra ruby logic to handle these.

fixes bugsnag 656e06f8feff8200073ef0cd


macowie:
Ope, there's a shoulda matcher test for this validation in RepositoryOrganisation spec

macowie (approved):

PR #3275, opened 2023-12-12
Version: replace most in-Ruby sorting

mpace965:
Some projects with many versions (several thousand) will take quite a while to render the `Project` show page. Sort `Versions` by `number` in SQL where possible. Limit the semver diff behavior on number of `Versions`.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -200,7 +200,7 @@ class ApplicationController < ActionController::Base
       end
       raise ActiveRecord::RecordNotFound if @tags.empty? && params[:number].present?
     else
-      @versions = @project.versions.order(number: :desc).limit(10)
+      @versions = @project.versions.order(published_at: :desc).limit(10)
       if params[:number].present?
         @version = @project.versions.find_by_number(params[:number])
         raise ActiveRecord::RecordNotFound if @version.nil?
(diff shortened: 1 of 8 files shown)


tiegz (approved):

mpace965:
This spec doesn't look like it should be passing in the first place, but I can't reproduce it locally. It looks nominally related to my changes (in that the failure has to do with `Versions`), but I don't understand why it would cause this Rake task to fail

https://app.circleci.com/pipelines/github/librariesio/libraries.io/2046/workflows/414f4839-8b3e-40c6-a28b-a09028565d96/jobs/6367/tests

tiegz (approved):

app/controllers/application_controller.rb
```diff
@@ -200,7 +200,7 @@ def find_version
       end
       raise ActiveRecord::RecordNotFound if @tags.empty? && params[:number].present?
     else
-      @versions = @project.versions.sort.first(10)
+      @versions = @project.versions.order(number: :desc).limit(10)
```
mikeyoung85, line 203: Would it make more sense to use `published_at` for this? It looks like the sorting for `Version` tends to use that first before defaulting to a string comparison on the version number if I'm reading the sort code correctly. That also might prevent weird versions showing up at the top of the list like the occasional `v99999.9999.9999` that show up every once in a while.

https://github.com/librariesio/libraries.io/blob/71da9e378da59ddd69dbaf7132bf8e934421cb8c/app/models/version.rb#L145-L151

mpace965: Changed to sort by `published_at` and added an on `versions.published_at`

tiegz (approved):

PR #3278, opened 2023-12-14
Disable single version updates for NPM

jhan217:
NPM is currently unreliable in its update publishing: https://github.com/npm/feedback/discussions/937 The updates are coming in out of order, which is throwing off single version updates.

In order to increase its reliability, we are disabling per-version updates for NPM until the update issues affecting the CouchDB feed are resolved.

It looks like the main entrypoint to this is the PackageManagerDownloadWorker, so I'll monitor that when this change goes out. If there's any other places to watch, let me know and I can keep an eye out.

### DIFF
--- a/app/models/package_manager/npm.rb
+++ b/app/models/package_manager/npm.rb
@@ -9,7 +9,7 @@ module PackageManager
     URL = "https://www.npmjs.com"
     COLOR = "#f1e05a"
     ENTIRE_PACKAGE_CAN_BE_DEPRECATED = true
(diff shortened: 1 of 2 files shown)


tiegz (approved):
yikes! NPM full syncs can be time-consuming, but this PR should help with that: https://github.com/librariesio/libraries.io/pull/3268

tiegz (approved):

johnbintz-tidelift (approved):

PR #3279, opened 2023-12-15
Fix up error when cargo is pointing ot a bogus version

katzj:
If the version doesn't actually exist, we get no body and it's throwing an error. Just assume the best case

(accidentally pushed straight to main but let's go ahead and review before deploying)


macowie (approved):
Oooh I wasn't sure what the main root cause was here but getting a bogus version would do it

PR #3280, opened 2023-12-20
feat: [sc-34983] Create a worker on Libraries that regularly polls npm and Maven Central for packages that are tracking maintenance stats

macowie:
Story details: https://app.shortcut.com/tidelift/story/34983


app/workers/proactive_project_sync_worker.rb
```diff
@@ -0,0 +1,29 @@
...
+    Project
+      .distinct
+      .visible
+      .platform(PLATFORMS)
+      .where.associated(:repository_maintenance_stats)
+      .where("projects.last_synced_at IS NULL OR projects.last_synced_at < ?", last_synced_at_cutoff)
+      .order("projects.last_synced_at ASC NULLS FIRST")
+  end
+
+  def last_synced_at_cutoff
```
johnbintz-tidelift, line 26: WDYT about making this a constant instead of a public instance method?

app/workers/proactive_project_sync_worker.rb
```diff
@@ -0,0 +1,29 @@
...
+  PLATFORMS = %w[NPM Maven].freeze
+  DEFAULT_LIMIT = 1000
+
+  def perform(limit = DEFAULT_LIMIT)
+    projects_query
+      .limit(limit)
+      .each(&:async_sync)
+  end
+
+  def projects_query
```
johnbintz-tidelift, line 16: Since this isn't called by anything other than this worker, it would be best to remove it from the public interface. It means the test for `perform` will have to do the heavy lifting of testing more cases, but it will make the worker's public interface less complex. I don't think it's worth moving this query to `Project` since it's pretty specific to this use case.

app/workers/proactive_project_sync_worker.rb
```diff
@@ -0,0 +1,29 @@
...
+  include Sidekiq::Worker
+  sidekiq_options queue: :small, lock: :until_executed
+
+  PLATFORMS = %w[NPM Maven].freeze
+  DEFAULT_LIMIT = 1000
+
+  def perform(limit = DEFAULT_LIMIT)
+    projects_query
+      .limit(limit)
+      .each(&:async_sync)
```
johnbintz-tidelift, line 13: In case the passed in `limit` is large, using `find_each` will prevent instantiating a ton of models:

```suggestion
      .find_each(&:async_sync)
```

app/workers/proactive_project_sync_worker.rb
```diff
@@ -0,0 +1,29 @@
...
+  end
+
+  def projects_query
+    Project
+      .distinct
+      .visible
+      .platform(PLATFORMS)
+      .where.associated(:repository_maintenance_stats)
+      .where("projects.last_synced_at IS NULL OR projects.last_synced_at < ?", last_synced_at_cutoff)
+      .order("projects.last_synced_at ASC NULLS FIRST")
```
jhan217, line 23: Are we guaranteed that this always gets populated? Would like to avoid repeated polling of null last_synced_ats.

app/workers/proactive_project_sync_worker.rb
macowie: If it's a constant, wouldn't it get frozen to whatever the value is at load time? That's what I was avoiding

app/workers/proactive_project_sync_worker.rb
macowie: Using `find_each` or the other ActiveRecord::Batches methods would throw out the ordering, and I think it makes sense to attempt the staler rows first?

app/workers/proactive_project_sync_worker.rb
macowie: No, it can be null. There's only ~1,000 from this set that are null, but I don't think it's a bad state

jhan217: As long as we are not always filling up the first ~1000 slots with the same packages every time

spec/workers/proactive_project_sync_worker_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+end
+
+RSpec::Matchers.define :have_not_been_queued_for_project_sync do
+  match do |actual|
+    expect(PackageManagerDownloadWorker).not_to receive(:perform_async).with(String, actual.name, any_args)
+    expect(CheckStatusWorker).not_to receive(:perform_async).with(actual.id)
+  end
+end
+
```
macowie, line 18: How do we feel about using custom matchers here to make the specs fluent? Easier to follow or nah?

johnbintz-tidelift: Easier to follow, this is a good abstraction. :+1:

app/workers/proactive_project_sync_worker.rb
macowie: I don't believe this will be the case now, but I do agree it would be nice to have some warning if/when this starts happening. We should see fewer nulls and the mean last_synced_at for this set increase (while accounting for the passage of time)

johnbintz-tidelift (approved):

PR #3286, opened 2024-01-04
Add alternative Elasticless project search

macowie:
This adds an alternative project search using only postgres' full text search (via [PgSearch](https://github.com/Casecommons/pg_search)). The goal of this first draft is just to get enough things in place such that we can start internal testing, get more eyes on it, and figure out the refinements later. As well as getting some baseline performance metrics along the way.


app/views/projects/_search_form.html.erb
```diff
@@ -4,10 +4,18 @@
...
       <%= hidden_field_tag field, params[field] if params[field].present? %>
     <% end %>
+
     <span class="input-group-btn">
       <button class="btn btn-primary" type="submit">
         <%= fa_icon 'search' %>
       </button>
     </span>
+
+    <% if logged_in? && (current_user.admin? || current_user.current_api_key.is_internal?) %>
```
katzj, line 14: Maybe add an `admin_or_internal?` method on `User` that does the two checks that could be used here?

app/controllers/application_controller.rb
```diff
@@ -184,6 +184,21 @@ def es_query(klass, query, filters)
...
+      platforms: current_platforms,
+      languages: current_languages,
+      keywords: current_keywords,
+      licenses: current_licenses,
+      sort: format_sort
+    ).results
+  end
+
+  def pg_search_projects_enabled?
+    params.key?("force_pg") || Rails.application.config.pg_search_projects_enabled
```
katzj, line 199: Probably only want to allow the param to be passed if the user is admin/internal?

app/controllers/search_controller.rb
```diff
@@ -3,24 +3,35 @@
...
+
+      respond_to do |format|
+        format.html { render :index_pg_search }
+        format.atom
+      end
+    else
+      @search = []
+      @projects = []
+      @facets = []
+      @title = page_title
```
katzj, line 22:
```suggestion
```

Already set above

katzj (approved):
This looks awesome to me. Added a few small suggestions but I'm game to give it a go

PR #3287, opened 2024-01-04
More Consistent Maintenance Stat Collection

mikeyoung85:
This PR does a few things:

* Raises an error when errors occur during stat collection instead of catching the errors and just ignoring those stats. This should make the stats more consistent and avoid the issue where one stat was collected while others are still their old data. This should also help notify us about issues gathering stats.

* Creates a `maintenance_stats_refreshed_at` column to track when we last attempted to do a refresh. This should help us avoid querying for the same repositories over and over if we cannot get stats for them. This still needs a follow up PR to use that column for queuing work. I will also likely need to manually run the database migration which is why it has `if_not_exists: true` on the add_column and add_index commands.

* One off rake task to get the dates set for repositories that have existing stats. I will need to run this manually for the 120k repositories that have stats now.

### DIFF
--- a/app/models/maintenance_stats/queries/query_utils.rb
+++ b/app/models/maintenance_stats/queries/query_utils.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+module MaintenanceStats
+  module Queries
+    class QueryError < StandardError; end
+
+    class QueryUtils
+      def self.check_for_graphql_errors(result)
+        if result.data.nil?
+          raise QueryError, "No data found in result"
+        end
+
+        if result.errors.any?
+          error_messages = result.errors.values.flatten
+          raise QueryError, "Error exists in query: #{error_messages}"
+        end
+
+        if result.data.errors.any?
+          error_messages = result.data.errors.values.flatten
+          raise QueryError, "Error exists in result: #{error_messages}"
+        end
+      end
+    end
+  end
+end
--- a/app/models/maintenance_stats/queries/github/commit_count_query.rb
+++ b/app/models/maintenance_stats/queries/github/commit_count_query.rb
@@ -57,7 +57,10 @@ module MaintenanceStats
           # merge params for query
           date_params.merge!(params.slice(:owner, :repo_name))
 
-          @client.query(COMMIT_COUNTS_QUERY, variables: date_params)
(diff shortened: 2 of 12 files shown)


app/models/maintenance_stats/queries/query_utils.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+module MaintenanceStats
+  module Queries
+    class QueryError < StandardError; end
+
+    class QueryUtils
+      def self.check_for_graphql_errors(result)
+        if result.data.nil?
+          raise QueryError, "No data found in result"
```
jhan217, line 10: Are these worth logging so that we can track them over time? If so, might also be worth including more details to help with debugging.

jhan217 (approved):

app/models/maintenance_stats/queries/query_utils.rb
mikeyoung85: Updated the error checking a bit to put some structured logs in with the error messages and the repository name as fields.

app/models/maintenance_stats/queries/query_utils.rb
```diff
@@ -0,0 +1,46 @@
...
+            "GITHUB_STAT_QUERY_ERROR",
+            {
+              repository_name: repository_name,
+              error_messages: error_messages,
+            }
+          )
+          raise QueryError, "Error exists in query: #{error_messages}"
+        end
+
+        if result.data.errors.any?
```
tiegz, line 32: just to check, have you run into any cases where we get a `data.errors` key back?

mikeyoung85: This [spec](https://github.com/librariesio/libraries.io/blob/270fe79f915321a672b8f4a49201a8562793483c/spec/models/repository_spec.rb#L256-L274) is responding with errors in `data.errors`. I'm going off of what was already here as I don't remember many of the details of how this all works. The GraphQL client page has a [small example](https://github.com/github/graphql-client/blob/master/guides/handling-errors.md) on handling different errors. I think `result.data.errors` would be the most common.

jhan217 (approved):

tiegz (approved):

PR #3288, opened 2024-01-05
Use Repository Refresh Date for Scheduling Stat Refresh

mikeyoung85:
Updated the Project scopes to find Projects with the oldest stats that should be refreshed. This value is updated during the stat refresh so we can avoid always selecting repositories that no longer exist on GitHub every time.

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -123,8 +123,10 @@ class Project < ApplicationRecord
   has_many :repository_maintenance_stats, through: :repository
 
   scope :updated_within, ->(start, stop) { where("updated_at >= ? and updated_at <= ? ", start, stop).order(updated_at: :asc) }
-  scope :least_recently_updated_stats, -> { joins(repository: :repository_maintenance_stats).where(repositories: { host_type: "GitHub" }).group("projects.id").order(Arel.sql("max(repository_maintenance_stats.updated_at) ASC")) }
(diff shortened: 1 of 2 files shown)


app/models/project.rb
```diff
@@ -123,8 +123,10 @@ class Project < ApplicationRecord
   has_many :repository_maintenance_stats, through: :repository
 
   scope :updated_within, ->(start, stop) { where("updated_at >= ? and updated_at <= ? ", start, stop).order(updated_at: :asc) }
-  scope :least_recently_updated_stats, -> { joins(repository: :repository_maintenance_stats).where(repositories: { host_type: "GitHub" }).group("projects.id").order(Arel.sql("max(repository_maintenance_stats.updated_at) ASC")) }
-  scope :no_existing_stats, -> { where.missing(:repository_maintenance_stats).where(repositories: { host_type: "GitHub" }) }
+  # restrict to only GitHub repositories where we have attempted to refresh the stats previously
+  scope :least_recently_updated_stats, -> { joins(:repository).where(repositories: { host_type: "GitHub" }).where.not(repositories: { maintenance_stats_refreshed_at: nil }).merge(Repository.order(maintenance_stats_refreshed_at: :asc)) }
```
jhan217, line 127: 🎱 Might be worth adding "github" to the scope name somewhere to make it more obvious to the downstream user of the scope

jhan217 (approved):

PR #3290, opened 2024-01-05
Fix Go's check_status_url and only mark "Removed" if it's pre-existing.

tiegz:
currently a pre-existing Go Project that was removed will not be marked as "Removed" because we're checking the Go index API, which returns a 200 but wouldn't return any versions. Since [we've already decided to only store Go Modules that are on pkg.go.dev](https://github.com/librariesio/libraries.io/blob/134010e61aefb0a91df8f525d0f3e07efca4de13/app/models/package_manager/go.rb#L124-L125), this changes the status url from the API to pkg.go.dev, which returns 404 for non-existent Go Modules.

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -633,6 +633,10 @@ class Project < ApplicationRecord
     response = Typhoeus.get(url)
     if platform.downcase == "packagist" && [302, 404].include?(response.response_code)
       update_attribute(:status, "Removed")
+    elsif platform.downcase == "go" && [404].include?(response.response_code)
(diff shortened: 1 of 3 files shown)


app/models/project.rb
```diff
@@ -633,7 +633,11 @@ def check_status
     response = Typhoeus.get(url)
     if platform.downcase == "packagist" && [302, 404].include?(response.response_code)
       update_attribute(:status, "Removed")
-    elsif platform.downcase != "packagist" && [400, 404, 410].include?(response.response_code)
+    elsif platform.downcase == "go" && [404].include?(response.response_code)
```
johnbintz-tidelift, line 636: If there's only one status code, do we need to use an array?

PR #3292, opened 2024-01-08
Move Stat Logic Into Repository and Add API to Queue Stats for Repository Directly

mikeyoung85:
This moves some of the calls and logic for queuing maintenance stats out of `Project` and into `Repository`. This also adds a new API to queue up stats gathering for repositories directly instead of needing to go through the Project.


app/controllers/api/maintenance_stats_controller.rb
```diff
@@ -19,6 +19,30 @@ def begin_watching_bulk
...
+    auth_token = AuthToken.find_token(:v3)
+
+    repos = lookup_names.map do |repo_host_and_name|
+      return render json: { error: "#{repo_host_and_name[:host_type]} is not a supported host" }, status: :bad_request unless supported_host_types.include?(repo_host_and_name[:host_type].downcase)
+
+      Repository.create_from_host(repo_host_and_name[:host_type], repo_host_and_name[:full_name], auth_token.token)
+    end
+
+    # compact repos array to remove any invalid/not found repositories
+    repos.compact.each do |repo|
```
mikeyoung85, line 39: This decision impacts what happens when we cannot find a repository by name. The code now basically just ignores them and queues up stats for the repositories we did find. This means if we have invalid repository names we won't get an error back from this API nor will it start retrieving stats for them.

app/models/repository.rb
```diff
@@ -321,9 +325,9 @@ def self.create_from_hash(repo_hash)
       g.assign_attributes repo_hash.slice(*Repository::API_FIELDS)
 
       if g.changed?
-        return g.save ? g : nil
```
mikeyoung85, line 324: Fun fact. In the latest version of Rails, returning early in a transaction will roll the transaction back. If I have time I would like to comb through the rest of the code in Libraries to see if we are doing this in other places.

jhan217 (approved):

PR #3298, opened 2024-01-16
Ensure that we don't ingest a Go Package that is not a Go Module.

tiegz:
some Go Packages are slipping into Libraries that are not actual Go Modules (reminder: a "Package" in Go lives within a "Module". "Modules" must have go.mod files with their name at the top). e.g. this is a "Package": https://libraries.io/go/k8s.io%2Fkubernetes%2Fthird_party%2Fforked%2Freflect

this change returns `nil` while scraping the package HTML from pkg.go.dev if it is not a Go Module.


app/models/package_manager/go.rb
```diff
@@ -128,6 +130,17 @@ def self.project(name)
       # base package manager handles if the project is not present
       return nil if doc_html.text.blank?
 
+      # pages on pkg.go.dev can be listed as 'package', 'module', or both. Only scrape Go Modules.
+      page_type = doc_html.css(".go-Main-headerTitle .go-Chip").text
+      is_package = page_type.include?("package")
+      is_module = page_type.include?("module")
+      if !is_package && !is_module
```
jhan217, line 137: Minor de-duplication:

```
if !is_module
  if !is_package
    raise UnknownGoType, "Unknown Go type (it is neither a package nor module) for #{name}: #{page_type}"
  end
  return nil
end
```

spec/models/package_manager/go_spec.rb
```diff
@@ -44,6 +44,22 @@
     end
   end
 
+  describe "#project" do
+    it "returns raw project data from pkg.go.dev" do
+      VCR.use_cassette("go/pkg_go_dev") do
```
jhan217, line 49: 🎱 Would be nice if only the relevant snippet was used here so it's easier to see what's being tested, vs replaying the entire request

mikeyoung85 (commented):
Did you see this section of code? Should this be updated with your HTML checking logic and reused?

https://github.com/librariesio/libraries.io/blob/c3a19b92477c6a6439da397f4147b29356921a4c/app/models/package_manager/go.rb#L254-L268

spec/models/package_manager/go_spec.rb
tiegz: the `project()` method returns raw data from upstream, so I think we need the whole HTTP request to test that it's working.

tiegz:
@mikeyoung85 yeah, I wasn't sure if it was safe to merge mine with that one, but maybe we can just say "a Go Module must have the 'module' label and also have a go.mod file." Probably safe

app/models/package_manager/go.rb
```diff
@@ -128,7 +130,18 @@ def self.project(name)
...
+      raw_project = { name: name, html: doc_html, overview_html: doc_html }
+
+      # pages on pkg.go.dev can be listed as 'package', 'module', or both. We only scrape Go Modules.
+      unless is_a_module?(raw_project: raw_project)
+        if !is_a_project?(raw_project: raw_project) && !is_a_command?(raw_project: raw_project) && !is_a_directory?(raw_project: raw_project) && defined?(Bugsnag)
+          # this is more for our record-keeping so we know what possible types there are.
+          Bugsnag.notify(UnknownGoType.new("Unknown Go project type (it is neither a package, a command or a module) for #{name}: #{page_type}"))
+        end
+        return nil
+      end
```
tiegz, line 142: @mikeyoung85 I basically moved the `is_a_module?` into a method that `module?` can use below, but I couldn't use `module?` itself here because there are a lot of actually legit Go Modules that are marked as "Not valid go.mod" on pkg.go.dev, so we should probably continue to ingest those

spec/models/package_manager/go_spec.rb
```diff
@@ -259,16 +275,17 @@
...
+        VCR.use_cassette("go/update/docker-distribution") do
           expect do
-            package = described_class.update("github.com/cloudfoundry/noaa/errors")
+            # this example doesn't have a valid go.mod, is a "module", and has only incompatible versions
+            package = described_class.update("github.com/docker/distribution")
           end.to change(Project, :count).from(0).to(1)
         end
 
-        expect(package.versions.count).to eq(0)
+        expect(package.versions.count).to eq(53)
```
tiegz, line 288: I had to switch from `noaa/errors` to `docker/distribution` here because the former wasn't technically a Go Module. Also noteworthy that we do ingest "incompatible" versions, so this is more explicit about that too now

app/models/package_manager/go.rb
```diff
@@ -128,7 +130,18 @@ def self.project(name)
       # base package manager handles if the project is not present
       return nil if doc_html.text.blank?
 
-      { name: name, html: doc_html, overview_html: doc_html }
+      raw_project = { name: name, html: doc_html, overview_html: doc_html }
+
+      # pages on pkg.go.dev can be categorized as 'package', 'module', 'command', or 'directory'. We only scrape Go Modules.
+      unless module_type?(raw_project: raw_project)
```
johnbintz-tidelift, line 136: WDYT about wrapping up this `raw_project` analysis into a separate object, rather than creating more public methods to process passed-in state on this class?

tiegz: yeah, that might make sense, or I was also thinking of parameterizing it, e.g. `has_type?("module", raw_project: raw_project)`. I'll play around with that now

app/models/package_manager/go.rb
```diff
@@ -128,7 +130,19 @@ def self.project(name)
       # base package manager handles if the project is not present
       return nil if doc_html.text.blank?
 
-      { name: name, html: doc_html, overview_html: doc_html }
+      raw_project = { name: name, html: doc_html, overview_html: doc_html }
+
+      # pages on pkg.go.dev can be categorized as 'package', 'module', 'command', or 'directory'. We only scrape Go Modules.
+      page_types = page_types(raw_project: raw_project)
```
tiegz, line 136: @johnbintz-tidelift WDYT about this single method instead?

tiegz:
I've tested these changes on a sample of 1000 Go Modules from libraries, and only 36 of them were not valid (i.e. they wouldn't get ingested after my changes), which confirms more that this PR should be safe.

app/models/package_manager/go.rb
```diff
@@ -280,6 +300,11 @@ def self.encode_for_proxy(str)
       str.gsub(/[A-Z]/) { |s| "!#{s.downcase}" }
     end
 
+    # Returns the types listed at the top of pkg.go.dev pages. Known values are: module, package, directory, command.
+    private_class_method def self.page_types(raw_project: nil)
```
mikeyoung85, line 305: Why default to nil here? I think we need nil protection if we are not requiring `raw_project` to actually be passed in.

mikeyoung85 (approved):

PR #3300, opened 2024-01-19
avoid triggering callbacks on repository update when gathering stats

mikeyoung85:


app/models/repository_host/github.rb
```diff
@@ -231,8 +231,10 @@ def gather_maintenance_stats
 
       # We have a valid token to run the queries so
       # set refreshed_at time so we don't try and refresh
-      # this repository again if the information is bad
-      repository.update!(maintenance_stats_refreshed_at: now)
+      # this repository again if the information is bad.
+      # Use update_column to avoid triggering callbacks that reassess
```
jhan217, line 235: Were there any specific callbacks getting triggered that were problematic? Just wondering, as it might become a bug in the future since expecting updates to a column to trigger callbacks is the norm.

mikeyoung85: These are the callbacks defined in `Repository`:
```
  before_save  :normalize_license_and_language
  after_commit :save_projects, on: :update
  after_commit :update_source_rank_async, on: [:update]
```

The source rank one can be fairly expensive. Normalizing license and language strings is unnecessary work since that data has not been changed here. It is also calling `force_save` on any attached Projects which will mark them as updated which is probably marking these incorrectly as recently updated. We aren't changing any meaningful data for the rest of the repository so there should not be a need to run any callbacks for this update.

jhan217 (approved):

PR #3303, opened 2024-01-23
chore: [sc-39323] Move prerelease calculations closer together

johnbintz-tidelift:
Story details: https://app.shortcut.com/tidelift/story/39323

### DIFF
--- a/app/domain/prerelease_for_platform.rb
+++ b/app/domain/prerelease_for_platform.rb
@@ -1,0 +1,28 @@
+# frozen_string_literal: true
+
+# This consolidates the prelease checking logic into one place.
+# It only checks against specific platforms, not against the
+# general cases.
+class PrereleaseForPlatform
+  PYPI_PRERELEASE = /(a|b|rc|dev)[-_.]?[0-9]*$/.freeze
+
+  def self.prerelease?(version_number:, platform:)
+    new(version_number: version_number, platform: platform).prerelease?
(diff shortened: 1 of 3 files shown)


app/domain/prerelease_for_platform.rb
```diff
@@ -0,0 +1,28 @@
...
+  PYPI_PRERELEASE = /(a|b|rc|dev)[-_.]?[0-9]*$/.freeze
+
+  def self.prerelease?(version_number:, platform:)
+    new(version_number: version_number, platform: platform).prerelease?
+  end
+
+  def initialize(version_number:, platform:)
+    @version_number = version_number
+    @platform = platform.downcase
+  end
```
mikeyoung85, line 16: Would it be simpler to have the class method only? I can't think of a reason we would need to hold on to an instance of this rather than just call the method directly which is what the code is doing now. Will there be other methods added in here?

mikeyoung85 (commented):
It's probably out of scope for what you wanted out of this PR but this would be a good time to get some specs added for this logic.

app/domain/prerelease_for_platform.rb
johnbintz-tidelift: I think it can be class method-only.

mpace965 (approved):

mikeyoung85 (approved):

PR #3308, opened 2024-01-25
Queue PackageManagerDownloadWorker if it previously failed

tradiff:
If the caller is interested in dependencies for a version which has never successfully updated these, queue an update. But only if it's been at least 1 day since the last retry.

### DIFF
--- a/app/controllers/api/projects_controller.rb
+++ b/app/controllers/api/projects_controller.rb
@@ -147,14 +147,14 @@ class Api::ProjectsController < Api::ApplicationController
     project_json[:dependencies_for_version] = version.number
     deps = version.dependencies.includes(:project)
 
-    # TODO: we may eventually be able to remove this, once all old Versions with depenencies have valid "dependencies_count" values instead of nil.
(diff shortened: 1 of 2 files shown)


app/controllers/api/projects_controller.rb
```diff
@@ -152,6 +152,11 @@ def find_project_as_json_with_dependencies!(platform, name, version_name, subset
       version.update_column(:dependencies_count, deps.size)
     end
 
+    if version.dependencies_count.nil? && deps.empty? && version.updated_at < 1.day.ago
+      version.touch
+      PackageManagerDownloadWorker.perform_async(platform, name, version.number)
+    end
```
tiegz, line 158: I just realized we'd never hit this bc of line 151-153 right now. I think we might be able to replace those lines with your stanza?

tiegz: woops, nm, I just realized the difference between these two. Could combine them into the same `if version.dependencies_count.ni?` block?

tiegz (approved):

PR #3313, opened 2024-02-07
PackageManager::Go.find_project_names: skip certain hosts

tiegz:
when we lookup Go Module names that we can't find to check for an alternative path to use, there are certain hosts that will be slow to respond. This adds a list of hosts that we want to explicitly skip when looking up Go Modules in `find_project_names` so we don't hold up `/api/checks` requests.

### DIFF
--- a/app/models/package_manager/go.rb
+++ b/app/models/package_manager/go.rb
@@ -12,6 +12,9 @@ module PackageManager
       "launchpad.net",
       "hub.jazz.net",
     ].freeze
+    SKIP_HOSTS = [
+      "jfrog.com"
+    ]
     KNOWN_VCS = [
       ".bzr",
       ".fossil",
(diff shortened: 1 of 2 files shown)


app/models/package_manager/go.rb
```diff
@@ -245,7 +249,18 @@ def self.project_find_names(name)
...
+        chosen_name = go_import&.start_with?(*KNOWN_HOSTS) ? [go_import] : [name]
+        # Collect info on what data we're mapping here, to ensure it's doing the right thing.
+        StructuredLog.capture(
+          "GO_MODULES_PROJECT_FIND_NAMES",
+          {
+            name: name,
+            found_name: go_import,
+            chosen_name: chosen_name,
+          }
+        )
```
tiegz, line 261: I'm half-skeptical that this method is choosing the right name, so this log gives us some evidence towards that.

app/models/package_manager/go.rb
```diff
@@ -254,7 +269,8 @@ def self.project_find_names(name)
         Rails.cache.write("unreachable-go-hosts:#{host}", true, ex: 1.day)
         Bugsnag.notify(e)
         [name]
-      rescue StandardError
+      rescue StandardError => e
+        Bugsnag.notify(e)
```
tiegz, line 273: seems weird to swallow these entirely, we should probably be aware of generic exceptions.

tradiff (approved):

PR #3317, opened 2024-02-20
Clean up Project Provider Map system

johnbintz-tidelift:
The Provider Maps are used to match Version#repository_sources to PackageManager classes for packages in ecosystems where multiple package providers exist. The original system was very lax in selecting the most appropriate provider, especially when multiple were involved.

Theoretically, a package should only be found at one provider, but in the case of Maven packages, this turned out to not be the case -- providers like Hortonworks and Atlassian provided their own versions of packages that also existed in Maven Central. We stopped examining Maven repos others than Google Maven and Maven Central, but unfortunately, some packages still present as belonging to these other less-supported repos.

The true fix for this is to remove any references to the old Maven repo in repository_sources, but that still left in place a Provider Map system that would break again in the case where a provider was added, then removed. This would cause events like a package being marked as removed happening because an old provider was checked and the URL 404'd.s This new code:

* structures the Provider Map system in a more structured, prioritized way
* adds checks to ensure if the data or Provider Map get into bad states, we are notified of it sooner
* centralizes more logic related to building package related URLs

### DIFF
--- a/app/models/package_manager/conda.rb
+++ b/app/models/package_manager/conda.rb
@@ -9,11 +9,11 @@ module PackageManager
     URL = "https://anaconda.org"
     API_URL = "https://conda.libraries.io"
 
-    PROVIDER_MAP = ProviderMap.new(
+    PROVIDER_MAP = ProviderMap.new(prioritized_provider_infos: [
       ProviderInfo.new(identifier: "Main", default: true, provider_class: Main),
       ProviderInfo.new(identifier: "CondaMain", provider_class: Main),
-      ProviderInfo.new(identifier: "CondaForge", provider_class: Forge)
-    )
+      ProviderInfo.new(identifier: "CondaForge", provider_class: Forge),
+    ])
 
     def self.formatted_name
       "conda"
--- a/app/models/package_manager/conda.rb
+++ b/app/models/package_manager/conda.rb
@@ -54,33 +54,18 @@ module PackageManager
       end
     end
 
-    def self.package_link(project, _version = nil)
-      db_version = project.versions.last
-      repository_source = db_version&.repository_sources&.first.presence || "default"
-      PROVIDER_MAP[repository_source].package_link(project)
-    end
-
     def self.install_instructions(db_project, _version = nil)
-      db_version = db_project.versions.last
-      repository_source = db_version&.repository_sources&.first.presence || "default"
-      PROVIDER_MAP[repository_source].install_instructions(db_project)
+      self::PROVIDER_MAP
+        .best_repository_source(project: db_project)
+        .provider_class
+        .install_instructions(db_project)
     end
 
-    PROVIDER_MAP = {
-      "CondaForge" => Forge,
-      "default" => Main,
-      "Main" => Main,
-      "CondaMain" => Main,
-    }.freeze
-
-    def self.providers(project)
-      project
-        .versions
-        .flat_map(&:repository_sources)
-        .compact
-        .uniq
-        .map { |source| PROVIDER_MAP[source] } || [PROVIDER_MAP["default"]]
-    end
+    PROVIDER_MAP = ProviderMap.new(
+      ProviderInfo.new(identifier: "Main", default: true, provider_class: Main),
+      ProviderInfo.new(identifier: "CondaMain", provider_class: Main),
+      ProviderInfo.new(identifier: "CondaForge", provider_class: Forge)
+    )
 
     def self.check_status_url(db_project)
       "#{API_URL}/package/#{db_project.name}"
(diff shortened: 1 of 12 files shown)


app/models/project.rb
```diff
@@ -750,6 +750,29 @@ def mailing_list(include_prereleases: false)
...
+  def repository_sources
+    versions
+      .flat_map(&:repository_sources)
+      .compact
+      .uniq
+  end
+
+  # @return [Version,nil] If version was provided, the specific Version or
+  #                       nil if not found. If not provided, the most
+  #                       recently created Version.
```
jhan217, line 764: Should this be `most_recently_created_version` or `most_recently_published_version`? Alternatively, worth splitting up into two methods so that one method isn't doing too much?

app/models/package_manager/conda.rb
```diff
@@ -54,33 +54,18 @@ def self.download_url(db_project, version = nil)
...
-
-    def self.providers(project)
-      project
-        .versions
-        .flat_map(&:repository_sources)
-        .compact
-        .uniq
-        .map { |source| PROVIDER_MAP[source] } || [PROVIDER_MAP["default"]]
-    end
+    PROVIDER_MAP = ProviderMap.new(
```
jhan217, line 64: Could this constant go to the top with the other constants?

app/models/package_manager/multiple_sources_base/provider_map.rb
```diff
@@ -0,0 +1,66 @@
...
+  class MultipleSourcesBase < Base
+    # Given a set of prioritized ProviderInfo objects, determine the most
+    # likely upstream package repository for a package. This uses the
+    # packages' versions' repository_sources field to
+    class ProviderMap
+      # @param prioritized_provider_infos [Array(ProviderInfo)]
+      def initialize(*prioritized_provider_infos)
+        @prioritized_provider_infos = prioritized_provider_infos
+
+        raise "Need at least one provider marked default" if @prioritized_provider_infos.none?(&:default?)
```
jhan217, line 13: Should there be some sort of check if there are multiple default providers for one ecosystem?

app/models/package_manager/multiple_sources_base/provider_info.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+module PackageManager
+  class MultipleSourcesBase < Base
+    class ProviderInfo
+      attr_reader :identifier, :provider_class
+
+      def initialize(identifier:, provider_class:, default: false)
```
jhan217, line 8: Should this include something about which ecosystem they are a provider for? e.g. 'Pypi' or 'Rubygems'?

app/models/package_manager/multiple_sources_base/provider_map.rb
```diff
@@ -0,0 +1,66 @@
+# frozen_string_literal: true
+
+module PackageManager
+  class MultipleSourcesBase < Base
+    # Given a set of prioritized ProviderInfo objects, determine the most
+    # likely upstream package repository for a package. This uses the
+    # packages' versions' repository_sources field to
+    class ProviderMap
+      # @param prioritized_provider_infos [Array(ProviderInfo)]
+      def initialize(*prioritized_provider_infos)
```
jhan217, line 10: 🎱 Is there a way to make it more obvious to the upstream that the order in which you pass in ProviderInfos is actually meaningful?

app/models/package_manager/multiple_sources_base/provider_map.rb
```diff
@@ -0,0 +1,66 @@
...
+      def providers_for(project:)
+        found_providers = project
+          .repository_sources
+          .map do |source|
+            @prioritized_provider_infos.find { |provider_info| provider_info.identifier == source }
+          end
+          .compact
+
+        if found_providers.empty?
+          unless project.repository_sources.empty?
```
jhan217, line 26: 🎱 `if project.repository_sources.present?`

app/models/project.rb
johnbintz-tidelift: The original code used most recently created, so I'm not changing the calculation for that.

app/models/package_manager/multiple_sources_base/provider_info.rb
johnbintz-tidelift: No, that's not information needed for these objects.

app/models/package_manager/multiple_sources_base/provider_map.rb
jhan217: 🎱 What do you think about forcing the upstream to pass in an array named something like "prioritized_provider_infos"? That would let the initialize method take in additional methods (future expansion of the class) as well as force the upstream to figure out what "prioritized_provider_infos" means before it can be passed in

johnbintz-tidelift: Sure, I can do that.

mikeyoung85 (approved):

PR #3319, opened 2024-02-20
Fix flaky spec

tradiff:
Fixes the flaky spec that failed here:
https://app.circleci.com/pipelines/github/librariesio/libraries.io/2217/workflows/ddbbef63-e414-442b-aa1a-eb22a8baba7e/jobs/6905/tests#failed-test-0

### DIFF
--- a/spec/requests/api/projects_spec.rb
+++ b/spec/requests/api/projects_spec.rb
@@ -18,6 +18,7 @@ describe "Api::ProjectsController" do
     internal_user.current_api_key.update_attribute(:is_internal, true)
     project.reload
     dependent_project.reload
+    project_with_unknown_deps.reload
     version.set_dependencies_count
   end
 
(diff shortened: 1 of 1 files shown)


spec/requests/api/projects_spec.rb
```diff
@@ -105,7 +105,7 @@
     end
 
     it "ignores stuff before start_time" do
-      after_everything = [project.updated_at, dependent_project.updated_at].max + 1
+      after_everything = [project.reload.updated_at, dependent_project.reload.updated_at].max + 1
```
mikeyoung85, line 108: It looks like this should already happening in the `before` block on line 19,20. Is there something happening between that block and this spec that the reload is fixing?

tradiff: Ah good catch. Now I see what happened. The failing test returned this third project that we weren't considering in the `after_everything` calculation.

mikeyoung85 (approved):

PR #3321, opened 2024-02-20
Update Maven delete/resync task to support offset/limit

johnbintz-tidelift:


lib/tasks/one_off.rake
```diff
@@ -285,45 +286,46 @@ namespace :one_off do
...
-          # - if any are Maven or Google, remove repository_sources that aren't Maven or Google
-          # - if none are Maven or Google, destroy version
-          ignored_source_versions.in_batches do |versions_batch|
-            versions_batch.each do |version|
-              if version.repository_sources.include?("Maven") || version.repository_sources.include?("Google")
-                version.update(repository_sources: version.repository_sources.select { |source| %w[Maven Google].include?(source) })
-              else
-                version.destroy
-              end
+    affected_projects.find_each(batch_size: batch_size) do |project|
```
mikeyoung85, line 300: `find_each` is going to ignore the SQL ordering you added on line 289 if that is important. `find_each` uses the rails batching which orders by ID only.

lib/tasks/one_off.rake
```diff
@@ -285,45 +286,46 @@ namespace :one_off do
...
+            else
+              version.destroy
             end
           end
         end
-        puts "Trying manual sync for #{project.platform}/#{project.name}."
-        project.try(:manual_sync) if commit
       end
+      puts "Trying manual sync for #{project.platform}/#{project.name}."
+      project.try(:manual_sync) if commit
```
mikeyoung85, line 323: Why not keep these in the `if commit` block above and remove the extra `if commit` on the manualy_sync line?

lib/tasks/one_off.rake
johnbintz-tidelift: Oh right, I forgot about that after having just fixed that in another one, thanks!

mikeyoung85 (approved):

PR #3324, opened 2024-02-27
Better error for POM file parsing error

johnbintz-tidelift:
It's possible for Ox.parse to return nil when parsing. If that happens, raise a better error than `nil doesn't respond to <<`.

### DIFF
--- a/app/models/package_manager/maven.rb
+++ b/app/models/package_manager/maven.rb
@@ -215,7 +215,7 @@ module PackageManager
       raise POMNotFound, url if pom_request.status == 404
 
       xml = Ox.parse(pom_request.body)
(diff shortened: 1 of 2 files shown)


app/models/package_manager/maven.rb
```diff
@@ -206,6 +215,8 @@ def self.download_pom(group_id, artifact_id, version)
       raise POMNotFound, url if pom_request.status == 404
 
       xml = Ox.parse(pom_request.body)
+      raise POMParseError, url unless xml
```
jhan217, line 218: 🎱 Could we be explicit about what this is checking? `if xml.nil?`

jhan217 (approved):

PR #3327, opened 2024-03-05
chore: [sc-40567] Limit version deletion for large projects

johnbintz-tidelift:


spec/tasks/one_off_spec.rb
```diff
@@ -125,6 +125,32 @@
         end
       end
 
+      context "with many ignored versions" do
+        before do
+          50.times do |i|
```
jhan217, line 130: 🎱 Should be able to do a `create_list(:version, 50, **opts)` but may need to update the factory to use sequences for that to work, same with the below one.

lib/tasks/one_off.rake
```diff
@@ -225,13 +225,22 @@ namespace :one_off do
 
     puts "Count of projects with versions to re-process: #{affected_projects.count}"
 
+    # prevent overloading consumers with too many changes per project
```
jhan217, line 228: Is this going to leave us in a bad state, if only some of the versions that should be removed are removed? How are we going to find and clean up the rest? Will there be dangling references to non-existent objects?

johnbintz-tidelift: No. We find versions to be cleaned up by their `repository_sources` contents. Removing only some of these from a project does not affect the other versions of a project. There would be no dangling references to anything else that Versions use if only part of them were deleted.

I updated the PR with code that illustrates this behavior is working as intended.

jhan217 (approved):

PR #3330, opened 2024-03-11
Stop Removing Go Pseudo Versions

mikeyoung85:
This changes the version remove logic for Go to not mark not found pseudo versions as removed. My rationale is that the upstream Go data sources are not very consistent with storing pseudo version information, but if we have seen a pseudo version previously then it should be considered valid since it is a combination of the module version, git commit date, and git commit hash that existed at some point. More info at https://go.dev/ref/mod#pseudo-versions. I think in order for a version like this to be removed and unreachable then the git history of the project would need to have been rewritten, but feel free to correct me on these assumptions. There will need to be follow up PRs/work to restore pseudo versions which have been marked as removed.

For testing I have added some data to the specs to check that the status is not touched for pseudo versions and have tested with a few packages locally by running a `PackageManager::Go.update()` on them with extra pseudo versions in my database.

### DIFF
--- a/app/models/package_manager/go.rb
+++ b/app/models/package_manager/go.rb
@@ -35,7 +35,7 @@ module PackageManager
     end
 
     def self.missing_version_remover
-      PackageManager::Base::MissingVersionRemover
+      PackageManager::Go::GoMissingVersionRemover
     end
 
     def self.check_status_url(db_project)
--- a/app/models/package_manager/go/go_missing_version_remover.rb
+++ b/app/models/package_manager/go/go_missing_version_remover.rb
@@ -1,0 +1,42 @@
+# frozen_string_literal: true
+
+module PackageManager
+  class Go
+    class GoMissingVersionRemover
+      # should match the same regex that Golang uses to identify pseudo version numbers
+      # https://github.com/golang/mod/blob/master/module/pseudo.go#L47C39-L47C135
+      PSEUDO_VERSION_REGEX = /^v[0-9]+\.(0\.0-|\d+\.\d+-([^+]*\.)?0\.)\d{14}-[A-Za-z0-9]+(\+[0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*)?$/.freeze
+
+      def initialize(
+        project:,
+        version_numbers_to_keep:,
+        target_status:,
+        removal_time:
+      )
+        @project = project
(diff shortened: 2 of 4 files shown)


tiegz (commented):
this makes sense to me since I also have a fuzzy memory of pseudoversions being inconsistent on pkg.go.dev.

One question: we've had the logic in place for a while to [sync a "/v2+" module's base module with the same versions](https://github.com/librariesio/libraries.io/blob/main/app/models/package_manager/go.rb#L104), but then logic with the MissingVersionRemover was added that marked those base module's fake versions as "Deleted" (since they can't actually be found for the base module). 

Is this a good time to stop syncing those fake versions on the base module? I'm wondering if after this PR we'd end up with base module versions for "/v2+" packages that are partly "Removed" (real versions) and partly not-"Removed" (for pseudoversions).

wenottingham:
Will this still handle the entire module itself being removed?

mikeyoung85:
> Will this still handle the entire module itself being removed?

I think that happens in a separate process with https://github.com/librariesio/libraries.io/blob/bc1d969c6415e65e90919585565b2758afdd3e32/app/workers/check_status_worker.rb

mikeyoung85:
> this makes sense to me since I also have a fuzzy memory of pseudoversions being inconsistent on pkg.go.dev.
> 
> One question: we've had the logic in place for a while to [sync a "/v2+" module's base module with the same versions](https://github.com/librariesio/libraries.io/blob/main/app/models/package_manager/go.rb?rgh-link-date=2024-03-11T16%3A21%3A45Z#L104), but then logic with the MissingVersionRemover was added that marked those base module's fake versions as "Deleted" (since they can't actually be found for the base module).
> 
> Is this a good time to stop syncing those fake versions on the base module? I'm wondering if after this PR we'd end up with base module versions for "/v2+" packages that are partly "Removed" (real versions) and partly not-"Removed" (for pseudoversions).

I thought that change had already been made :) I don't think this PR would negatively affect things too much. It would be possible to have incorrect pseudo versions on the base module for any prereleases for module versions v2 and up, but we could go back and clean those up with the removal of the base module sync we are doing now. I'm not sure how common a pseudo version for a versioned module is, but it seems like something that would be possible.

tiegz (approved):

PR #3331, opened 2024-03-12
Remove Custom Update Code to Add Versions to Base Go Module

mikeyoung85:
This code existed to make it easier to lookup versions for a Go module without needing the specific version of the module in the name. Since we have been updating things to be more correct around version information this feature is now a problem and a lot of the versions tied to the base module for version 2.x and up are marked as being removed from the upstream source. This PR removes the custom update method that was creating versions for the base module name if we were updating a versioned module name. There will be some follow up clean up work to get rid of invalid versions on the base module Projects, but this PR will stop creating any new incorrect versions.


jhan217 (approved):

app/models/package_manager/go.rb
```diff
@@ -86,27 +86,6 @@ def self.one_version(raw_project, version_string)
...
-      # when reusing the PackageManagerDownloadWorker to initialize the base name Project
-      base_sync_version = sync_version unless sync_version == :all
-      base_sync_version = base_sync_version.presence || project.versions.first&.number&.presence
-      update_base_module(project.name, base_sync_version) if project.present? && project.name.match?(VERSION_MODULE_REGEX)
-
-      project
-    end
-    # rubocop: enable Lint/UnusedMethodArgument
-
     def self.update_base_module(name, base_sync_version)
```
tiegz, line 89: is it safe to remove this method now too?

mikeyoung85: Thanks for noticing I missed removing that method. PR here: https://github.com/librariesio/libraries.io/pull/3332

PR #3333, opened 2024-03-19
Handle Exhausted Tokens When Setting Authorized

mikeyoung85:
Exceptions are being raised on the scheduled task to remove unauthorized tokens for ones that have hit their limit before we check on them. This should handle that error directly in the task so it the loop can continue.

### DIFF
--- a/lib/tasks/auth_tokens.rake
+++ b/lib/tasks/auth_tokens.rake
@@ -16,7 +16,14 @@ namespace :auth_tokens do
         .authorized
         .find_in_batches(batch_size: args.batch_size, start: args.start).each do |token_batch|
           token_batch.each do |token|
-            result = token.still_authorized?
+            begin
+              result = token.still_authorized?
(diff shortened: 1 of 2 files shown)


lib/tasks/auth_tokens.rake
```diff
@@ -16,7 +16,14 @@ namespace :auth_tokens do
...
         .find_in_batches(batch_size: args.batch_size, start: args.start).each do |token_batch|
           token_batch.each do |token|
-            result = token.still_authorized?
+            begin
+              result = token.still_authorized?
+            rescue Octokit::TooManyRequests
+              # Tokens can still be authorized but may have exhausted their API limit while making
+              # these calls. For these tokens do not mark them as unauthorized but keep going
+              # through the loop.
+              next
```
jhan217, line 25: Is there a way to do a backoff or log every x times it happens so we know how bad the issue is still?

tiegz (approved):

PR #3334, opened 2024-03-20
NPM: change deprecation logic to only require latest non-prerelease to be deprecated

tiegz:
These changes tweak the behavior of NPM deprecation detection again:

####  pre-March 2023

we marked an NPM package as "Deprecated" if the latest-release had a "deprecation" message in the API.

#### March 2023

we [changed the logic](https://github.com/librariesio/libraries.io/pull/3053) to require *all releases* to have a "deprecation" message for the package to be marked as "Deprecated". This was for two reasons:
  * we were marking packages like [graphql](https://www.npmjs.com/package/graphql?activeTab=versions) as "Deprecated" because they publish frequent canary builds that are marked with a "deprecation" message. Whenever the latest release was a deprecated canary build and Libraries marked it "Deprecated", this gave the illusion that the package was deprecated even though it was very active.  
  * and also the `npm deprecate "This package is deprecated..."` command will deprecate all the releases for a package, implying that NPM's philosophy is that "all releases deprecated == the package is deprecated"

#### Mar 2024 

in this PR, we go back to the latest-release only for detection, but we ignore any version that's a prerelease, to handle the case of "graphql" where the latest version may be a deprecated prerelease. 

### Why?

despite the aforementioned `npm deprecate` logic, there's conflicting behavior on npmjs.com where a package page will be display as deprecated if the latest non-prerelease release is deprecated, even if the older releases aren't deprecated. For example, `@types/faker`:

https://www.npmjs.com/package/@types/faker
Screenshot 2024-03-19 at 5 39 49 PM
Screenshot 2024-03-19 at 5 39 57 PM

since there's not a first-class "deprecation" for packages in NPM, we have to pick a side, and users who visit NPM will think that this example means that the package is deprecated. If Libraries doesn't mark the package as deprecated, this will lead to more confusion.

### DIFF
--- a/app/models/package_manager/npm.rb
+++ b/app/models/package_manager/npm.rb
@@ -44,9 +44,21 @@ module PackageManager
     end
 
     def self.deprecation_info(db_project)
+      # NPM only allows versions to be deprecated (not the whole package), so we count a package deprecated
+      # on NPM if the latest version is deprecated (vs requiring all versions to be deprecated), because
+      # the package page will say it's deprecated if even the latest version is actually deprecated.
       versions = project(db_project.name)&.dig("versions")&.values || []
-      is_deprecated = versions.any? && versions.all? { |version| !version["deprecated"].nil? }
-      message = is_deprecated ? versions.last["deprecated"] : nil
+      last_stable_version = versions
+        # Ignore prerelease versions as some packages will regularly push prereleases and mark them deprecated, e.g. graphql.
+        .reject do |v|
+          Semantic::Version.new(v["version"]).pre.present?
+      rescue StandardError
+        false
+        end
+        .last
+
+      message = last_stable_version&.dig("deprecated")
+      is_deprecated = !message.nil?
 
       {
         is_deprecated: is_deprecated,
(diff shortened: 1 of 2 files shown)


app/models/package_manager/npm.rb
```diff
@@ -44,9 +44,21 @@ def self.project(name)
...
 
     def self.deprecation_info(db_project)
+      # NPM only allows versions to be deprecated (not the whole package), so we count a package deprecated
+      # on NPM if the latest version is deprecated (vs requiring all versions to be deprecated), because
+      # the package page will say it's deprecated if even the latest version is actually deprecated.
       versions = project(db_project.name)&.dig("versions")&.values || []
-      is_deprecated = versions.any? && versions.all? { |version| !version["deprecated"].nil? }
-      message = is_deprecated ? versions.last["deprecated"] : nil
+      last_stable_version = versions
+        # Ignore prerelease versions as some packages will regularly push prereleases and mark them deprecated, e.g. graphql.
```
macowie, line 52: Yeah, ignoring prereleases here is a good idea 👍

macowie (approved):
I looked at the original ticket I worked on for the reasoning we made that change, but knowing what I know now about deprecation, I do think this better reflects upstream state

mikeyoung85 (approved):

PR #3336, opened 2024-03-21
Set Repository Status Based on Upstream Status

mikeyoung85:
This slightly changes the logic in the Github `fetch_repo` logic to set the `status` property every time instead of just if the repository was found to be archived. This will allow Libraries to reset the Repository status if something is unarchived. I also moved some status setting logic out of the `Readme` class and moved it into `Repository` to make it more visible. I've tested the updated logic locally for the use cases I know about, but let me know if something looks off in here as I've had several different versions of refactors going at the same time.

 I think there is a much larger refactor needed to separate out the update logic between the RepositoryHost class and the Repository class that I am aiming to tackle in the next few days, so I expect that logic to move somewhere else into a discrete method/class responsible for setting statuses correctly based on upstream repository data.

### DIFF
--- a/app/models/repository.rb
+++ b/app/models/repository.rb
@@ -296,7 +296,10 @@ class Repository < ApplicationRecord
       download_manifests(token)
       update_source_rank(force: true)
     end
-    update(last_synced_at: Time.now)
+    update_unmaintained_statuses
+    self.last_synced_at = Time.current
+
+    save
   end
 
   def update_from_repository(token)
@@ -404,5 +407,20 @@ class Repository < ApplicationRecord
 
   def gather_maintenance_stats_async(priority: :medium)
     RepositoryMaintenanceStatWorker.enqueue(id, priority: priority)
+  end
+
+  def update_unmaintained_statuses
+    repository_unmaintained = unmaintained?
+    readme_unmaintained = readme.unmaintained?
+
+    if readme_unmaintained
+      update(status: "Unmaintained")
+      projects.update_all(status: "Unmaintained")
+    elsif repository_unmaintained
+      update(status: "Unmaintained")
+    else
+      update(status: nil)
+      projects.unmaintained.update_all(status: nil)
+    end
   end
 end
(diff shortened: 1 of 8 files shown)


app/models/repository_host/github.rb
```diff
@@ -62,7 +62,7 @@ def self.fetch_repo(id_or_name, token = nil)
       hash[:keywords] = hash[:topics]
       hash[:host_type] = "GitHub"
       hash[:scm] = "git"
-      hash[:status] = "Unmaintained" if hash[:archived]
+      hash[:status] = hash[:archived] ? "Unmaintained" : nil
```
mikeyoung85, line 65: One concern I have is this status could potentially overwrite other statuses outside of unmaintained. Another option I had in the works was to make a specific `archived` field and have `RepositoryHost` set the unmaintained status specifically based on the value of that field instead of doing what I have here now.

mikeyoung85: I think I'm going to go ahead and make this change so that this code doesn't overwrite existing statuses.

mikeyoung85:
I'll reopen after I work through how to set nil and Unmaintained statuses correctly

mikeyoung85:
I refactored things a little bit. I've changed the `RepositoryHost.update_from_host` method to read the `archived` field, which is supported by Github and [Gitlab](https://docs.gitlab.com/ee/api/projects.html#get-single-project) only, and use that to ask the `Repository` for the correct status to set. I've changed the moved Readme code to be a little more explicit about what it is doing.

app/models/repository.rb
```diff
@@ -405,4 +408,39 @@ def hide
...
   end
+
+  # Set the unmaintained status for this Repository and related Projects if the readme file
+  # indicates that the repository is not currently being maintained. This method will also
+  # removed the unmaintained status if neither the readme or repository indicate that it
+  # is not being maintained in the case where a project was picked up for work again.
+  def update_unmaintained_status_from_readme
+    if readme.unmaintained?
+      # update repository status only if needed
+      update(status: "Unmaintained") unless unmaintained?
```
tiegz, line 419: the `unless unmaintained?` check might be unnecessary because ActiveModel::Dirty will skip saving if no values have changed, but also can't harm to keep it here.

tiegz (approved):

PR #3338, opened 2024-03-25
Use a Class To Pass Around Repository Host Data

mikeyoung85:
This PR changes the RepositoryHost classes to map their upstream data to a defined Class and pass that Class around instead of unstructured Hash objects. This will help clear up what data is available in the different spots where we are sending upstream repository data around the app. I would like to follow up on this PR with some more changes and add some more util classes that would be responsible for assigning Repository column data to reduce the number of spots where that is happening right now and also split out some of the logic around managing duplicate Repositories that currently exists in the RepositoryHost and Repository classes.


app/models/repository_host/repository_host_data.rb
```diff
@@ -0,0 +1,89 @@
...
+    @name = name
+    @owner = owner
+    @parent = parent
+    @private = is_private
+    @repository_uuid = repository_uuid
+    @scm = scm
+    @size = size
+
+    raise "Unexpected arguments sent: #{kwargs.keys.join(', ')}" unless kwargs.keys.empty?
+  end
```
mpace965, line 55:
```suggestion
RepositoryHost::RepositoryHostData = Struct.new(
  :archived, :default_branch, :description, :fork, :full_name, :has_issues, :has_wiki, :homepage, :host_type,
  :keywords, :language, :license, :name, :owner, :parent, :is_private, :repository_uuid, :scm, :size,
  keyword_init: true
) do
```

This could be DRYed up by making `RepositoryHostData` a `Struct` with `keyword_init: true`.

That approach would mean you couldn't use the `**kwargs` parameter, but I'm not sure you need it. Could the factories pass the keyword arguments directly instead of creating a hash that is later splatted out? This also has the benefit of making key mismatch errors for callers catchable via static analysis, whereas this approach defers that error to runtime.

app/models/repository_host/repository_host_data.rb
```diff
@@ -0,0 +1,89 @@
...
+    @has_wiki = has_wiki
+    @homepage = homepage
+    @host_type = host_type
+    @keywords = keywords
+    @language = language
+    @license = license
+    @name = name
+    @owner = owner
+    @parent = parent
+    @private = is_private
```
mpace965, line 49:
```suggestion
    @is_private = is_private
```

This error would not be possible using a Struct

mpace965 (commented):
This pattern is a nice improvement, had one idea about the implementation

app/models/repository_host/repository_host_data.rb
mikeyoung85: I wasn't sure what direction this class was going to end up going but I think a `Struct` makes sense here, actually a `Data` would be better but we are stuck on an older Ruby version for now. The splat was me being lazy and copying code around, but I can switch it over to keywords directly now that I'm more confident on what keys are expected to exist. I'll make these changes today.

app/models/repository_host/base.rb
```diff
@@ -85,22 +85,19 @@ def repository_owner_class
...
-      r = self.class.fetch_repo(repository.id_or_name)
-      return unless r.present?
+      repo_data = self.class.fetch_repo(repository.id_or_name)
+      return unless repo_data.present?
 
-      repository.uuid = r[:id] unless repository.uuid.to_s == r[:id].to_s
-      if repository.full_name.downcase != r[:full_name].downcase
-        clash = Repository.host(r[:host_type]).where("lower(full_name) = ?", r[:full_name].downcase).first
+      if repository.full_name.downcase != repo_data.full_name.downcase
+        clash = Repository.host(repo_data.host_type).where("lower(full_name) = ?", repo_data.full_name.downcase).first
```
jhan217, line 92: 🎱 It looks like `data.full_name.downcase` is used a bunch - could that get abstracted out a `normalized_name` or something, and have the non-standard-name as the odd name out?

app/models/repository_host/repository_host_data.rb
```diff
@@ -0,0 +1,89 @@
+# frozen_string_literal: true
+
+# This class is meant to be a facade over the raw upstream data coming
+# from the different repository hosts' repository data. It's main goal
+# is to standardize the output from each repository host into a concrete
+# set of data so we can make sure the raw data is being mapped to the same
+# schema within the Libraries.io models and code.
+class RepositoryHost::RepositoryHostData
```
jhan217, line 8: 🎱 `class RepositoryHost::RawUpstreamData`

app/models/repository_host/repository_host_data.rb
jhan217: 🎱 I'm wondering if it makes sense to have the callers just pass in the raw data hash, and then have this object (or maybe a controller/interaction) figure out how to convert all the inputs into the appropriate data models to be persisted, vs having a bunch of factories/temp objects floating around. Mostly thinking that the factory classes aren't really doing a ton, or represent objects that will be reused a lot. 

I'm thinking something like: 

```
RawUpstreamDataToDatabaseModelConverter

def convert_gitlab_repository_raw_data(data_hash)
  repo_hash = {
      repository_uuid: data_hash.api_project.id,
      host_type: "GitLab",
      fork: data_hash.api_project.try(:forked_from_project).present?,
      scm: "git",
      is_private: api_project.visibility != "public",
    }

   create_the_thing(repo_hash)
end

def convert_github_repository_raw_data(data_hash);

def create_the_thing(data_hash_to_be_persisted)
      Repository.create!(repo_hash) # replace the Repository.create_from_hash
end

end

# some caller that gets the api data
github_api_data = Curl.get("https://www.github.com/foo/bar")

foo = RawUpstreamDataToDatabaseModelConverter.convert_gitlab_repository_raw_data(github_api_data) 
# you can play around with how much of that data should go into some sort of initializer in the converter, the caller, or the converter method

return foo
```

I think this has the advantage of keeping all the conversion logic in one place (easier to DRY up), no need to have intermediary temporary/facade objects, and a clean & consistent interface/pattern for the upstream to use (e.g. with error handling on bad input data). The single responsibility here could be something like "takes in garbage upstream stuff and turns it into our own objects."

app/models/repository.rb
```diff
@@ -310,23 +310,21 @@ def self.create_from_host(host_type, full_name, token = nil)
     RepositoryHost.const_get(host_type.capitalize).create(full_name, token)
   end
 
-  def self.create_from_hash(repo_hash)
-    return unless repo_hash
-
-    repo_hash = repo_hash.to_hash.with_indifferent_access
+  def self.create_from_hash(repo_host_data)
```
tiegz, line 313: could rename this to something like `create_from_data`

app/models/repository_host/repository_host_data.rb
tiegz: I like the single-factory idea from @jhan217 , additionally because it makes it easy to see the differences among all the sources in one place too.

mikeyoung85:
* changed from class to Struct with some appended method
* moved the different "factories" into a single converter
* updated method name to `create_from_data` and updated some callers I didn't see before
* added convenience method for `full_name.downcase`

tiegz (approved):

spec/models/repository_host/gitlab_spec.rb
```diff
@@ -3,6 +3,14 @@
...
 describe RepositoryHost::Gitlab do
-  let(:repository) { build(:repository, host_type: "GitLab") }
+  let(:repository) { build(:repository, host_type: "GitLab", full_name: "ase/ase") }
   let(:repository_host) { described_class.new(repository) }
+  let(:api_token) { "TEST_TOKEN" }
+
+  it "can fetch repository data" do
+    VCR.use_cassette("gitlab/ase") do
+      repository_data = described_class.fetch_repo(repository.id_or_name, api_token)
+      expect(repository_data).not_to be_nil
```
jhan217, line 13: Should these spex do more? e.g. expect the repository_data object to have successfully extracted specific elements / keys with the correct values, and that it's not just getting all the keys and populating them with nils or data from the wrong fields?

app/models/concerns/github_identity.rb
```diff
@@ -43,7 +43,7 @@ def update_repo_permissions
 
     update_column(:currently_syncing, true)
     download_orgs
-    r = github_client.repos
+    r = github_client.repos.map { |repo_data| GithubRepositoryHostDataFactory.generate_from_api(repo_data) }
```
jhan217, line 46: Think this factory got removed in [this commit](https://github.com/librariesio/libraries.io/pull/3338/commits/5f85a015837df5237fb234d7304d3cbb5848e6dc#diff-114d130d1fe1bd53c12fb4eafaeaa8d3fde39abaae7752b13c2f47cde4a2e170)...?

app/models/concerns/github_identity.rb
```diff
@@ -53,7 +53,7 @@ def update_repo_permissions
 
     r.each do |repo|
       unless (github_repo = existing_repos.find { |re| re.uuid.to_s == repo.id.to_s })
-        github_repo = Repository.host("GitHub").find_by("lower(full_name) = ?", repo.full_name.downcase) || Repository.create_from_hash(repo)
+        github_repo = Repository.host("GitHub").find_by("lower(full_name) = ?", repo.full_name.downcase) || Repository.create_from_data(repo)
```
jhan217, line 56: `repo.lower_name`

app/models/repository_host/raw_upstream_data_converter.rb
```diff
@@ -0,0 +1,79 @@
+# frozen_string_literal: true
+
+class RepositoryHost::RawUpstreamDataConverter
+  def self.convert_from_github_api(upstream_repository_data_hash)
```
jhan217, line 4: Should the error handling get standardized in here too? Saw a few `rescue *self::IGNORABLE_EXCEPTIONS nil` in the code. 

Alternatively, now that this is written out, it looks like there's some pretty tight coupling w/ the api return objects, so the convert_ methods could be moved to their respective methods in the repository_host/<host>.rb files, as there doesn't seem to be a ton that's similar between the convert methods besides the object that's tying everything together, unless stuff like error handling is getting standardized. I was thinking there'd be more similarities between the objects being converted that there could be more extracted out and shared during the conversion process, like defaulting values/data validations/logging/etc.

PR #3343, opened 2024-03-27
Fix Error in User Permission Sync

mikeyoung85:
Fix name error in code and handle repository data more correctly when syncing user repository permission data.


spec/workers/sync_permissions_worker_spec.rb
```diff
@@ -7,10 +7,39 @@
...
+    it "should sync user permissions and repositories" do
+      VCR.use_cassette("github/user_sync") do
+        user = create(:user)
+        user.identities.update_all(token: "TEST_TOKEN")
+
+        expect(user.repository_permissions.count).to be 0
+        expect(Repository.count).to be 0
+        expect(RepositoryDownloadWorker.jobs.size).to be 0
+
+        expect(User).to receive(:find_by_id).with(user.id).and_return(user)
```
jhan217, line 31: 🎱 Can this be changed to the `have_received` expectations? I don't know the deets but iirc the `expect to receive` doesn't guarantee that the thing you want to have happen actually happens, probably similar to the sql `explain` vs `explain analyze` where one of them is the predicted path vs actually running the query to get the actual path.

spec/workers/sync_permissions_worker_spec.rb
```diff
@@ -7,10 +7,39 @@
...
+        subject.perform(user.id)
+
+        # expect permissions were added
+        expect(user.repository_permissions.count).to be 109
+        # expect repositories for those permissions to have been added
+        expect(Repository.count).to be 109
+        # expect work to be queued up to sync data for those repositories
+        expect(RepositoryDownloadWorker.jobs.size).to be 218
```
jhan217, line 39: Do these actually get queued up/run in the test env or has all that been stubbed?

mikeyoung85: They are only queued, not run.

jhan217 (approved):

PR #3345, opened 2024-03-29
Move Repository create_or_update Logic to a Domain Class

mikeyoung85:
I wanted to move this logic out of the `Repository` and `RepositoryHost` classes. It felt complicated enough where it warranted being separated so it could be easier to test. There is also some dependency in between Repository and RepositoryHost that made it hard to follow the code path and I thought separating this out would be helpful in reducing that complexity. The interface to this logic still comes through the same methods and parameters, but a further refactor could be changing how callers to these functions are interacting with this create or update logic. There were no specs checking this section of the code so I have added some based on how I read the code but if it looks incorrect please call that out. Let me know if you think this refactor is worth the possible troubles that come with touching this portion of the code base.


app/domain/repository/persist_repository_from_upstream.rb
```diff
@@ -0,0 +1,129 @@
+# frozen_string_literal: true
+
+class Repository::PersistRepositoryFromUpstream
+  def self.create_or_update_from_host_data(upstream_raw_host_data)
+    return unless upstream_raw_host_data
+
+    ActiveRecord::Base.transaction do
```
jhan217, line 7: Any particular reason to put all the finds & assignments in a db transaction?

app/domain/repository/persist_repository_from_upstream.rb
```diff
@@ -0,0 +1,129 @@
...
+      existing_repo = find_repository_from_host_data(upstream_raw_host_data)
+
+      if existing_repo.nil?
+        new_repository = Repository.new(uuid: upstream_raw_host_data.repository_uuid, full_name: upstream_raw_host_data.full_name)
+
+        new_repository.host_type = upstream_raw_host_data.host_type
+
+        new_repository.license = upstream_raw_host_data.formatted_license if upstream_raw_host_data.formatted_license
```
jhan217, line 15: 🎱 Are the postfix ifs needed?

app/domain/repository/persist_repository_from_upstream.rb
```diff
@@ -0,0 +1,129 @@
...
+  def self.create_or_update_from_host_data(upstream_raw_host_data)
+    return unless upstream_raw_host_data
+
+    ActiveRecord::Base.transaction do
+      existing_repo = find_repository_from_host_data(upstream_raw_host_data)
+
+      if existing_repo.nil?
+        new_repository = Repository.new(uuid: upstream_raw_host_data.repository_uuid, full_name: upstream_raw_host_data.full_name)
+
+        new_repository.host_type = upstream_raw_host_data.host_type
```
jhan217, line 13: 🎱 Can these two methods get consolidated?

```
def create_or_update
  repo = find_repo || repo.new(..)

  repo.update(
     host_type: upstream_raw_host_data.host_type
     license: upstream_raw_host_data.formatted_license
  )
end
```

app/domain/repository/persist_repository_from_upstream.rb
```diff
@@ -0,0 +1,129 @@
...
+
+    if repository.changed?
+      repository.save ? repository : nil
+    else
+      repository
+    end
+  end
+
+  def self.find_repository_from_host_data(upstream_raw_host_data)
+    existing_repo = Repository.where(host_type: (upstream_raw_host_data.host_type || "GitHub")).find_by(uuid: upstream_raw_host_data.repository_uuid)
```
jhan217, line 50:
```   
    Repository.find_by(
       host_type: (upstream_raw_host_data.host_type || "GitHub"),
       uuid: upstream_raw_host_data.repository_uuid,
    ) || Repository.host(
       upstream_raw_host_data.host_type || "GitHub"
    ).find_by("lower(full_name) = ?", upstream_raw_host_data.lower_name)
```

app/domain/repository/persist_repository_from_upstream.rb
```diff
@@ -0,0 +1,129 @@
...
+    existing_repo = Repository.host(upstream_raw_host_data.host_type || "GitHub").find_by("lower(full_name) = ?", upstream_raw_host_data.lower_name) if existing_repo.nil?
+
+    existing_repo
+  end
+
+  def self.remove_repository_name_clash(host_type, full_name)
+    # TODO: work on refactoring this out of this class and handle it in another process
+    clash = Repository.host(host_type).where("lower(full_name) = ?", full_name.downcase).first
```
jhan217, line 58: Should this remove anything that comes up in the where clause, instead of just the first one that is found?

app/domain/repository/persist_repository_from_upstream.rb
```diff
@@ -0,0 +1,129 @@
...
+
+    if clash && (!clash.update_from_repository || clash.status == "Removed")
+      StructuredLog.capture("REMOVE_REPOSITORY_BY_NAME_CLASH",
+                            {
+                              repository_full_name: clash.full_name,
```
jhan217, line 63: Should this have all of the repository attributes in case we need to restore it?

app/domain/repository/persist_repository_from_upstream.rb
```diff
@@ -0,0 +1,129 @@
...
+    attrs
+  end
+
+  # Suggest the correct Repository status based on if the upstream repository data
+  # indicates it should be and the current status set on this Repository.
+  def self.correct_status_from_upstream(repository, archived_upstream:)
+    if archived_upstream && repository.status.nil?
+      StructuredLog.capture("REPOSITORY_SET_UNMAINTAINED_STATUS",
+                            {
+                              repository_host: repository.host_type,
```
jhan217, line 111: + repository.id here and below?

spec/domain/repository/persist_repository_from_upstream_spec.rb
```diff
@@ -0,0 +1,243 @@
...
+      expected_attributes = described_class.raw_data_repository_attrs(raw_upstream_data)
+
+      expect(Repository.first.attributes.symbolize_keys).to include(**expected_attributes)
+    end
+
+    context "with existing repository" do
+      let!(:existing_repository) { create(:repository, full_name: raw_upstream_data.full_name) }
+
+      it "does not create a new Repository" do
+        expect { described_class.create_or_update_from_host_data(raw_upstream_data) }.not_to change(Repository, :count)
```
jhan217, line 25: This should probably check that the ID is also the same, otherwise the method could delete the existing one and replace it with a new one.

spec/domain/repository/persist_repository_from_upstream_spec.rb
```diff
@@ -0,0 +1,243 @@
...
+      end
+
+      it "updates existing repository" do
+        described_class.create_or_update_from_host_data(raw_upstream_data)
+
+        expected_attributes = described_class.raw_data_repository_attrs(raw_upstream_data)
+
+        # TODO: figure out how to check subset of attributes correctly
+        expect(existing_repository.reload.attributes.symbolize_keys).to include(**expected_attributes)
```
jhan217, line 34: 🎱 Could get a clean copy from the db directly to check? `existing_repository.reload` -> `Repository.find_by(....).attributes`. Or you could stuff all the expected attrs into the find_by and check that the result isn't nil.

spec/domain/repository/persist_repository_from_upstream_spec.rb
```diff
@@ -0,0 +1,243 @@
...
+  describe "#remove_repository_name_clash" do
+    let(:full_name) { "test/clash" }
+    let(:existing_repository) { create(:repository, full_name: full_name) }
+
+    before do
+      allow_any_instance_of(Repository).to receive(:update_from_repository).and_return(true)
+    end
+
+    it "doesn't destroy clash on name" do
+      # TODO: figure out what is supposed to happen in this case
```
jhan217, line 62: Will this be address in this PR or in a future one?

spec/factories.rb
```diff
@@ -246,4 +246,36 @@
...
   end
+
+  factory :raw_upstream_data, class: RepositoryHost::RawUpstreamData do
+    archived { true }
+    default_branch { "main" }
+    description { "description" }
+    fork { true }
+    fork_policy { "" }
+    forks_count { 10 }
+    full_name { "test/repo" }
```
jhan217, line 257: 🎱 Would be nice to sequence this / some other traits so `create_list` can create non-identical-except-for-id objects.

app/domain/repository/persist_repository_from_upstream.rb
tiegz: yeah, I'm curious if the transaction there was unnecessary -- there's only one INSERT or UPDATE happening here, and in the INSERT case it doesn't raise a RecordInvalid when it's not valid anyway, so it wouldn't rollback the transaction

app/domain/repository/persist_repository_from_upstream.rb
tiegz: yeah, seems like the ones above and below aren't necessary, if it's a new record

app/domain/repository/persist_repository_from_upstream.rb
mikeyoung85: I initially kept this from the old code because the code read like it was uncertain of the existing data. With that in mind it made sense to make sure the find and update existed in the same transaction so that a name clash was found and updated/destroyed correctly in tandem with the targeted create/update here. That actually may not work that way with how transactions function. I'm wondering if removing the name clash logic out of here and doing an optimist update to the data instead would uncomplicate some of this.

app/domain/repository/persist_repository_from_upstream.rb
mikeyoung85: I kept them separate because that is how it was in the original code. Putting them in one implies `find_repo` would always find the right one instead of having the target repository passed in directly. Right now the only time we have a specific `Repository` being passed in is via `Repository.update_from_repository()` which is called in `Repository.update_all_info()` and in the name clash logic in this class. Maybe I need to try and run some queries to figure out if we do have name clashes in the database that need to be handled. It would be nice to implement your suggestion here.

app/domain/repository/persist_repository_from_upstream.rb
```diff
@@ -0,0 +1,126 @@
...
+    existing_repo = Repository.host(upstream_raw_host_data.host_type || "GitHub").find_by("lower(full_name) = ?", upstream_raw_host_data.lower_name) if existing_repo.nil?
+
+    existing_repo
+  end
+
+  def self.remove_repository_name_clash(host_type, full_name)
+    # TODO: work on refactoring this out of this class and handle it in another process
+    clash = Repository.host(host_type).where("lower(full_name) = ?", full_name.downcase).first
+
+    if clash && (!clash.update_from_repository || clash.status == "Removed")
```
johnbintz-tidelift, line 55: This took a little while for me to parse. WDYT about something without the nested conditionals?

```suggestion
    return unless clash
    return unless clash.update_from_repository
    return unless clash.status == "Removed"
```

mikeyoung85: I don't think the two returns here capture the scenario correctly. We are not deleting the clashing Repository if we still get info from the Repository Host and the host is not marking it as Removed. I think I still need an `or` in here.

`return if clash.update_from_repository && clash.status != "Removed"` else delete it.

app/domain/repository/persist_repository_from_upstream.rb
jhan217: Discussed a bit offline and adding here - it looks like the clashing behavior was introduced around [here](https://github.com/librariesio/libraries.io/commit/5d2b4968590f63649392b488dff87bd5abd3b698#diff-efd45bdfac24905df71fad0af0d27a758f406f7dca1e151f2568bc0962bd9383L201-R207). It seems like the intention was to keep the repository in sync with what the upstream's knowledge of the repository is, based off the id. So if libraries has an existing repository with id: 111/name: foo, then the next request returns with id: 111/name: bar, then we'd delete the existing row for foo and replace it with new row bar. This would ensure that the entry mirrors what's in the upstream, vs an update that could potentially end up with a mix of attributes from the old and new repository data. 

I'm not really sure what the actual use case is in github/gitlab/bitbucket is/was though with same ids/changed names that also changes entire rows (maybe if the main branch gets force pushed with the contents of a totally different repo/fork of a different repo?), but that appears to be the intention AFAICT. 

Also to note from the brief archaeology splunking, it looks like the history of collecting the different repository datas:
1) started off as an async sidekiq job to poll the upstream every once in a while
2) moved into a centralized Repository.rb model
3) moved into individual concerns when there got to be too much if this_repo/elsif that_repo branching in all the methods
4) moved into where they are now as separate classes

spec/domain/repository/persist_repository_from_upstream_spec.rb
```diff
@@ -47,41 +47,74 @@
...
+      let(:raw_upstream_data) { build(:raw_upstream_data, full_name: full_name) }
+      let(:existing_repository) { create(:repository, full_name: full_name) }
+      let(:incoming_repository) { build(:repository) }
+
+      context "upstream update fails" do
+        before do
+          allow(RepositoryHost::Github).to receive(:fetch_repo).with(existing_repository.id_or_name, nil).and_return(nil)
+        end
+
+        it "deletes clashed name" do
```
jhan217, line 62: Should there also be an expectation that not only is the thing deleted, but the new thing was created in its place? Seems like those two operations should be atomic/put into a transaction, so that we don't end up in a partial state of "well we deleted the clash but didn't replace it with anything," though if it's preexisting behavior...

spec/domain/repository/persist_repository_from_upstream_spec.rb
```diff
@@ -47,41 +47,74 @@
...
+          described_class.update_from_host_data(incoming_repository, raw_upstream_data)
+          expect(Repository.find_by(id: existing_repository.id)).to be_nil
```
jhan217, line 64: 🎱 Could be worth putting in a changed by -1 matcher, though that's testing the setup a bit (making sure the setup also created the object & got deleted, as opposed to the object never existing in the first place)

jhan217 (approved):

Issue #3283, opened 2023-12-28
Missing `x-ratelimit-limit` and `x-ratelimit-remaining` headers


chris48s:
Libraries.io enforces a 60 requests per minute rate limit: https://libraries.io/api

API requests used to return `x-ratelimit-limit` and `x-ratelimit-remaining` headers in the response.

API responses no longer contain these headers. This seems to have changed around the 15th or 16th December.

This makes it very difficult to stay under the rate limit if you have more than one client sharing the same token.

Would it be possible to restore these headers?

Thanks

chris48s:
Just to add clarification:

- This is on the `200 OK` responses where the headers are missing
- I think this is the commit where this functionality was removed https://github.com/librariesio/libraries.io/commit/6c400b52305ec51636eec00f5b7de5b23b7aefa4
  Looks like as of that commit, the `x-ratelimit-*` headers are now _only_ present on a 429 response

PR #3351, opened 2024-04-05
Re-add rate-limit headers back to non-rate-limited responses.

tiegz:
before https://github.com/librariesio/libraries.io/pull/3277 , we served `X-RateLimit-Limit` and `X-RateLimit-Remaining` headers in non-rate-limited responses to transmit info about rate-limiting to clients.

after https://github.com/librariesio/libraries.io/pull/3277 , those successful response headers were removed, and we added `RateLimit-Limit`, `RateLimit-Remaining`, and `RateLimit-Reset` headers to non-rate-limited responses, so unfortunately clients could no longer track their usage in successful responses.

these changes bring back the former non-rate-limited headers, and also prefix the rate-limited headers with `X-` since those headers are not quite a standard yet, at time of writing: https://www.ietf.org/archive/id/draft-ietf-httpapi-ratelimit-headers-07.html .
 
### Example:

```
# Non-rate-limited API response
HTTP/1.1 200 OK
...
X-RateLimit-Limit: 60
X-RateLimit-Remaining: 3
X-RateLimit-Reset: 57
...

# Rate-limited API response
HTTP/1.1 429 Too Many Requests
...
X-RateLimit-Limit: 60
X-RateLimit-Remaining: 0
X-RateLimit-Reset: 54
Retry-After: 54
...
```

NOTE: I've only added the non-rate-limited headers to API responses. Do we want to include them for Web responses too?

closes https://github.com/librariesio/libraries.io/issues/3283

### DIFF
--- a/config/initializers/rack_attack.rb
+++ b/config/initializers/rack_attack.rb
@@ -46,13 +46,22 @@ class Rack::Attack
   # Adds RateLimit-X headers to throttled responses
   self.throttled_responder = lambda do |request|
     match_data = request.env["rack.attack.match_data"]
-    now = match_data[:epoch_time]
+    match_data[:epoch_time]
 
-    headers = {
-      "RateLimit-Limit" => match_data[:limit].to_s,
-      "RateLimit-Remaining" => "0",
-      "RateLimit-Reset" => (now + (match_data[:period] - (now % match_data[:period]))).to_s,
-    }
+    # Include the equivalent "X-RateLimit-Reset" and "Retry-After" as seconds left before reset,
+    # for clients that might look for either one. Docs:
+    #
+    #   * "A response that includes the RateLimit-Limit header field MUST also include the RateLimit-Reset."
+    #     (from the IETF draft https://www.ietf.org/archive/id/draft-ietf-httpapi-ratelimit-headers-07.html#section-4)
+    #
+    #   * "Retry-After" is a multi-purpose header that means the same thing
+    #     (fropm RFC9110 https://www.rfc-editor.org/rfc/rfc9110#field.retry-after)
+    retry_after = (throttle_data[:limit] - throttle_data[:count]).to_s
+
+    headers["X-RateLimit-Limit"]     = throttle_data[:limit].to_s
+    headers["X-RateLimit-Remaining"] = "0"
+    headers["X-RateLimit-Reset"]     = retry_after
+    headers["Retry-After"]           = retry_after
 
     [429, headers, ["Retry later\n"]]
   end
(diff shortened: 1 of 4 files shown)


config/initializers/rack_attack.rb
```diff
@@ -45,13 +45,24 @@ def remote_ip
...
+    # Include the equivalent "X-RateLimit-Reset" and "Retry-After" as seconds left before reset,
+    # for clients that might look for either one. Docs:
+    #
+    #   * "A response that includes the RateLimit-Limit header field MUST also include the RateLimit-Reset."
+    #     (from the IETF draft https://www.ietf.org/archive/id/draft-ietf-httpapi-ratelimit-headers-07.html#section-4)
+    #
+    #   * "Retry-After" is a multi-purpose header that means the same thing
+    #     (fropm RFC9110 https://www.rfc-editor.org/rfc/rfc9110#field.retry-after)
+    now = throttle_data[:epoch_time]
+    retry_after = (throttle_data[:period] - (now % throttle_data[:period])).to_s
```
tiegz, line 59: just a note that this is the logic Rack::Attack uses to compute time periods

config/initializers/rack_attack.rb
```diff
@@ -45,13 +45,24 @@ def remote_ip
...
 
     headers = {
-      "RateLimit-Limit" => match_data[:limit].to_s,
-      "RateLimit-Remaining" => "0",
-      "RateLimit-Reset" => (now + (match_data[:period] - (now % match_data[:period]))).to_s,
+      "X-RateLimit-Limit" => throttle_data[:limit].to_s,
+      "X-RateLimit-Remaining" => "0",
+      "X-RateLimit-Reset" => retry_after,
+      "Retry-After" => retry_after,
```
tiegz, line 65: the headers without `X-` were only in the wild for ~4 months, so if anyone complains we can let them know it was on purpose.

lib/subscribers/rack_attack.rb
```diff
@@ -14,7 +14,7 @@
 
   Rails.logger.info "[RACK_ATTACK] " \
                     "method=#{req.request_method} " \
-                    "path=#{req.fullpath} " \
+                    "path=#{req.path} " \
```
tiegz, line 17: unrelated, we just don't need the querystring in these logs

pranav (approved):
i don't have any strong opinion on including rate limit headers on web requests

tradiff (approved):
Thanks!  cc @chris48s

chris48s:
thanks for getting this sorted :+1:

PR #3352, opened 2024-04-05
Search: allow opting out of versions

tradiff:
For performance gains / payload reduction, allow the querystring param `?include_versions=false`. Defaults to `true`


macowie (approved):
Looks good to me

PR #3353, opened 2024-04-05
Project#check_status: don't override status if it's manually set.

tiegz:
if admins set a project's status to `"Hidden"`, we don't want to override that decision whenever the project's status is checked regularly.

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -640,7 +640,7 @@ class Project < ApplicationRecord
     return if url.blank?
 
     # "Hidden" is a state set by admins, and we don't want to override that decision.
-    return if status == "Hidden"
+    return if hidden?
 
     response = Typhoeus.get(url)
     if platform.downcase == "packagist" && [302, 404].include?(response.response_code)
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -639,6 +639,9 @@ class Project < ApplicationRecord
     update_column(:status_checked_at, DateTime.current)
     return if url.blank?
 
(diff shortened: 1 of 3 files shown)


spec/models/project_spec.rb
```diff
@@ -300,6 +300,22 @@
         end
       end
     end
+
+    context "a hidden project that is active" do
+      let!(:project) { Project.create(platform: "NPM", name: "react", status: "Hidden") }
```
mikeyoung85, line 305:
```suggestion
      let!(:project) { create(:project, :npm, name: "react", status: "Hidden") }
```
Why not use the factory here?

app/models/project.rb
```diff
@@ -639,6 +639,9 @@ def check_status
     update_column(:status_checked_at, DateTime.current)
     return if url.blank?
 
+    # "Hidden" is a state set by admins, and we don't want to override that decision.
+    return if status == "Hidden"
```
mikeyoung85, line 643:
```suggestion
    return if hidden?
```
Could use this convenience method from the Concern: https://github.com/librariesio/libraries.io/blob/c66742fbc77f20ebdc310e887e18a95ad9b6db11/app/models/concerns/status.rb#L20-L22

mikeyoung85 (commented):
We could also update the rake task to not call `check_status` on hidden Projects to save some effort on the cronjob.

https://github.com/librariesio/libraries.io/blob/c66742fbc77f20ebdc310e887e18a95ad9b6db11/lib/tasks/projects.rake#L72-L77

mikeyoung85 (approved):

PR #3357, opened 2024-04-10
Add Project#latest_version assocation, keep it up-to-date on save, and backfill it.

tiegz:
(followup to https://github.com/librariesio/libraries.io/pull/3356)

* a backfill migration to populate `Projects#latest_version_id` (this has already been run on production)
* replace `Project#latest_version` method with a new `:latest_version` belongs_to association
* add a `set_latest_version` method that updates that column after a `Project` is saved

### DIFF
--- a/db/migrate/20240409222653_backfill_projects_latest_version_id.rb
+++ b/db/migrate/20240409222653_backfill_projects_latest_version_id.rb
@@ -1,8 +1,11 @@
+# frozen_string_literal: true
+
 class BackfillProjectsLatestVersionId < ActiveRecord::Migration[7.0]
   def up
     # This migration has already been run manually in production.
     return if Rails.env.production?
 
+    say "Backfilling #{Project.count} project 'latest_version_id' values. Each '.' = 10k records completed."
     Project
       .in_batches(of: 10_000) do |projects|
         sql = <<~SQL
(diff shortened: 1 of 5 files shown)


db/migrate/20240409222653_backfill_projects_latest_version_id.rb
```diff
@@ -0,0 +1,37 @@
...
+
+    Project
+      .in_batches(of: 10_000) do |projects|
+        sql = <<~SQL
+          WITH rows_to_update AS (
+            SELECT p.id as project_id, latest_version.id as latest_version_id
+            FROM projects p
+            LEFT JOIN LATERAL (
+              SELECT versions.*
+              FROM versions#{' '}
```
jhan217, line 16: What are the `#{' '}`s doing?

db/migrate/20240409222653_backfill_projects_latest_version_id.rb
```diff
@@ -0,0 +1,37 @@
...
+    # This migration has already been run manually in production.
+    return if Rails.env.production?
+
+    Project
+      .in_batches(of: 10_000) do |projects|
+        sql = <<~SQL
+          WITH rows_to_update AS (
+            SELECT p.id as project_id, latest_version.id as latest_version_id
+            FROM projects p
+            LEFT JOIN LATERAL (
```
jhan217, line 14: 🎱 Another way to do this would be to use a `rank() over partition by` clause to get the latest version per project. 

I think in either case it would be good to have a spec to test out this functionality to make sure it's not like setting all 10000 projects' latest_version_id to the one same version row.

db/migrate/20240409222653_backfill_projects_latest_version_id.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+class BackfillProjectsLatestVersionId < ActiveRecord::Migration[7.0]
+  def up
+    # This migration has already been run manually in production.
+    return if Rails.env.production?
+
+    Project
+      .in_batches(of: 10_000) do |projects|
+        sql = <<~SQL
```
jhan217, line 10: 🎱 I wonder if will need to get/set this for multiple projects at a time in biz logic, in which case it might be useful to have this in a class method or interaction instead of just in a backfill. Then it can also be tested via the usual routes.

jhan217 (commented):
How long is the backfill expected to take / will it impact deploys going out if it takes a long time or fails for some reason? If it will, could be worth making it a rake task and manually running it on a console.

db/migrate/20240409222653_backfill_projects_latest_version_id.rb
tiegz: ah, rubocop strikes again!

tiegz:
> How long is the backfill expected to take / will it impact deploys going out if it takes a long time or fails for some reason? If it will, could be worth making it a rake task and manually running it on a console.

I'd mentioned in the PR description that I've already run the migration on production, and I've double-checked the data for each row too to make sure the migration was correct.

mikeyoung85 (approved):

jhan217 (approved):

PR #3358, opened 2024-04-10
Treat Redirected Go Packages as Removed

mikeyoung85:
This PR updates the `check_status` logic for Go packages to treat a redirect from pkg.go.dev as a "Removed" package. The reasons for this change are to help capture when packages are renamed or recased and mark the now incorrect one appropriately. This is just an idea, so feel free to disagree with this approach. 

One example is:
https://libraries.io/go/github.com%2FAzure%2Fazure-pipeline-go vs https://libraries.io/go/github.com%2Fazure%2FAzure-pipeline-go which both reference `github.com/Azure/azure-pipeline-go`. This PR would mark `github.com/azure/Azure-pipeline-go` as removed.

```shell
> url = "https://pkg.go.dev/github.com/azure/Azure-pipeline-go"
> response = Typhoeus.get(url)
> response.response_code
=> 302
> response.body
=> "<a href=\"/github.com/Azure/azure-pipeline-go\">Found</a>.\n\n"
```

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -645,7 +645,7 @@ class Project < ApplicationRecord
     response = Typhoeus.get(url)
     if platform.downcase == "packagist" && [302, 404].include?(response.response_code)
       update_attribute(:status, "Removed")
-    elsif platform.downcase == "go" && [400, 404].include?(response.response_code)
+    elsif platform.downcase == "go" && [302, 400, 404].include?(response.response_code)
       # pkg.go.dev can be 404 on first-hit for a new package (or alias for the package), so ensure that the package existed in the past
       # by ensuring its age is old enough to not be just uncached by pkg.go.dev yet.
       update_attribute(:status, "Removed") if created_at < 1.week.ago
--- a/spec/models/project_spec.rb
+++ b/spec/models/project_spec.rb
@@ -210,21 +210,23 @@ describe Project, type: :model do
     end
 
     context "a go project missing from upstream" do
+      let(:check_status_url) { PackageManager::Go.check_status_url(project) }
+
       context "recently created" do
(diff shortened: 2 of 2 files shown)


app/models/project.rb
```diff
@@ -645,7 +645,7 @@ def check_status
     response = Typhoeus.get(url)
     if platform.downcase == "packagist" && [302, 404].include?(response.response_code)
       update_attribute(:status, "Removed")
-    elsif platform.downcase == "go" && [400, 404].include?(response.response_code)
+    elsif platform.downcase == "go" && [302, 400, 404].include?(response.response_code)
```
jhan217, line 648: Based off the logic in the description, should all packages that redirect be marked as "Removed"?

Also, maybe not for this PR, but thoughts about moving the platform specific statusing logic into their own classes vs a giant if/elsif/else statement?

spec/models/project_spec.rb
```diff
@@ -213,27 +213,45 @@
       context "recently created" do
         let!(:project) { create(:project, platform: "Go", name: "github.com/some-nonexistent-fake/pkg", status: nil, created_at: 1.hour.ago) }
 
-        it "should not mark it as Removed" do
+        it "should not mark it as Removed for a 404" do
```
jhan217, line 216: 🎱 Thoughts about cleaning up the specs to standardize the mocking / separate out the setup from the test expectations?

jhan217 (approved):

tiegz (approved):
this change seems fine, although I don't think it'll cover 100% of badly cased packages. For instance here's a valid Go Module (has a go.mod) where I've entered the wrong casing on pkg.go.dev, but pkg.go.dev still serves up the package without a redirect: 

https://pkg.go.dev/github.com/go-sql-driver/mysQL

(but maybe we've fixed the Go ingestor far enough that it'll use the proper name `github.com/go-sql-driver/mysql` instead when ingesting?)

PR #3361, opened 2024-04-11
Ask for Canonical Name when Mapping Go Project

mikeyoung85:
Thoughts on adding a call to get the "canonical" name from Go proxy during the project name mapping for Go projects? This could cut down on casing problems creating multiple Projects, but only for valid Go module projects. There is a noticeable lag when making this request and the proxy does not have a cached response so I think that is the main tradeoff here.

I think the proxy protocol specifies that for projects that do not use `go.mod` files it will just create a virtual `go.mod` with the original name passed back, which I think means this is still a problem for any name that comes in for a non module package.

image


app/models/package_manager/go.rb
```diff
@@ -87,17 +87,21 @@ def self.one_version(raw_project, version_string)
     end
 
     def self.project(name)
+      # try and find the canonical name from the Go proxy and use that if we find it
```
tiegz, line 90: might be worth mentioning that the reasoning is because pkg.go.dev has been known to store the wrong casings of a Go Module name, eg https://pkg.go.dev/github.com/go-sql-driver/mysQL

app/models/package_manager/go.rb
```diff
@@ -161,15 +165,15 @@ def self.mapping(raw_project)
...
 
             # if we didn't find one then try and get the base project
             unless existing_project_name.present? # rubocop: disable Metrics/BlockNesting
-              versioned_name = Project.where(platform: "Go").where("lower(repository_url) = ? and name not like '%/v'", url.downcase).first&.name
+              versioned_name = Project.visible.where(platform: "Go").where("lower(repository_url) = ? and name not like '%/v'", url.downcase).first&.name
               existing_project_name = versioned_name&.concat("/#{versioned_module_regex[2]}")
             end
           else
-            existing_project_name = Project.where(platform: "Go").where("lower(name) = ?", raw_project[:name].downcase).first&.name
+            existing_project_name = Project.visible.where(platform: "Go").lower_name(raw_project[:name].downcase).first&.name
```
tiegz, line 176: seems like we could assume by this point that we have the correct name now, so might not need the `lower()` anymore. WDYT?

mikeyoung85: I think we will still run into problems if the project is not a module project. The proxy will echo back whatever string you sent. We don't have too many duplicate names right now, a little less than 2k, so I might be able to figure out which one is "right" and remove the "wrong" one so this code would only ever find one and it would end up being the correct name. I could also look into figuring out if the proxy responded with the correct name and pass that as a boolean through to this where it can better determine how the name should be set.

mikeyoung85: After re-reading this section a bit, this name matching code is only for non module names that we get. Since those can't be considered canonical due to the way the proxy works we will still need to keep this matching. I've added a spec to capture that use case. I've also checked duplicate names again and cleaned up that data.

tiegz (approved):

app/workers/go_project_verification_worker.rb
```diff
@@ -14,8 +14,17 @@ def perform(name)
     # check to see if the module is found on pkg.go.dev and if it isn't then go ahead and delete this name
     return project.destroy unless PackageManager::Go.valid_project?(project.name)
 
-    # if this name is found and is considered a module then that should be considered canonical
-    unless PackageManager::Go.module?(project.name)
+    # if this name is found and is considered a module then we can determine the canonical name
+    if PackageManager::Go.module?(project.name)
+      canonical_name = PackageManager::Go.canonical_module_name(project.name)
```
tiegz, line 19: if `module?` returns true, doesn't that mean that we have the canonical name by this point? For instance, this would return `false` because the HTML says it doesn't have a valid go.mod: https://pkg.go.dev/github.com/go-sql-driver/mysQL

mikeyoung85: Going forward yes, but we have a bunch of duplicates for names that should be canonical right now. This is updating the worker to remove the non canonical name since we for sure know it's incorrect now.

tiegz (approved):

PR #3363, opened 2024-04-15
MavenCentral: if latest_version returns ${}, scrape the HTML for the package to find the latest_version instead.

tiegz:
when we update Maven packages, the `PackageManager::Maven.project()` method looks for the latest version of the package in order to fetch the correct POM.

when fetching that latest version, sometimes a Maven Central package's `maven-metadata.xml` may contain an interpolation string for the version, e.g. `${revision}` in https://repo1.maven.org/maven2/io/github/caffetteria/data-service-opencmis/maven-metadata.xml .

Libraries has no general way of resolving this to fetch the `revision` variable. When we run into this case, we can scrape the HTML file listing page for the package instead, and pick the highest version folder to use as `latest_version` instead.


app/models/package_manager/maven/maven_central.rb
```diff
@@ -16,6 +16,32 @@ def self.missing_version_remover
...
+    if metadata_version =~ /\$\{/
+      version_listing = get_html(MavenUrl.from_name(name, repository_base, NAME_DELIMITER).base)
+      scraped_version = version_listing
+        .css("#contents a")                                  # scrape the list of file/folders
+        .map(&:text)                                         # get each innerText
+        .select { |text| text.end_with?("/") }               # only look at folders
+        .map { |folder| folder.chomp("/") }                  # remove folder trailing slash
+        .grep(/^\d+.\d/)                                     # only folders that look like versions
+        .max_by do |text|
+          # Maven versions range from 1 to many "." and may not be valid SemVer. Use the more forgiving Gem::Version to sort
```
tiegz, line 33: I took a sample of 30k latest_versions for Maven packages, and found a pretty wide spectrum of version digits ("2" being SemVer-compliant for most cases):

```
> latest_versions.map { |v| v.count(".") }.tally
=> {1=>6704, 2=>18558, 0=>332, 3=>2591, 4=>71, 6=>1, 5=>13
```

app/models/package_manager/maven/maven_central.rb
```diff
@@ -16,6 +16,32 @@ def self.missing_version_remover
...
+          Gem::Version.new(text)
+        rescue ArgumentError
```
jhan217, line 35: Should we log on it in case there's an edge case we didn't anticipate but could be processable?

app/models/package_manager/maven/maven_central.rb
```diff
@@ -16,6 +16,32 @@ def self.missing_version_remover
...
+          nil
+        end
+
+      return scraped_version if scraped_version
+    end
+
+    metadata_version
+  end
```
jhan217, line 43: 🎱 What do you think about a structure with a single return/exit to make it easier to understand?

```
def latest_version
  metadata_version = super

  scraped_version = if metadata_version =~ /\$\{/
     version_listing....
  end

  scraped_version || metadata_version
end
```

app/models/package_manager/maven/maven_central.rb
```diff
@@ -16,6 +16,32 @@ def self.missing_version_remover
     PackageManager::Base::MissingVersionRemover
   end
 
+  def self.latest_version(name)
+    metadata_version = super
+
+    # When maven-metadata.xml returned an interpolation string for latest version, e.g. ${revision},
+    # attempt to scrape MavenCentral's index HTML to infer the latest version.
+    if metadata_version =~ /\$\{/
+      version_listing = get_html(MavenUrl.from_name(name, repository_base, NAME_DELIMITER).base)
```
jhan217, line 25: Is it possible to pass the `get_html` call from elsewhere? Just trying to minimize the amount of methods making calls to external third parties that are unexpected and might cause delays in the processing chain (e.g. when asking an object about an attribute I would not expect it to make a call externally).

spec/models/package_manager/maven/maven_central_spec.rb
```diff
@@ -3,6 +3,20 @@
 require "rails_helper"
 
 describe PackageManager::Maven::MavenCentral do
+  describe ".latest_version" do
+    context "when maven-metadata.xml's version contained an interpolation string" do
+      let(:project_name) { "io.github.caffetteria:data-service-opencmis" }
+
+      it "scrapes the maven HTML to find a real version" do
+        latest_version = VCR.use_cassette("maven-central/data-service-opencmis") do
```
jhan217, line 11: Is it possible to include only the relevant bits of the yml in here, perhaps via a stubbed method, so it's easier to understand what the structure is that we're working with in this test?

app/models/package_manager/maven/maven_central.rb
tiegz: too late, the Maven class is already riddled with these 😆 seriously though, I can try to move this into the project() method.

app/models/package_manager/maven/maven_central.rb
tiegz: I think I'd rather notify us, since logs can get forgotten and just linger in the code forever.

spec/models/package_manager/maven/maven_central_spec.rb
tiegz: done!

jhan217 (approved):

app/models/package_manager/maven/maven_central.rb
```diff
@@ -16,6 +16,23 @@ def self.missing_version_remover
...
+      .css("#contents a")                                  # scrape the list of file/folders
+      .map(&:text)                                         # get each innerText
+      .select { |text| text.end_with?("/") }               # only look at folders
+      .map { |folder| folder.chomp("/") }                  # remove folder trailing slash
+      .grep(/^\d+.\d/)                                     # only folders that look like versions
+      .max_by do |text|
+        # Maven versions range from 1 to many "." and may not be valid SemVer. Use the more forgiving Gem::Version to sort
+        Gem::Version.new(text)
+      rescue ArgumentError
+        Bugsnag.notify("Couldn't find scraped HTML version for #{name}. Check the HTML and ensure scraping still works.")
```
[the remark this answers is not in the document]
tiegz: would be nice to extrapolate on this, but I think I'll push that to next time 👍🏻 This "${revision}" use case might be rare enough that it's not a huge deal to let this sit for a while

PR #3364, opened 2024-04-15
Repository: allow status: "Removed" to be unset

mpace965:
Presently, if we mark a Repository as with `status: "Removed"`, that is a terminal state which the Repository cannot leave. This PR enables unsetting the Repository's `status`, provided that the repository is reachable, and we get `archived: false` from the RepositoryHost.

### DIFF
--- a/app/domain/repository/persist_repository_from_upstream.rb
+++ b/app/domain/repository/persist_repository_from_upstream.rb
@@ -111,13 +111,16 @@ class Repository::PersistRepositoryFromUpstream
                             })
       # set to unmaintained if we do not have another status already assigned
       "Unmaintained"
-    elsif !archived_upstream && repository.status == "Unmaintained"
-      StructuredLog.capture("REPOSITORY_REMOVE_UNMAINTAINED_STATUS",
-                            {
-                              repository_host: repository.host_type,
-                              repository_id: repository.id,
-                              full_name: repository.full_name,
-                            })
+    elsif !archived_upstream && %w[Unmaintained Removed].include?(repository.status)
+      if repository.status == "Unmaintained"
+        StructuredLog.capture("REPOSITORY_REMOVE_UNMAINTAINED_STATUS",
+                              {
+                                repository_host: repository.host_type,
+                                repository_id: repository.id,
+                                full_name: repository.full_name,
+                              })
+      end
+
       # set back to nil if we currently have it marked as unmaintained
       nil
     else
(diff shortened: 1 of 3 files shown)


app/domain/repository/persist_repository_from_upstream.rb
```diff
@@ -111,13 +111,16 @@ def self.correct_status_from_upstream(repository, archived_upstream:)
...
       "Unmaintained"
-    elsif !archived_upstream && repository.status == "Unmaintained"
-      StructuredLog.capture("REPOSITORY_REMOVE_UNMAINTAINED_STATUS",
-                            {
-                              repository_host: repository.host_type,
-                              repository_id: repository.id,
-                              full_name: repository.full_name,
-                            })
+    elsif !archived_upstream && %w[Unmaintained Removed].include?(repository.status)
+      if repository.status == "Unmaintained"
```
jhan217, line 115: 🎱 Thoughts about removing the nested conditional and just logging on both, something like `REPOSITORY_UNSETTING_REMOVED_STATUS` with a `current_status` key/value? If we're interested in when unmaintained gets removed, we could also be interested in when the removed status gets removed too.

app/models/repository_host/raw_upstream_data_converter.rb
```diff
@@ -52,7 +52,7 @@ def self.convert_from_gitlab_api(api_project)
       is_private: api_project.visibility != "public",
       keywords: api_project.topics,
       language: nil, # separate API endpoint that doesn't seem to be supported by the API gem we use,
-      license: api_project.license.key,
+      license: api_project.license&.key,
```
jhan217, line 55: Is this related to the status change....?

mpace965: It's not, but I did encounter it when playing with some Gitlab requests

app/domain/repository/persist_repository_from_upstream.rb
```diff
@@ -111,13 +111,15 @@ def self.correct_status_from_upstream(repository, archived_upstream:)
                             })
       # set to unmaintained if we do not have another status already assigned
       "Unmaintained"
-    elsif !archived_upstream && repository.status == "Unmaintained"
-      StructuredLog.capture("REPOSITORY_REMOVE_UNMAINTAINED_STATUS",
+    elsif !archived_upstream && %w[Unmaintained Removed].include?(repository.status)
+      StructuredLog.capture("REPOSITORY_UNSETTING_REMOVED_STATUS",
```
mikeyoung85, line 115:
```suggestion
      StructuredLog.capture("REPOSITORY_UNSETTING_STATUS",
```
I know the previous comment suggested this but maybe remove the `REMOVED_STATUS` naming and have `REPOSITORY_UNSETTING_STATUS` because it could be either Removed or Unmaintained being unset here. Minor detail so feel free to ignore since we will likely be searching by whatever is string is set here.

mikeyoung85 (approved):

PR #3365, opened 2024-04-16
Repository: do a manual sync if we don't get a 404 for the repository

mpace965:
Follow-up to https://github.com/librariesio/libraries.io/pull/3364

We don't actually unset `status: "Removed"` until `update_from_repository`. This will short-circuit the update if we get a 404, but will proceed if we can resolve the URL.

### DIFF
--- a/app/models/repository.rb
+++ b/app/models/repository.rb
@@ -295,8 +295,7 @@ class Repository < ApplicationRecord
 
   def update_all_info(token = nil)
     token ||= AuthToken.token if host_type == "GitHub"
-    check_status
-    return if status == "Removed"
+    return unless check_status
 
     update_from_repository(token)
     unless last_synced_at && last_synced_at > 2.minutes.ago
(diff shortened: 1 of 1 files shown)


app/models/repository.rb
```diff
@@ -350,7 +349,11 @@ def self.check_status(host_type, repo_full_name)
           project.update_attribute(:status, "Removed")
         end
       end
+
+      true
     end
+
+    false
```
tiegz, line 356: it didn't seem obvious at first what the return values should be for this method. Should we document it here, e.g. "Returns true if status was updated, and false otherwise."?

jhan217: Any reason to not have a spec to enforce the behavior?

mpace965: I changed around the cases, so that false is the unreachable case and true is a non 404 case. And documented the method.

tiegz (approved):

PR #3371, opened 2024-04-22
Remove Manifest logic and code

tiegz:
in the quest to [phase out RepositoryDependency](https://github.com/librariesio/libraries.io/pull/3362), this removes any views or logic around Manifests. Manifests aren't used at all outside of RepositoryDependency code.

### Why?

Because data in the Manifest table are disconnected from data in the Projects and Dependency tables, which we want to use instead of RepositoryDependency:

``` mermaid
flowchart BT
Project --> Repository
Dependency --> Project
Manifest --> Repository
RepositoryDependency --> Manifest

```

### DIFF
--- a/app/models/concerns/repo_metadata.rb
+++ b/app/models/concerns/repo_metadata.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-module RepoManifests
+module RepoMetadata
   def download_metadata(token = nil)
     file_list = get_file_list(token)
     return if file_list.blank?
--- a/app/models/repository.rb
+++ b/app/models/repository.rb
@@ -66,6 +66,7 @@
 #
 class Repository < ApplicationRecord
   include Status
+  include RepoMetadata
   include RepositorySourceRank
 
   # eager load this module to avoid clashing with Gitlab gem in development
--- a/app/models/repository.rb
+++ b/app/models/repository.rb
@@ -66,7 +66,6 @@
 #
 class Repository < ApplicationRecord
   include Status
-  include RepoManifests
   include RepositorySourceRank
 
   # eager load this module to avoid clashing with Gitlab gem in development
@@ -84,10 +83,8 @@ class Repository < ApplicationRecord
   has_many :contributors, through: :contributions, source: :repository_user
   has_many :tags, dependent: :delete_all
   has_many :published_tags, -> { published }, anonymous_class: Tag
-  has_many :manifests, dependent: :destroy
   has_many :repository_dependencies
   has_many :repository_maintenance_stats, dependent: :destroy
-  has_many :dependencies, through: :manifests, source: :repository_dependencies
   has_many :dependency_projects, -> { group("projects.id").order(Arel.sql("COUNT(projects.id) DESC")) }, through: :dependencies, source: :project
   has_many :dependency_repos, -> { group("repositories.id") }, through: :dependency_projects, source: :repository
 
@@ -121,9 +118,6 @@ class Repository < ApplicationRecord
   scope :open_source, -> { where(private: false) }
   scope :from_org, ->(org_id) { where(repository_organisation_id: org_id) }
 
-  scope :with_manifests, -> { joins(:manifests) }
-  scope :without_manifests, -> { includes(:manifests).where(manifests: { repository_id: nil }) }
-
   scope :with_description, -> { where("repositories.description <> ''") }
   scope :with_license, -> { where("repositories.license <> ''") }
   scope :without_license, -> { where("repositories.license IS ? OR repositories.license = ''", nil) }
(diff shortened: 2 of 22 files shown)


app/models/concerns/repo_metadata.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-module RepoManifests
+module RepoMetadata
```
tiegz, line 3: forgot to rename this in the previous PR

app/models/repository.rb
```diff
@@ -357,7 +352,6 @@ def self.update_from_hook(uuid, sender_id)
     repository = Repository.where(host_type: "GitHub").find_by_uuid(uuid)
     user = Identity.where("provider ILIKE ?", "github%").where(uid: sender_id).first.try(:user)
     if user.present? && repository.present?
-      repository.download_metadata(user.token)
```
tiegz, line 360: `download_metadata` actually gets called in the `update_all_info_async` hook on the next line, so Libraries was fetching manifests twice when it received a hook 🤦🏻

app/models/repository_organisation.rb
```diff
@@ -57,6 +55,19 @@ class RepositoryOrganisation < ApplicationRecord
            :download_org_from_host_by_login, :download_repos, :download_members,
            :check_status, to: :repository_owner
 
+  # TODO: can this be an association again if we made projects_dependencies a Repository association again?
```
mpace965, line 58: How difficult is this to do? Seems like a lot (all?) of the calls to `projects_dependencies` in this PR are N+1s in the number of repositories

jhan217 (approved):
Seems fine to me, might be good to get another signoff though as I'm not sure what the downstream impacts would be (if any)

app/models/user.rb
```diff
@@ -42,13 +42,6 @@ class User < ApplicationRecord
 
   has_many :watched_repositories, source: :repository, through: :repository_subscriptions
 
-  has_many :dependencies, through: :source_repositories
-  has_many :really_all_dependencies, through: :all_repositories, source: :dependencies
-  has_many :all_dependent_projects, -> { group("projects.id") }, through: :really_all_dependencies, source: :project
```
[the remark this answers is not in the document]
tiegz: yeah, those were both unused. This made me realize that `RepositoryOrganization#all_dependent_repos` is unused as well, so thanks for asking! 😆

app/models/repository_organisation.rb
tiegz: looks like I'm able to turn `projects_dependencies` into an association, so this should be possible. I'll give it a go

app/models/repository.rb
```diff
@@ -257,15 +253,6 @@ def fallback_avatar_url(size = 60)
...
 
-  # TODO: this could probably be refactored into an association
-  def projects_dependencies(includes: nil, only_visible: false)
-    (only_visible ? projects.visible : projects)
-      .map(&:latest_version)
-      .compact
-      .flat_map { |v| v.dependencies.includes(includes) }
-      .uniq { |d| [d.project_id, d.requirements] }
-  end
-
```
tiegz, line 268: was able to move this back to associations

app/models/repository_organisation.rb
```diff
@@ -31,9 +31,8 @@ class RepositoryOrganisation < ApplicationRecord
   has_many :repositories
   has_many :source_repositories, -> { where fork: false }, anonymous_class: Repository
   has_many :open_source_repositories, -> { where fork: false, private: false }, anonymous_class: Repository
-  has_many :dependencies, through: :open_source_repositories
-  has_many :favourite_projects, -> { group("projects.id").order(Arel.sql("COUNT(projects.id) DESC, projects.rank DESC")) }, through: :dependencies, source: :project
-  has_many :all_dependent_repos, -> { group("repositories.id") }, through: :favourite_projects, source: :repository
```
tiegz, line 36: `all_dependent_repos` was unused

app/models/repository_user.rb
```diff
@@ -35,9 +35,8 @@ class RepositoryUser < ApplicationRecord
   has_many :repositories
   has_many :source_repositories, -> { where fork: false }, anonymous_class: Repository
   has_many :open_source_repositories, -> { where fork: false, private: false }, anonymous_class: Repository
-  has_many :dependencies, through: :open_source_repositories
-  has_many :favourite_projects, -> { group("projects.id").order(Arel.sql("COUNT(projects.id) DESC, projects.rank DESC")) }, through: :dependencies, source: :project
-  has_many :all_dependent_repos, -> { group("repositories.id") }, through: :favourite_projects, source: :repository
```
tiegz, line 40: `all_dependent_repos` was unused

mikeyoung85 (approved):

PR #3374, opened 2024-04-26
Remove remaining RepositoryDependency code

tiegz:
this is the last set of changes needed before we can drop the `RepositoryDependency` table:

* removes any more references to `RepositoryDependency,` other than the basic model definition
* also removes references and code for the `ProjectDependentRepository` db view


lib/tasks/one_off.rake
```diff
@@ -3,6 +3,18 @@
 require_relative "input_tsv_file"
 
 namespace :one_off do
+  # this table is huge and requires a manual batch-deletion
+  desc "clean out repository_dependencies table"
+  task batch_delete_repository_dependencies: :environment do
+    RepositoryDependency.in_batches.each_with_index do |b, idx|
```
mikeyoung85, line 9: Why wouldn't we drop the table? Is there a performance issue with dropping everything at once?

tiegz: this table doesn't have any foreign keys attached and shouldn't have any transactions using it, but I've still never dropped a table this large 😬 It's mostly out of caution, but I can just go for it if you think it's safe.

lib/tasks/projects.rake
```diff
@@ -155,7 +153,7 @@ namespace :projects do
   desc "Refresh project_dependent_repos view"
   task refresh_project_dependent_repos_view: :environment do
     exit if ENV["READ_ONLY"].present?
-    ProjectDependentRepository.refresh
+    raise "The projects:refresh_project_dependent_repos_view task has been disabled after the ProjectDependentRepository code has been removed. This task is a no-op now."
```
tiegz, line 156: leaving the rake task in here, just in case anyone out there is running Libraries themselves.

mpace965 (approved):

PR #3375, opened 2024-04-29
Api::Repositories#show: pass readme back when ?include_readme=true

tiegz:
### Example

``` shell
curl https://libraries.io/api/github/librariesio/bibliothecary?include_readme=true

{
  "full_name": "librariesio/bibliothecary",
 "readme_html_body": "<!DOCTYPE html PUBLIC \"-//W3C//DT ... ",
 ...
}
```

### DIFF
--- a/spec/requests/api/repositories_spec.rb
+++ b/spec/requests/api/repositories_spec.rb
@@ -93,5 +93,19 @@ describe "Api::RepositoriesController" do
 
       expect(json["maintenance_stats"].to_json).to be_json_eql([maintenance_stat.attributes.symbolize_keys.slice(*RepositoryMaintenanceStat::API_FIELDS)].to_json)
     end
+
+    it "doesn't include readme when include_readme=false" do
+      get "/api/github/#{repository.full_name}?include_readme=false", params: { api_key: internal_user.api_key }
+      expect(response).to have_http_status(:success)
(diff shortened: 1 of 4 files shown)


spec/requests/api/repositories_spec.rb
```diff
@@ -93,5 +93,19 @@
...
+      expect(response).to have_http_status(:success)
+      expect(response.content_type).to start_with("application/json")
+      expect(json.as_json.keys).to_not include("readme_html_body")
+    end
+
+    it "includes readme when include_readme=true" do
+      get "/api/github/#{repository.full_name}?include_readme=true", params: { api_key: internal_user.api_key }
+      expect(response).to have_http_status(:success)
+      expect(response.content_type).to start_with("application/json")
+      expect(json.as_json.keys).to include("readme_html_body")
```
mikeyoung85, line 108: Can you add a check in this spec to make sure the body is getting passed back correctly? I know it's covered in the serializer spec but it would be good to have it here in case we ever get rid of the serializer and somehow break the lookup for readme content.

```
let!(:readme) { create(:readme, repository: repository, html: "<html>this is my readme</html>") }
```
and

```
expect(json.readme_html_body).to eq("<html>this is my readme</html>")
```

mikeyoung85 (approved):

PR #3378, opened 2024-05-01
Adds an Api::Repositories#sync internal endpoint

tiegz:
this endpoint is similar to the `Api::Projects#sync` endpoint, and only works for internal keys as well.

e.g. `curl https://libraries.io/api/github/librariesio/bibliothecary/sync?api_key=...

### DIFF
--- a/app/controllers/api/repositories_controller.rb
+++ b/app/controllers/api/repositories_controller.rb
@@ -2,6 +2,7 @@
 
 class Api::RepositoriesController < Api::ApplicationController
   before_action :find_repo, except: :search
+  before_action :require_internal_api_key, only: :sync
 
   def show
     include_readme = ActiveModel::Type::Boolean.new.cast(params[:include_readme])
@@ -43,6 +44,22 @@ class Api::RepositoriesController < Api::ApplicationController
     end
 
     render json: json_hash
+  end
+
+  def sync
+    if @repository.recently_synced?
+      StructuredLog.capture(
+        "REPOSITORY_SYNC_REQUEST_SKIPPED", {
+          host_type: @repository.host_type,
+          full_name: @repository.full_name,
+          repository_last_synced_at: @repository.last_synced_at,
+        }
+      )
+      render json: { error: "Repository has already been synced recently" }
+    else
+      @repository.manual_sync
+      render json: { message: "Repository queued for re-sync" }
+    end
   end
 
   private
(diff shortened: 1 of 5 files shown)


config/routes.rb
```diff
@@ -54,6 +54,7 @@ def matches?(request)
       get "/:host_type/:owner/:name/dependencies", to: "repositories#dependencies", constraints: { name: /[^\/]+/ }
       get "/:host_type/:owner/:name/shields_dependencies", to: "repositories#shields_dependencies", constraints: { name: /[^\/]+/ }
       get "/:host_type/:owner/:name/projects", to: "repositories#projects", constraints: { name: /[^\/]+/ }
+      get "/:host_type/:owner/:name/sync", to: "repositories#sync", constraints: { name: /[^\/]+/ }
```
tiegz, line 57: I don't know why all the API endpoints are `get`, but I went with convention. Since they're behind api tokens and don't need cookie auth, they're safe from CSRF.

spec/requests/api/repositories_spec.rb
```diff
@@ -113,4 +113,38 @@
       end
     end
   end
+
+  describe "GET /api/github/:owner/:name/sync", type: :request do
+    context "without internal api key" do
+      it "forbids action" do
+        get "/api/github/#{repository.full_name}/sync", params: { api_key: create(:user).api_key }
```
jhan217, line 120: What happens with a nil or missing api_key?

tiegz: just added a spec above for that case

app/controllers/api/repositories_controller.rb
```diff
@@ -45,6 +46,22 @@ def shields_dependencies
...
+      StructuredLog.capture(
+        "REPOSITORY_SYNC_REQUEST_SKIPPED", {
+          host_type: @repository.host_type,
+          full_name: @repository.full_name,
+          repository_last_synced_at: @repository.last_synced_at,
+        }
+      )
+      render json: { error: "Repository has already been synced recently" }
+    else
+      @repository.manual_sync
```
mikeyoung85, line 60: Could you add a corresponding log message for when we trigger the manual sync. It probably makes more sense to do that in `Repository#manual_sync` than here. We could capture that data looking through API requests, but would be easier to get a sense of how often we are doing manual syncs per repository with a direct log message.

tiegz: I've done that, and added one for `Project#manual_sync` too

mikeyoung85 (approved):

PR #3380, opened 2024-05-01
Change Dependencies [project_id] index to [project_id version_id]

tiegz:
* removes a `dependencies(project_id)` index because we have a `dependencies(project_id, created_at)` covering index now
* adds a `dependencies(:project_id, :version)` to improve the "get a project's dependent repos" query [here](https://github.com/librariesio/libraries.io/commit/8b09b45e87adb05651b5745657274b41ea6751fc#diff-611e7045e8b0212d101cd856c335296959519af63b80129f51c246a7bbfe7b91R409).

This should help the "dependent repos" query on Project, which goes thru Version.

### DIFF
--- a/db/migrate/20240501171850_add_index_to_dependencies.rb
+++ b/db/migrate/20240501171850_add_index_to_dependencies.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+class AddIndexToDependencies < ActiveRecord::Migration[7.0]
+  disable_ddl_transaction!
+
+  def change
+    # this has already been run on production
+    return if Rails.env.production?
+
+    # the covering index (project_id, ((created_at)::date)) already enables this
+    remove_index :dependencies, :project_id, if_exists: true, algorithm: :concurrently
+    add_index :dependencies, %i[project_id version_id], algorithm: :concurrently
+  end
+end
--- a/app/models/dependency.rb
+++ b/app/models/dependency.rb
@@ -17,9 +17,9 @@
 #
 # Indexes
 #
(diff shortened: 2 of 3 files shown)


db/migrate/20240501171850_add_index_to_dependencies.rb
```diff
@@ -0,0 +1,14 @@
...
+class AddIndexToDependencies < ActiveRecord::Migration[7.0]
+  disable_ddl_transaction!
+
+  def change
+    # this has already been run on production
+    return if Rails.env.production?
+
+    # the covering index (project_id, ((created_at)::date)) already enables this
+    remove_index :dependencies, :project_id, if_exists: true, algorithm: :concurrently
+    add_index :dependencies, %i[project_id version_id], algorithm: :concurrently
```
tiegz, line 12: I'm not 100% sure this will help all cases, but it did seem to help with a fairly popular project:

```
p = Project.find_by platform: "Rubygems", name: "devise"

# BEFORE
> Benchmark.ms { Version.where(id: p.dependents.select(:version_id)).select(:project_id).distinct.count }
=> 4299.041126971133

# AFTER
> Benchmark.ms { Version.where(id: p.dependents.select(:version_id)).select(:project_id).distinct.count }
=> 127.43865209631622
```

The query plans don't seem too different cost-wise, but the new one does seem simpler:

# BEFORE

Screenshot 2024-05-01 at 12 27 05 PM


# AFTER


Screenshot 2024-05-01 at 12 26 59 PM

mikeyoung85 (approved):

PR #3381, opened 2024-05-02
Remove last code reference to "project_dependent_repositories" view

tiegz:


tradiff (approved):
My understanding is this is basically unused endpoint, so it's fine to nerf it.

PR #3385, opened 2024-05-07
Set last_synced_at for Removed Repositories

mikeyoung85:
It seems like we should be setting this if we get a 404 response when requesting repository data.

### DIFF
--- a/app/models/repository.rb
+++ b/app/models/repository.rb
@@ -282,7 +282,10 @@ class Repository < ApplicationRecord
 
   def update_all_info(token = nil)
     token ||= AuthToken.token if host_type == "GitHub"
-    return unless check_status
+    unless check_status
+      update!(last_synced_at: Time.current)
+      return
+    end
 
     update_from_repository(token)
     unless last_synced_at && last_synced_at > 2.minutes.ago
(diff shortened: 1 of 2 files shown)


jhan217 (approved):
If it works it's probably fine, but I think the code could be cleaned up a bit

app/models/repository.rb
```diff
@@ -282,7 +282,10 @@ def update_all_info_async(token = nil)
 
   def update_all_info(token = nil)
     token ||= AuthToken.token if host_type == "GitHub"
-    return unless check_status
+    unless check_status
+      update!(last_synced_at: Time.current)
+      return
+    end
```
tiegz, line 288: could consolidate this with the one below by moving them into an `ensure` section, e.g.:

```
  ...
  ensure
     update!(last_synced_at: Time.current)
  end
```

tiegz (approved):

app/models/repository.rb
mikeyoung85: hmm I don't think we want to use `ensure` here because we specifically aren't setting it unless we were able to gather all the data.

tiegz: oh, like if an exception was raised? True

mikeyoung85:
Refactored the `update_all_info` method a bit to make it a little more straightforward. Let me know what you think of the changes and if I screwed up that `unless last_synced_at && last_synced_at < 2.minutes.ago` logic change.

tiegz (approved):

PR #3386, opened 2024-05-07
Strip null bytes from any data we get from upstream repository hosts.

tiegz:
avoids this error when updating users from upstream repository hosts:

`string contains null byte`

### DIFF
--- a/app/models/repository_owner/base.rb
+++ b/app/models/repository_owner/base.rb
@@ -131,7 +131,7 @@ module RepositoryOwner
       response
         .to_hash
         .with_indifferent_access
-        .transform_values { |v| v.is_a?(String) ? v.strip_null_bytes : v }
+        .transform_values { |v| v.is_a?(String) ? StringUtils.strip_null_bytes(v) : v }
     end
 
     private
--- a/app/models/repository_owner/base.rb
+++ b/app/models/repository_owner/base.rb
@@ -121,6 +121,19 @@ module RepositoryOwner
       raise NotImplementedError
     end
 
+    # Return a HashWithIndifferentAccess, where null bytes are removed from string values,
+    # for each repositories' client response. The possible response Hash wrappers are:
+    #
+    #   * Bitbucket returns Hashie::Mash
+    #   * GitHub returns Sawyer::Resource
+    #   * GitLab returns Gitlab::ObjectifiedHash
+    def self.sanitized_hash_with_indifferent_access_from_client_response(response)
+      response
+        .to_hash
+        .with_indifferent_access
+        .transform_values { |v| v.is_a?(String) ? v.strip_null_bytes : v }
+    end
+
     private
 
     def top_favourite_project_ids
--- a/app/models/repository_owner/bitbucket.rb
+++ b/app/models/repository_owner/bitbucket.rb
@@ -72,10 +72,7 @@ module RepositoryOwner
     def self.create_user(user_hash)
       return if user_hash.nil?
 
-      user_hash = user_hash
-        .to_hash
-        .with_indifferent_access
-        .transform_values { |v| v.is_a?(String) ? v.gsub("\u0000", "") : v }
+      user_hash = sanitized_hash_with_indifferent_access_from_client_response(user_hash)
 
       user_hash = {
         id: user_hash[:uuid],
@@ -112,10 +109,7 @@ module RepositoryOwner
     def self.create_org(org_hash)
       return if org_hash.nil?
 
-      org_hash = org_hash
-        .to_hash
-        .with_indifferent_access
-        .transform_values { |v| v.is_a?(String) ? v.gsub("\u0000", "") : v }
+      org_hash = sanitized_hash_with_indifferent_access_from_client_response(org_hash)
 
       org_hash = {
         id: org_hash[:uuid],
--- a/app/models/repository_owner/bitbucket.rb
+++ b/app/models/repository_owner/bitbucket.rb
@@ -72,7 +72,11 @@ module RepositoryOwner
     def self.create_user(user_hash)
       return if user_hash.nil?
 
-      user_hash = user_hash.to_hash.with_indifferent_access
+      user_hash = user_hash
+        .to_hash
+        .with_indifferent_access
+        .transform_values { |v| v.is_a?(String) ? v.gsub("\u0000", "") : v }
+
       user_hash = {
         id: user_hash[:uuid],
         login: user_hash[:username],
@@ -108,7 +112,11 @@ module RepositoryOwner
     def self.create_org(org_hash)
       return if org_hash.nil?
 
-      org_hash = org_hash.to_hash.with_indifferent_access
+      org_hash = org_hash
+        .to_hash
+        .with_indifferent_access
+        .transform_values { |v| v.is_a?(String) ? v.gsub("\u0000", "") : v }
+
       org_hash = {
         id: org_hash[:uuid],
         login: org_hash[:username],
--- a/config/application.rb
+++ b/config/application.rb
@@ -86,5 +86,8 @@ module Libraries
     config.gtm_id = ENV.fetch("GTM_ID", nil)
     config.tidelift_api_key = ENV.fetch("TIDELIFT_API_KEY", nil)
     config.pg_search_projects_enabled = ENV.key?("PG_SEARCH_PROJECTS_ENABLED")
+
+    # In rails 7.1 we can use autoload_lib instead
+    require "extensions"
   end
 end
--- a/config/application.rb
+++ b/config/application.rb
@@ -87,7 +87,6 @@ module Libraries
     config.tidelift_api_key = ENV.fetch("TIDELIFT_API_KEY", nil)
     config.pg_search_projects_enabled = ENV.key?("PG_SEARCH_PROJECTS_ENABLED")
 
-    # In rails 7.1 we can use autoload_lib instead
     require "string_utils"
   end
 end
--- a/config/application.rb
+++ b/config/application.rb
@@ -88,6 +88,6 @@ module Libraries
     config.pg_search_projects_enabled = ENV.key?("PG_SEARCH_PROJECTS_ENABLED")
 
     # In rails 7.1 we can use autoload_lib instead
-    require "extensions"
+    require "string_utils"
   end
 end
--- a/spec/models/repository_owner/base_spec.rb
+++ b/spec/models/repository_owner/base_spec.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+describe RepositoryOwner::Base do
+  context ".sanitized_hash_with_indifferent_access_from_client_response" do
+    let(:hash) do
+      {
+        "id" => 1234,
+        "name" => "My \u0000Name With\u0000 Null Bytes",
+        "company" => "My \u0000Company",
+      }
+    end
+    let(:result) { RepositoryOwner::Base.sanitized_hash_with_indifferent_access_from_client_response(hash) }
+
+    it "should return a HashWithIndifferentAccess" do
+      expect(result).to be_a(HashWithIndifferentAccess)
+      expect(result[:id]).to eq 1234
+      expect(result["id"]).to eq 1234
+    end
+
+    it "should strip null bytes" do
+      expect(result["name"]).to eq("My Name With Null Bytes")
+      expect(result["company"]).to eq("My Company")
+    end
+  end
+end
--- a/app/models/package_manager/base.rb
+++ b/app/models/package_manager/base.rb
@@ -93,7 +93,7 @@ module PackageManager
 
     private_class_method def self.transform_mapping_values(mapping)
       mapping.try do |p|
-        p.compact.transform_values { |v| v.is_a?(String) ? v.gsub("\u0000", "") : v }
+        p.compact.transform_values { |v| v.is_a?(String) ? v.strip_null_bytes : v }
       end
     end
 
--- a/app/models/package_manager/base.rb
+++ b/app/models/package_manager/base.rb
@@ -93,7 +93,7 @@ module PackageManager
 
     private_class_method def self.transform_mapping_values(mapping)
       mapping.try do |p|
-        p.compact.transform_values { |v| v.is_a?(String) ? v.strip_null_bytes : v }
+        p.compact.transform_values { |v| v.is_a?(String) ? StringUtils.strip_null_bytes(v) : v }
       end
     end
 
--- a/app/models/repository_owner/github.rb
+++ b/app/models/repository_owner/github.rb
@@ -74,7 +74,10 @@ module RepositoryOwner
     def self.create_org(org_hash)
       return if org_hash.nil?
 
-      org_hash = org_hash.to_hash.with_indifferent_access
+      org_hash = org_hash
+        .to_hash
+        .with_indifferent_access
+        .transform_values { |v| v.is_a?(String) ? v.gsub("\u0000", "") : v }
       org = nil
       org_by_id = RepositoryOrganisation.where(host_type: "GitHub").find_by_uuid(org_hash[:id])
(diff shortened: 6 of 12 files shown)


app/models/repository_owner/bitbucket.rb
```diff
@@ -72,7 +72,11 @@ def download_members
     def self.create_user(user_hash)
       return if user_hash.nil?
 
-      user_hash = user_hash.to_hash.with_indifferent_access
+      user_hash = user_hash
+        .to_hash
+        .with_indifferent_access
+        .transform_values { |v| v.is_a?(String) ? v.gsub("\u0000", "") : v }
```
mikeyoung85, line 78: The amount of repeats suggest we need this logic in a common place. Maybe something like a `StringUtils` utility class or maybe in `RepositoryOwner::Base`?

app/models/repository_owner/base.rb
```diff
@@ -121,6 +121,13 @@ def self.fetch_org(id_or_login)
       raise NotImplementedError
     end
 
+    def self.sanitize_response_hash(hash)
+      hash
+        .to_hash
```
jhan217, line 126: isn't the input already a hash...?

config/application.rb
```diff
@@ -86,5 +86,8 @@ class Application < Rails::Application
     config.gtm_id = ENV.fetch("GTM_ID", nil)
     config.tidelift_api_key = ENV.fetch("TIDELIFT_API_KEY", nil)
     config.pg_search_projects_enabled = ENV.key?("PG_SEARCH_PROJECTS_ENABLED")
+
+    # In rails 7.1 we can use autoload_lib instead
```
jhan217, line 90: wait this is just making one method reusable across a bunch of classes why do we need to make it an extension vs just a utils method...?

spec/lib/string_spec.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+describe ActiveRecord do
+  describe "with_statement_timeout" do
```
jhan217, line 6: How does string munging relate to db transactions? Feels like something has gone off the tracks if modifying a string leads to db issues in the fundamental functionality...

spec/models/repository_owner/base_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+      {
+        "id" => 1234,
+        "name" => "My \u0000Name With\u0000 Null Bytes",
+        "company" => "My \u0000Company",
+      }
+    end
+    let(:result) { RepositoryOwner::Base.sanitize_response_hash(hash) }
+
+    it "should return a HashWithIndifferentAccess" do
+      expect(result).to be_a(HashWithIndifferentAccess)
```
jhan217, line 17: 🎱 Any particular reason why the output of a method named `sanitize.....hash`  needs to have indifferent access...? Should it be named something like `sanitize_and_transform_to_indifferent_access_hash`? Should this also deal with the url/utf-8 escaping etc.? (that's what I think of when I see `sanitize_...` in a method name)

app/models/repository_owner/base.rb
tiegz: each client (Github, Gitlab, Bitbucket) returns their own magic Hash wrapper, so we need to convert it to Hash here before we create the HashWithIndifferentAccess. I'll note this in some docs here to make it more clear for the next person looking at this code.

config/application.rb
tiegz: I prefer extensions defined on the primitive types, a la ActiveSupport::CoreExtensions, instead of util classes, and it also seems like a common Ruby convention and I never saw a downside to it.

spec/lib/string_spec.rb
tiegz: woops! Accidentally committed this file

spec/models/repository_owner/base_spec.rb
tiegz: yeah, I can definitely make the method name more specific. I don't think we should add encoding logic here until we need it though (just to keep it simple).

app/models/repository_owner/base.rb
mikeyoung85: We will also be receiving regular Hashes into this method as well. Calls like https://github.com/librariesio/libraries.io/blob/ef9f9e29ffaad4dc45891fe86b5094d6ecb89ec8/app/models/concerns/github_identity.rb#L86 are already transforming some upstream client output into a Hash and passing it in. I don't think the code needs to change to handle those but wanted to point it out.

jhan217: Can each client (Github, Gitlab, Bitbucket) do the magic hash wrapper -> hash transformation before passing in a hash to the method? Not sure why the method should have to deal with all the different types of specialized objects being passed in when just a plain ol' hash in and hash out would do. From a more technical perspective, it means that whatever gets passed in now implicitly needs to have a `to_hash` method (for better or worse), vs the expectations around if you only trade around Hash instances. 

Or you could name it "sanitize_client_responses" to try and denote that this method trades in specialized types of hashes.

config/application.rb
jhan217: The method being added is really only applicable to a specialized type of hashes, so I'm not sure if it's really a globally useful method for all hashes.

Some of the downsides I can see are that it makes it harder to find where this method comes from if it needs to be debugged, as it's a method that's custom but not in an easily accessible class that most custom util methods are expected to be in. If all utils get loaded in this way and there's a name collision in a different utility package then it'll be harder to find and rectify, especially if its an edge case or if another library takes up the name. I'm not sure what this means for application start/load order but it might become a performance drag on application loading time.

tiegz: > The method being added is really only applicable to a specialized type of hashes, so I'm not sure if it's really a globally useful method for all hashes.

the method I'm adding here is a string extension: `String#strip_null_bytes`

> Some of the downsides I can see are that it makes it harder to find where this method comes from if it needs to be debugged

`"some string".method(:strip_null_bytes).source_location`

app/models/repository_owner/base.rb
tiegz: > Can each client (Github, Gitlab, Bitbucket) do the magic hash wrapper -> hash transformation before passing in a hash to the method? 

this PR is really just to fix a specific input bug that is really noisy, and I'm trying to avoid changing functionality. If this code becomes more of a priority I think it'd make sense to go back and refactor.

config/application.rb
tiegz: I've been fine with the CoreExtensions pattern for the past 19 years I've been using it, but if y'all are more comfortable with a utils class I can do that.

tiegz: I don't want to belabor my point and I've already pushed the "StringUtils" class instead, but it seems like a good moment to point out how common it is to extend primitives in Ruby:

```
 (String.instance_methods - Object.methods).map { |s| [s, String.instance_method(s).source_location] }.reject { |a| a[1].nil? || a[1][0] =~ /\/ruby\/3\.2\.0|internal/ }.sort_by(&:first)

[:acts_like_string?, ["lib/ruby/gems/3.2.0/gems/activesupport-7.1.3.2/lib/active_support/core_ext/string/behavior.rb", 5]],
 [:at, ["lib/ruby/gems/3.2.0/gems/activesupport-7.1.3.2/lib/active_support/core_ext/string/access.rb", 29]],
 [:camelcase, ["lib/ruby/gems/3.2.0/gems/activesupport-7.1.3.2/lib/active_support/core_ext/string/inflections.rb", 101]],
 [:camelize, ["lib/ruby/gems/3.2.0/gems/activesupport-7.1.3.2/lib/active_support/core_ext/string/inflections.rb", 101]],
 [:classify, ["lib/ruby/gems/3.2.0/gems/activesupport-7.1.3.2/lib/active_support/core_ext/string/inflections.rb", 239]],
 [:column_of, ["lib/ruby/gems/3.2.0/gems/treetop-1.6.12/lib/treetop/ruby_extensions/string.rb", 2]],
 [:constantize, ["lib/ruby/gems/3.2.0/gems/activesupport-7.1.3.2/lib/active_support/core_ext/string/inflections.rb", 73]],
 [:dasherize, ["lib/ruby/gems/3.2.0/gems/activesupport-7.1.3.2/lib/active_support/core_ext/string/inflections.rb", 148]],
 [:deconstantize, ["lib/ruby/gems/3.2.0/gems/activesupport-7.1.3.2/lib/active_support/core_ext/string/inflections.rb", 177]],
 [:demodulize, ["lib/ruby/gems/3.2.0/gems/activesupport-7.1.3.2/lib/active_support/core_ext/string/inflections.rb", 162]],
 [:downcase_first, ["lib/ruby/gems/3.2.0/gems/activesupport-7.1.3.2/lib/active_support/core_ext/string/inflections.rb", 284]],
 [:exclude?, ["lib/ruby/gems/3.2.0/gems/activesupport-7.1.3.2/lib/active_support/core_ext/string/exclude.rb", 10]],
 [:ext, ["lib/ruby/gems/3.2.0/gems/rake-13.1.0/lib/rake/ext/string.rb", 14]],
 [:first, ["lib/ruby/gems/3.2.0/gems/activesupport-7.1.3.2/lib/active_support/core_ext/string/access.rb", 78]],
 [:foreign_key, ["lib/ruby/gems/3.2.0/gems/activesupport-7.1.3.2/lib/active_support/core_ext/string/inflections.rb", 297]],
 [:from, ["lib/ruby/gems/3.2.0/gems/activesupport-7.1.3.2/lib/active_support/core_ext/string/access.rb", 46]],
 [:html_safe, ["lib/ruby/gems/3.2.0/gems/activesupport-7.1.3.2/lib/active_support/core_ext/string/output_safety.rb", 232]],
 [:humanize, ["lib/ruby/gems/3.2.0/gems/activesupport-7.1.3.2/lib/active_support/core_ext/string/inflections.rb", 262]],
 [:in_time_zone, ["lib/ruby/gems/3.2.0/gems/activesupport-7.1.3.2/lib/active_support/core_ext/string/zones.rb", 9]],
 [:indent, ["lib/ruby/gems/3.2.0/gems/activesupport-7.1.3.2/lib/active_support/core_ext/string/indent.rb", 42]],
 [:indent!, ["lib/ruby/gems/3.2.0/gems/activesupport-7.1.3.2/lib/active_support/core_ext/string/indent.rb", 7]],
 [:inquiry, ["lib/ruby/gems/3.2.0/gems/activesupport-7.1.3.2/lib/active_support/core_ext/string/inquiry.rb", 13]],
 [:is_utf8?, ["lib/ruby/gems/3.2.0/gems/activesupport-7.1.3.2/lib/active_support/core_ext/string/multibyte.rb", 48]],
 [:last, ["lib/ruby/gems/3.2.0/gems/activesupport-7.1.3.2/lib/active_support/core_ext/string/access.rb", 92]],
 [:line_of, ["lib/ruby/gems/3.2.0/gems/treetop-1.6.12/lib/treetop/ruby_extensions/string.rb", 12]],
 [:mb_chars, ["lib/ruby/gems/3.2.0/gems/activesupport-7.1.3.2/lib/active_support/core_ext/string/multibyte.rb", 37]],
 [:parameterize, ["lib/ruby/gems/3.2.0/gems/activesupport-7.1.3.2/lib/active_support/core_ext/string/inflections.rb", 215]],
 [:pathmap, ["lib/ruby/gems/3.2.0/gems/rake-13.1.0/lib/rake/ext/string.rb", 138]],
 [:pathmap_explode, ["lib/ruby/gems/3.2.0/gems/rake-13.1.0/lib/rake/ext/string.rb", 27]],
 [:pathmap_partial, ["lib/ruby/gems/3.2.0/gems/rake-13.1.0/lib/rake/ext/string.rb", 41]],
 [:pathmap_replace, ["lib/ruby/gems/3.2.0/gems/rake-13.1.0/lib/rake/ext/string.rb", 59]],
 [:pluralize, ["lib/ruby/gems/3.2.0/gems/activesupport-7.1.3.2/lib/active_support/core_ext/string/inflections.rb", 35]],
 [:remove, ["lib/ruby/gems/3.2.0/gems/activesupport-7.1.3.2/lib/active_support/core_ext/string/filters.rb", 32]],
 [:remove!, ["lib/ruby/gems/3.2.0/gems/activesupport-7.1.3.2/lib/active_support/core_ext/string/filters.rb", 40]],
 [:safe_constantize, ["lib/ruby/gems/3.2.0/gems/activesupport-7.1.3.2/lib/active_support/core_ext/string/inflections.rb", 86]],
 [:singularize, ["lib/ruby/gems/3.2.0/gems/activesupport-7.1.3.2/lib/active_support/core_ext/string/inflections.rb", 60]],
 [:squish, ["lib/ruby/gems/3.2.0/gems/activesupport-7.1.3.2/lib/active_support/core_ext/string/filters.rb", 13]],
 [:squish!, ["lib/ruby/gems/3.2.0/gems/activesupport-7.1.3.2/lib/active_support/core_ext/string/filters.rb", 21]],
 [:strip_heredoc, ["lib/ruby/gems/3.2.0/gems/activesupport-7.1.3.2/lib/active_support/core_ext/string/strip.rb", 22]],
 [:tableize, ["lib/ruby/gems/3.2.0/gems/activesupport-7.1.3.2/lib/active_support/core_ext/string/inflections.rb", 227]],
 [:tabto, ["lib/ruby/gems/3.2.0/gems/treetop-1.6.12/lib/treetop/ruby_extensions/string.rb", 17]],
 [:titlecase, ["lib/ruby/gems/3.2.0/gems/activesupport-7.1.3.2/lib/active_support/core_ext/string/inflections.rb", 126]],
 [:titleize, ["lib/ruby/gems/3.2.0/gems/activesupport-7.1.3.2/lib/active_support/core_ext/string/inflections.rb", 126]],
 [:to, ["lib/ruby/gems/3.2.0/gems/activesupport-7.1.3.2/lib/active_support/core_ext/string/access.rb", 63]],
 [:to_bool, ["/Users/tiegzaharia/Sites/dependencyci/lib/extensions/string/to_bool.rb", 4]],
 [:to_d, ["lib/ruby/gems/3.2.0/gems/bigdecimal-3.1.6/lib/bigdecimal/util.rb", 72]],
 [:to_date, ["lib/ruby/gems/3.2.0/gems/activesupport-7.1.3.2/lib/active_support/core_ext/string/conversions.rb", 47]],
 [:to_datetime, ["lib/ruby/gems/3.2.0/gems/activesupport-7.1.3.2/lib/active_support/core_ext/string/conversions.rb", 57]],
 [:to_json_raw, ["lib/ruby/gems/3.2.0/gems/oj-3.13.21/lib/oj/mimic.rb", 289]],
 [:to_json_raw_object, ["lib/ruby/gems/3.2.0/gems/oj-3.13.21/lib/oj/mimic.rb", 283]],
 [:to_msgpack, ["lib/ruby/gems/3.2.0/gems/msgpack-1.7.2/lib/msgpack/core_ext.rb", 3]],
 [:to_time, ["lib/ruby/gems/3.2.0/gems/activesupport-7.1.3.2/lib/active_support/core_ext/string/conversions.rb", 22]],
 [:treetop_camelize, ["lib/ruby/gems/3.2.0/gems/treetop-1.6.12/lib/treetop/ruby_extensions/string.rb", 31]],
 [:truncate, ["lib/ruby/gems/3.2.0/gems/activesupport-7.1.3.2/lib/active_support/core_ext/string/filters.rb", 70]],
 [:truncate_bytes, ["lib/ruby/gems/3.2.0/gems/activesupport-7.1.3.2/lib/active_support/core_ext/string/filters.rb", 101]],
 [:truncate_words, ["lib/ruby/gems/3.2.0/gems/activesupport-7.1.3.2/lib/active_support/core_ext/string/filters.rb", 142]],
 [:underscore, ["lib/ruby/gems/3.2.0/gems/activesupport-7.1.3.2/lib/active_support/core_ext/string/inflections.rb", 139]],
 [:upcase_first, ["lib/ruby/gems/3.2.0/gems/activesupport-7.1.3.2/lib/active_support/core_ext/string/inflections.rb", 273]]]
```

katzj (approved):

PR #3387, opened 2024-05-08
Add repository#show Route Using Query Parameters Instead of Path Parameters

mikeyoung85:
This adds another route to get to the `repository#show` API endpoint. We occasionally hit route name conflicts when the `name` path parameter matches to another route. For example sending something like `repositories` will link to the `repository_users#repositories` route instead of our intended route.

### DIFF
--- a/app/controllers/api/repositories_controller.rb
+++ b/app/controllers/api/repositories_controller.rb
@@ -64,8 +64,15 @@ class Api::RepositoriesController < Api::ApplicationController
 
   private
 
+  # handle a missing owner or name parameter when doing the lookup for Repository in find_repo
+  # and return a formatted JSON with the missing parameter name with a 400 HTTP status code
+  rescue_from ActionController::ParameterMissing do |e|
+    errors = { e.param => ["is required"] }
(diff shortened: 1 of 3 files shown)


app/controllers/api/repositories_controller.rb
```diff
@@ -64,8 +64,15 @@ def sync
...
+  # handle a missing owner or name parameter when doing the lookup for Repository in find_repo
+  # and return a formatted JSON with the missing parameter name with a 400 HTTP status code
+  rescue_from ActionController::ParameterMissing do |e|
+    errors = { e.param => ["is required"] }
+    render json: errors, status: :bad_request
+  end
+
   def find_repo
-    full_name = [params[:owner], params[:name]].join("/")
+    full_name = params.require(%i[owner name]).join("/")
```
mikeyoung85, line 75: Without the `params.require` here we get the `ActiveRecord::RecordNotFound` error raised after running a sql query with a bad `full_name` parameter. I think this is probably a slight misuse of strong params so let me know if there is a better alternative.

tiegz (approved):

PR #3389, opened 2024-05-09
Remove Logic to Deprecate a PyPI Project if Latest Version is Yanked

mikeyoung85:
I could not find any support for this logic when researching how to deprecate a package on PyPI and it is causing issues for legitimate packages being labeled incorrectly.

### DIFF
--- a/app/models/package_manager/pypi/json_api_project.rb
+++ b/app/models/package_manager/pypi/json_api_project.rb
@@ -130,12 +130,7 @@ module PackageManager
         is_deprecated = false
         message = nil
 
-        latest_stable = releases.reject(&:prerelease?).last
-
-        if latest_stable&.yanked?
-          is_deprecated = true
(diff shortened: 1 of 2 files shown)


mikeyoung85:
Thoughts on instead of removing the logic instead changing the logic to check if all versions are yanked? There are 158 projects that I think match the original logic and a number of those that I have spot checked have every version yanked and seem to be omitted from search results on PyPI.

jhan217 (approved):

tiegz (approved):
I think I like the idea of checking "every version is yanked?". Even if someone later pushed a new version, we'd refresh the status and it would get marked as non-Deprecated.

wenottingham:
repeating a slack comment - "every version yanked" reads as 'removed', not 'deprecated', to me. But that's messier to do.

tiegz (approved):
oh, good call! That's true

PR #3393, opened 2024-05-13
Mark a PyPI Project Deprecated If All Versions are Yanked

mikeyoung85:
Follow up to https://github.com/librariesio/libraries.io/pull/3389

This PR uses the upstream API data to determine if a PyPI project should be marked as deprecated. It checks for all non pre-release version numbers and if they are all yanked then takes the reason text from the most recent release. This should capture projects like https://pypi.org/project/gitlab/#history which have essentially been completely pulled down for use.

### DIFF
--- a/app/models/package_manager/pypi/json_api_project.rb
+++ b/app/models/package_manager/pypi/json_api_project.rb
@@ -133,6 +133,10 @@ module PackageManager
         stable_releases = releases.reject(&:prerelease?)
         latest_stable = stable_releases.last
 
+        # The VersionProcessor class is also inspecting yanked status
+        # on Releases. Make sure that the code here and in the VersionProcessor
+        # do not conflict.
+        # TODO: refactor this deprecation method to reuse the VersionProcessor status logic
         if stable_releases.all?(&:yanked?)
           is_deprecated = true
           message = latest_stable.yanked_reason
--- a/app/models/package_manager/pypi/json_api_project.rb
+++ b/app/models/package_manager/pypi/json_api_project.rb
@@ -130,7 +130,13 @@ module PackageManager
         is_deprecated = false
         message = nil
 
-        if classifiers.include?(CLASSIFIER_INACTIVE)
+        stable_releases = releases.reject(&:prerelease?)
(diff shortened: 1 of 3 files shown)


app/models/package_manager/pypi/json_api_project.rb
```diff
@@ -130,7 +130,13 @@ def deprecation_status
         is_deprecated = false
         message = nil
 
-        if classifiers.include?(CLASSIFIER_INACTIVE)
+        stable_releases = releases.reject(&:prerelease?)
+        latest_stable = stable_releases.last
+
+        if stable_releases.all?(&:yanked?)
```
jhan217, line 136: Could you add a code comment to link this to https://github.com/librariesio/libraries.io/blob/aed17eeb7130ead25f82b8508b9c78f7f9706d6c/app/models/package_manager/pypi/version_processor.rb#L16-L30, as this is the other place (that I'm aware of) that is processing pypi yanked releases and we probably don't want the two places doing this to have any conflicts

jhan217 (commented):
Should this deprecation call also be added to https://github.com/librariesio/libraries.io/blob/aed17eeb7130ead25f82b8508b9c78f7f9706d6c/app/models/package_manager/base.rb#L148-L157 or somewhere nearby? Otherwise this method could update all the versions to "removed" but not the overall project status (unless it's being done in some method/callback after this blob).

mikeyoung85:
> Should this deprecation call also be added to
> 
> https://github.com/librariesio/libraries.io/blob/aed17eeb7130ead25f82b8508b9c78f7f9706d6c/app/models/package_manager/base.rb#L148-L157
> or somewhere nearby? Otherwise this method could update all the versions to "removed" but not the overall project status (unless it's being done in some method/callback after this blob).

It probably makes sense to set the project deprecation status and message in the `PackageManager.update()` method. Right now it is limited to only being set when we check the Project status and I'm not sure if there is a reason for that. I would like to do that in a separate PR from this since there could be some performance issue that needs to be ironed out if we add an additional call in the project update pipeline.

jhan217 (approved):

PR #3394, opened 2024-05-13
Check if There are No Stable Releases Before Calling PyPI Project Deprecated

mikeyoung85:


app/models/package_manager/pypi/json_api_project.rb
```diff
@@ -137,7 +137,7 @@ def deprecation_status
         # on Releases. Make sure that the code here and in the VersionProcessor
         # do not conflict.
         # TODO: refactor this deprecation method to reuse the VersionProcessor status logic
-        if stable_releases.all?(&:yanked?)
```
mikeyoung85, line 140: I think another option is to consider every release instead of just stable ones and pull the reason from the most recent release even if it's a pre-release version number.

app/models/package_manager/pypi/json_api_project.rb
```diff
@@ -137,7 +137,7 @@ def deprecation_status
...
+        if !stable_releases.empty? && stable_releases.all?(&:yanked?)
           is_deprecated = true
           message = latest_stable.yanked_reason
```
jhan217, line 142: Should this be a safe nav anyways?

mikeyoung85: It should not be nil if `stable_releases` has data in it. I would rather not hide the error if this case does come up as that would mean there is a problem with mapping the upstream data to our release objects in here.

app/models/package_manager/pypi/json_api_project.rb
```diff
@@ -137,7 +137,7 @@ def deprecation_status
         # on Releases. Make sure that the code here and in the VersionProcessor
         # do not conflict.
         # TODO: refactor this deprecation method to reuse the VersionProcessor status logic
-        if stable_releases.all?(&:yanked?)
+        if !stable_releases.empty? && stable_releases.all?(&:yanked?)
```
jhan217, line 140:
```suggestion
        if !stable_releases.empty? && stable_releases.all?(&:yanked?)
        if stable_releases.any? && stable_releases.all?(&:yanked?)
```

tiegz (approved):

PR #3397, opened 2024-06-02
Adds builder classes for mapping() and versions() methods.

tiegz:
The sub-classes of `PackageManager::Base` must define `mapping()` and `versions()` class methods, which return Hashes of data we need to update the project/versions in `PackageManager::Base.update()`.

these two changes should help standardize those two methods:

* adds a `PackageManager::Base::MappingBuilder.build_hash()` method that defines a required set of key and optional set of keys, and returns a Hash with the given keywords. 
* adds a `PackageManager::Base::VersionBuilder.build_hash()` method that defines a required set of keys and optional set of keys, and returns a Hash with the given keywords.

with these two classes, passing unknown or accidental keys back will raise an error, and we'll have a documented way of returning this data so it's obvious and clear.

### DIFF
--- a/app/models/package_manager/base.rb
+++ b/app/models/package_manager/base.rb
@@ -5,6 +5,8 @@ module PackageManager
   # specific release of a pacakge. This retrieval can be triggered via
   # asynchronous jobs, or directly by API endpoints or the console.
   class Base
+    class MethodNotImplementedError < StandardError; end
+
     COLOR = "#fff"
     BIBLIOTHECARY_SUPPORT = false
     BIBLIOTHECARY_PLANNED = false
@@ -164,14 +166,14 @@ module PackageManager
     # package manager. This can be any arbitrary data, and will passed on to the
     # mapping() method to get a standard shape of data.
     def self.project(_name)
-      nil
+      raise MethodNotImplementedError
     end
 
     # Override this in the subclass to map the raw data from project() to
     # a Hash of data that we'll need to save Project and Version records.
     # Use the PackageManager::MappingBuilder to create the Hash.
     def self.mapping(_raw_project)
-      nil
+      raise MethodNotImplementedError
     end
 
     # Returns the versions found within the raw project data for the package.
@@ -179,7 +181,7 @@ module PackageManager
     # retrieve all the information to pass into
     # version_hash_to_version_object.
     def self.versions(_raw_project, _name)
-      nil
+      raise MethodNotImplementedError
     end
 
     def self.versions_as_version_objects(raw_project, name)
--- a/app/models/package_manager/base.rb
+++ b/app/models/package_manager/base.rb
@@ -101,7 +101,7 @@ module PackageManager
       db_project = Project.find_or_initialize_by({ name: mapped_project[:name], platform: db_platform })
       db_project.reformat_repository_url if reformat_repository_url && !db_project.new_record?
       mapped_project[:repository_url] = db_project.repository_url if mapped_project[:repository_url].blank?
-      db_project.attributes = mapped_project.except(:name, :releases, :versions, :version, :dependencies, :properties)
+      db_project.attributes = mapped_project.except(:name, :versions, :version, :dependencies, :properties)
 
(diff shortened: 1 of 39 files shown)


app/models/package_manager/nu_get.rb
```diff
@@ -137,15 +137,15 @@ def self.mapping(raw_project)
...
       nuspec_repo = nuspec_repo["url"] if nuspec_repo
 
-      {
+      MappingBuilder.build_hash(
         name: raw_project[:name],
         description: item.description,
         homepage: item.project_url,
         keywords_array: item.tags,
         repository_url: repo_fallback(nuspec_repo, item.project_url),
-        releases: raw_project[:releases],
```
tiegz, line 146: here's one example of a key that's not needed. It was added years ago in https://github.com/librariesio/libraries.io/commit/70396e221d59b568f2f110022260395a2dc3e64a#diff-8b25ada8935cd42afbba863d31a65269a35dbc467397160353adf3425030ed25R68 , but we actually only allow `:versions` as a key here.

app/models/package_manager/base/mapping_builder.rb
```diff
@@ -0,0 +1,26 @@
...
+          description: description,
+          repository_url: repository_url,
+        }
+
+        hash[:homepage] = homepage if homepage != MISSING
+        hash[:keywords_array] = keywords_array if keywords_array != MISSING
+        hash[:licenses] = licenses if licenses != MISSING
+        hash[:versions] = versions if versions != MISSING
+
+        hash
```
mpace965, line 22: Can we use `nil` / `Object#not_nil?` instead? Same for VersionBuilder?

app/models/package_manager/base.rb
```diff
@@ -160,6 +160,20 @@ def self.update(name, sync_version: :all, force_sync_dependencies: false)
...
+      nil
+    end
+
+    # Override this in the subclass to map the raw data from project() to
+    # a Hash of data that we'll need to save Project and Version records.
+    # Use the PackageManager::MappingBuilder to create the Hash.
+    def self.mapping(_raw_project)
+      nil
+    end
+
```
mpace965, line 176: Is it required that subclasses implement these? We could raise with a more specific error if so.

app/models/package_manager/base/mapping_builder.rb
tiegz: if we set `nil` as the default value for the kwargs above, then we don't have a way to tell if the kwarg was passed as a param or if the value is just the `nil` default value. We'd only want to set the optional keys above (like `:homepage`) if the user explicitly passed a kwarg for it.

I've run into this problem before in Ruby and haven't found a way to tell if the arg was the default value or not, so this seemed like a clever workaround 😬

tiegz: (maybe there's a better name for this than "MISSING" too)

mpace965: Do we rely on the difference in semantics of a key being present and mapped to `nil` and the key being absent from the hash?

tiegz: we're passing almost the entire `mapped_project()` result to the ActiveRecord objects in `ensure_project`, so it's possible we might set an attribute to `nil` and overwrite data, even though the `mapped_project()` method didn't set that key itself.

the goal with that is mostly to maintain the same functionality as before. Before this, each PackageManager class mapped its own combination of keys to set on Project/Versions. But if I start passing all keys for each PackageManager then we've changed the behavior. I could leave a comment here maybe, e.g. "# TODO: if we determine it's safe to default to 'nil' for any of these keys then we can change the kwarg default to 'nil'." 

(I thought we did with the repository_url [here](https://github.com/librariesio/libraries.io/blob/main/app/models/package_manager/base.rb#L103C1-L103C108), but it turns out it's using `blank?` instead of `key?`, so this wouldn't help there.)

mpace965 (approved):

PR #3398, opened 2024-06-03
PackageManager::NuGet: stop recording "1900" published dates

tiegz:
changes to `PackageManager::Nuget`:
* rename `get_releases()` to `raw_versions()`, to avoid confusion and make more consistent with the `versions()` method
* also rename the `:releases` key in its raw project data to `:raw_versions` for consistency
* when we get a raw version from NuGet's API that has "published=1900-01-01...", **do not** update the "published_at" on our side to 1900 (so we keep the original published_at date, and set the `Version`'s "status" column to "Deprecated"

this would be Libraries' first use-case for deprecated Versions.

### DIFF
--- a/app/models/package_manager/nu_get.rb
+++ b/app/models/package_manager/nu_get.rb
@@ -151,11 +151,21 @@ module PackageManager
 
     def self.versions(raw_project, _name)
       raw_project[:raw_versions].map do |raw_version|
-        VersionBuilder.build_hash(
+        version_data = {
           number: raw_version.version_number,
           published_at: raw_version.published_at,
-          original_license: raw_version.original_license
-        )
+          original_license: raw_version.original_license,
+        }
+
+        if raw_version.deprecation.present?
+          # We considered NuGet versions that are unlisted as "Deprecated", and those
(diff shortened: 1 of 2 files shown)


app/models/package_manager/nu_get.rb
```diff
@@ -132,34 +133,44 @@ def self.get_releases(name)
...
+          original_license: raw_version.original_license,
+        }
+
+        if raw_version.deprecation.present?
+          # We considered NuGet versions that are unlisted as "Deprecated", and those
+          # versions have their "published" reset to 1/1/1900 in the NuGet API.
+          # Keep the original "published_at" on our side but mark it as deprecated.
+          version_data.delete(:published_at)
+          version_data[:status] = "Deprecated"
+        end
```
tiegz, line 166: @mpace965 re: [this discussion](https://github.com/librariesio/libraries.io/pull/3397#discussion_r1624786708), I forgot to mention that we'll want to avoid setting `published_at: nil` in this change. Instead we're keeping the original "published_at" value instead of setting to 1900 or to nil.

PR #3399, opened 2024-06-03
CocoaPods bugfix: pass kwargs to VersionBuilder instead of a Hash

tiegz:
(followup to https://github.com/librariesio/libraries.io/pull/3397)

```ArgumentError 
wrong number of arguments (given 1, expected 0; required keyword: number)
```


app/models/package_manager/cocoa_pods.rb
```diff
@@ -50,9 +50,7 @@ def self.mapping(raw_project)
 
     def self.versions(raw_project, _name)
```
jhan217, line 51: 🎱 Should there be a spec or something to test these lines of code before it hits prod?

jhan217 (approved):

PR #3400, opened 2024-06-07
Add rake task to backfill nuget published_at dates

tiegz:
(followup to https://github.com/librariesio/libraries.io/pull/3398 )

This adds a `one_off:backfill_nuget_published_at` rake task that will 
1) pull a list of all NuGet Version records where `published_at: '1900-01-01 00:00:00'`, which means unlisted (or "Deprecated" in Libraries terms)
2) iterate over [NuGet's Catalog API endpoint](https://learn.microsoft.com/en-us/nuget/api/catalog-resource#catalog-leaf)
3) if one of the releases from the API is in the list from step (1), and the API date is not 1900, then update the Version record to the original "published" date (and mark it as `status: "Deprecated"`)

* the `catalog_start_idx` argument lets you skip over catalog pages, to not re-do work if you restart the task
* the `threads` argument lets you adjust the level of parallelism when fetching release pages

Screenshot 2024-06-07 at 12 01 49 PM

### DIFF
--- a/lib/tasks/one_off.rake
+++ b/lib/tasks/one_off.rake
@@ -25,4 +25,66 @@ namespace :one_off do
       end
     end
   end
+
+  # This task tracks down the original "published" dates for NuGet Versions that
+  # are marked as "published_at=1900-01-01 00:00:00" in Libraries, and updates
+  # them with the original "published" date, and marks them 'status: "Deprecated"'.
+  #
+  # It processes lists-of-lists from NuGet's API:
+  #   * [CatalogRoot] Catalog API (https://learn.microsoft.com/en-us/nuget/api/catalog-resource)
+  #     * .items => [CatalogPage]
+  #       * .items => [PackageDetails] (each PackageDetails may contain the original "published")
+  #
+  # @param catalog_start_idx Integer the index of the catalog page to start with,
+  #                          to skip already-processed pages (default: 0)
+  # @param threads Integer the number of parallel threads to process version items (default: 8)
+  desc "backfill unknown nuget published_at dates"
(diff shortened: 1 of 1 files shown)


lib/tasks/one_off.rake
```diff
@@ -25,4 +25,66 @@ namespace :one_off do
...
+        published = page_item["published"]
+
+        if published.nil?
+          raise "'published' field not found on page #{catalog_item['@id']} in #{item['@id']}."
+        elsif unlisted_name_versions.key?([version, name]) && published !~ /1900-/
+          p = Project.find_by(platform: "NuGet", name: name)
+          v = p.versions.find_by(number: version)
+          if v.published_at.year == 1900
+            published_at = Time.parse(published)
+            v.update_columns(published_at: published, status: "Deprecated")
```
mpace965, line 79: Might these versions be undeprecated if they now have a published_at > 1900?

tiegz: the "Catalogs" list of a list of all events, so the event we're looking at here would be an older one before the version was unlisted/1900'ed. 

I'm not sure if NuGet lets you re-list an "unlisted" version, but if that happened then we'll see it in the next `PackageManager::NuGet.update()` for the  package. This task is basically just responsible for swapping the `published_at: '1900-01-01 00:00:00'" == DEPRECATED` equivalency with `status: "Deprecated" == DEPRECATED`.

but that is a great question, because I should probably handle the case of "re-listing" in the NuGet updater here: https://github.com/librariesio/libraries.io/blob/49a931d438f13799d673b37b677ad31ff3559a44/app/models/package_manager/nu_get.rb#L160-L166

mpace965 (approved):

PR #3405, opened 2024-06-10
Some tweaks to NuGet update logic and rake task

tiegz:
just a few fixes that I ran into over the weekend, while running this rake task.

### DIFF
--- a/app/models/package_manager/nu_get.rb
+++ b/app/models/package_manager/nu_get.rb
@@ -158,10 +158,14 @@ module PackageManager
         }
 
         if raw_version.deprecation.present?
-          # We considered NuGet versions that are unlisted as "Deprecated", and those
-          # versions have their "published" reset to 1/1/1900 in the NuGet API.
-          # Keep the original "published_at" on our side but mark it as deprecated.
-          version_data.delete(:published_at)
+          # NuGet versions can be marked as deprecated as returned by the API, e.g.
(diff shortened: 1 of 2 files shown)


app/models/package_manager/nu_get.rb
```diff
@@ -158,10 +158,14 @@ def self.versions(raw_project, _name)
...
-          # Keep the original "published_at" on our side but mark it as deprecated.
-          version_data.delete(:published_at)
+          # NuGet versions can be marked as deprecated as returned by the API, e.g.
+          # "EventBus.RabbitMQ.Standard@1.0.9".
+          # Additionally, if they have a "published" date of "1900-01-01 00:00:00"
+          # then they're "unlisted" too, and won't be shown on nuget.org, e.g.
+          # "Microsoft.IdentityModel@6.1.7600.16394".
+          # When "unlisted", we want to maintain the original "published" date
+          # so don't overwrite it with "1900-01-01 00:00:00".
+          version_data.delete(:published_at) if version_data[:published_at].year < 1901
```
tiegz, line 168: it became clear on Friday that there are "deprecated" versions that don't have "1900" as a published date, which means they might have predated the "1900" designation. 

Just in case the API returns versions with a deprecation but no "1900", let's update the "published_at" value on our side.

jhan217 (approved):

PR #3407, opened 2024-07-02
Handle weird case where latest in maven-metadata doesn't exist

katzj:
There are some cases (for example, commons-discovery:commons-discovery) where the version listed as latest in maven-metadata.xml isn't actually present on central. This lets us fallback to our scraping and using that.

Also improves the scraping of the latest version to prefer non date formatted versions as that's an older maven practice

### DIFF
--- a/app/models/package_manager/maven/maven_central.rb
+++ b/app/models/package_manager/maven/maven_central.rb
@@ -18,19 +18,24 @@ class PackageManager::Maven::MavenCentral < PackageManager::Maven::Common
 
   # Attempt to scrape MavenCentral's index HTML to infer the latest version.
   def self.latest_version_scraped(name)
-    get_html(MavenUrl.from_name(name, repository_base, NAME_DELIMITER).base)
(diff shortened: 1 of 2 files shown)


app/models/package_manager/maven/maven_central.rb
```diff
@@ -18,19 +18,24 @@ def self.missing_version_remover
...
-        Gem::Version.new(text)
-      rescue ArgumentError
-        Bugsnag.notify("Couldn't find scraped HTML version for #{name}. Check the HTML and ensure scraping still works.")
-        nil
-      end
+
+    # Maven versions range from 1 to many "." and may not be valid SemVer. Use the more forgiving Gem::Version to sort
+    # but we also want to prefer things that _don't_ look like a date as that's an ancient maven practice
+    dated_versions, nodate_versions = versions.partition {|v| v =~ /\d{8}.\d+/ }
+    if nodate_versions.count
```
tiegz, line 31: should this be `nodate_versions.present?` or `nodate_versions.count > 0`?

tiegz (approved):

PR #3419, opened 2024-07-24
Add Logic To Fetch Tokens With Specific Scopes

mikeyoung85:
This adds some logic to grab tokens with specific scopes when needed. The logic is used to help re-enable to the API call to gather document URLs for repositories as that call needs some sort of repo scope enabled.

### DIFF
--- a/app/models/auth_token.rb
+++ b/app/models/auth_token.rb
@@ -15,6 +15,8 @@
 class AuthToken < ApplicationRecord
   validates_presence_of :token
   scope :authorized, -> { where(authorized: [true, nil]) }
+  # find tokens that include ANY of the scopes provided
+  scope :with_either_scope, ->(required_scope) { where("scopes && array[?]::varchar[]", Array(required_scope)) }
 
   LOW_RATE_LIMIT_REMAINING_THRESHOLD = 500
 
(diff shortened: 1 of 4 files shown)


app/models/auth_token.rb
```diff
@@ -15,6 +15,8 @@
 class AuthToken < ApplicationRecord
   validates_presence_of :token
   scope :authorized, -> { where(authorized: [true, nil]) }
+  # find tokens that include ANY of the scopes provided
+  scope :with_either_scope, ->(required_scope) { where("scopes && array[?]::varchar[]", Array(required_scope)) }
```
mikeyoung85, line 19: The name of this scope is a little weird. I wanted something to imply that if any of the scope names match then a result will be returned rather than having to match all of the scopes passed in.

app/models/repository_host/github.rb
```diff
@@ -58,23 +58,31 @@ def commits_url(author = nil)
...
+      required_scope = %w[repo public_repo]
+      token ||= AuthToken.find_token(:v4, required_scope: required_scope).token
+
+      github_client = AuthToken.fallback_client(token)
+      api_hash = github_client.repo(id_or_name, accept: "application/vnd.github.drax-preview+json,application/vnd.github.mercy-preview+json").to_hash
+
+      owner = api_hash.dig(:owner, :login)
+      repository_name = api_hash[:name]
+
+      # verify the token being used has the correct scopes for the document URLs API call
```
mikeyoung85, line 70: I wrote it this way in case the token already contains the right scope. It's a difficult tracking down all the places where this method could be called and where the passed in token is decided on so this logic checks the existing token first before finding a new one. This also limits the usage of these tokens to just this API call so they don't become exhausted quickly making other API calls.

app/models/auth_token.rb
macowie: `has_scope`?

macowie (approved):

PR #3423, opened 2024-07-31
Handle removed Pypi projects gracefully when fetching releases

tiegz:
fixes this error:

```
NoMethodError 
CheckStatusWorker@status
undefined method `map' for nil:NilClass @data["releases"].map do |version_number, details| ^^^^
```

### DIFF
--- a/app/models/package_manager/pypi/json_api_project.rb
+++ b/app/models/package_manager/pypi/json_api_project.rb
@@ -79,7 +79,7 @@ module PackageManager
 
       def releases
         JsonApiProjectReleases.new(
-          @data["releases"].map do |version_number, details|
+          @data.fetch("releases", []).map do |version_number, details|
             # this assumes that each file type distribution of the package/version has the same yanked status & reason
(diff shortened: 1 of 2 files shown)


wenottingham:
does this have any data consequences beyond "silencing an error"

tiegz:
nah, from what I can tell this is happening on "Removed" projects, so it makes sense that their releases would be empty.

tiegz:
<rethinking this>

tiegz:
I'm gonna close this. I think the error here is a rare race condition, where:

1) `Project#check_status`: we fetch the package url, and it 200s, so we skip the line that would mark it as "Removed"
2) the package is removed from pypi, which would then the package it to 404
3) a few lines of code after (1), we check `project.platform_class.deprecation_info(project)`, which pulls the package, but it's gone now so we don't get any data back

once this worker retries, it should halt on (1) and mark the package as "Removed", so I don't think we should bother messing with the Pypi ingestion code.

PR #3437, opened 2024-08-14
Add "audit_comment" to Project and Repository whenever we update their statuses.

tiegz:
we've run into a Repository that temporarily had `"status: 'Removed'"`, but is now `"status: nil"`, and we don't know how it got into the "Removed" state.

now that we have the "audited" gem (https://github.com/librariesio/libraries.io/pull/3438), we can audit "status" column changes on both Project and Repository. 

~Libraries doesn't have a great papertrail for "status" changes, so this adds these two columns, and populates them whenever "status" changes:~
~* `Project#status_reason`~
~* `Repository#status_reason`~

~also, some more improvements:~
~* merge `Repository.check_status` into `Repository#check_status` to simplify~
~* in `Repository#check_status`, if we get a 200 back then be sure to update status to `nil`~

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -32,6 +32,7 @@
 #  score_last_calculated              :datetime
 #  status                             :string
 #  status_checked_at                  :datetime
+#  status_reason                      :string
 #  versions_count                     :integer          default(0), not null
 #  created_at                         :datetime         not null
 #  updated_at                         :datetime         not null
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -61,7 +61,7 @@
 class Project < ApplicationRecord
   require "query_counter"
 
-  # include ProjectSearch
+  include ProjectSearch
   include SourceRank
   include Status
   include Releases
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -55,7 +55,8 @@
 #  index_projects_on_status_checked_at              (status_checked_at)
 #  index_projects_on_updated_at                     (updated_at)
 #  index_projects_on_versions_count                 (versions_count)
-#  index_projects_search_on_name                    (COALESCE((name)::text, ''::text) gist_trgm_ops) USING gist
+#  index_projects_search_on_description             (to_tsvector('simple'::regconfig, COALESCE(description, ''::text))) USING gist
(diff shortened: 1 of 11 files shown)


app/models/project.rb
```diff
@@ -664,27 +666,25 @@ def check_status
     StructuredLog.capture("CHECK_STATUS_CHANGE", { platform: platform, name: name, status_code: response.response_code }) if platform.downcase == "npm"
 
     if platform.downcase == "packagist" && [302, 404].include?(response.response_code)
-      update_attribute(:status, "Removed")
+      update(status: "Removed", status_reason: "Response #{response.response_code}")
```
tiegz, line 669: note that `update_attribute` wasn't running validations, but I don't see a big reason to not run them.

wenottingham:
tagged @bee64 since she was just in here

app/models/repository.rb
```diff
@@ -321,27 +323,25 @@ def self.create_from_data(repo_host_data)
...
+      projects.each do |project|
+        next unless %w[bower go elm alcatraz julia nimble].include?(project.platform.downcase)
+
+        project.update_attribute(status: "Removed", status_reason: "Response 404")
       end
 
       return false
+    elsif response.response_code == 200
+      update(status: nil, status_reason: "Response 200")
     end
```
tiegz, line 345: @wenottingham I think this might fix the cases you noticed, where the Repository is "Removed" but the repository link is actually working, e.g.:

Screenshot 2024-08-14 at 1 08 22 PM

app/models/repository.rb
```diff
@@ -321,27 +323,25 @@ def self.create_from_data(repo_host_data)
...
-  def self.check_status(host_type, repo_full_name)
+  def check_status
     domain = RepositoryHost::Base.domain(host_type)
-    response = Typhoeus.head("#{domain}/#{repo_full_name}")
+    response = Typhoeus.head("#{domain}/#{full_name}")
+
+    return true if status == "Hidden" # don't overwrite Hidden projects
 
     if response.response_code == 404
-      repo = Repository.includes(:projects).find_by_full_name(repo_full_name)
```
tiegz, line 334: just noticed this bug! We're not limiting the repo to "host_type: host_type", so this might've been updating repos that had the same path but different host types! 🐛

tradiff (approved):

PR #3438, opened 2024-08-15
Introduce "audited" gem to Librarires

tiegz:
this adds the [`audited` gem](https://github.com/collectiveidea/audited) to Libraries so we can start tracking historical data, e.g. `Project#homepage`. 

In a following PR, I'll lax-en the requirements for a Project's data to be updated, so having `audited` around will give us a little backup in case something goes wrong.

#### Note 

I tried overwriting the "audited" method so that it requires you to pass the "only:" key, such that we limit the data we audit to specific data, but I'm having trouble monkeypatching that method.

### DIFF
--- a/db/migrate/20240815181932_install_audited.rb
+++ b/db/migrate/20240815181932_install_audited.rb
@@ -14,16 +14,16 @@ class InstallAudited < ActiveRecord::Migration[7.0]
       t.column :username, :string
       t.column :action, :string
       t.column :audited_changes, :text
-      t.column :version, :integer, :default => 0
+      t.column :version, :integer, default: 0
       t.column :comment, :string
       t.column :remote_address, :string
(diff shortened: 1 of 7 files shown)


db/migrate/20240815181932_install_audited.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+class InstallAudited < ActiveRecord::Migration[7.0]
+  disable_ddl_transaction!
+
+  def self.up
+    create_table :audits do |t|
+      t.column :auditable_id, :integer
+      t.column :auditable_type, :string
+      t.column :associated_id, :integer
```
tiegz, line 10: Libraries doesn't use any uuid columns, so we should be safe here.

mpace965 (approved):

PR #3439, opened 2024-08-15
Start setting "nil" values we get from upstream package managers.

tiegz:
since 2016 (b066184c9a565f8e1da1e75978d156677e5846ba), Libraries has been ignoring "nil" values it gets from upstream package managers, when updating Projects.

the drawback of this is that we don't update attributes when they become "nil", such as DefinitelyTyped projects (which automatically get replaced with a stubbed empty package when they become deprecated):

[
Screenshot 2024-08-15 at 2 30 57 PM
](url)

now that Libraries has a defined "MappingBuilder" class that explicitly defines the fields that we update (https://github.com/librariesio/libraries.io/pull/3397), we can be more confident about the fields that could be "nil", and we can update fields like "homepage" when they go nil.

### DIFF
--- a/app/models/package_manager/base.rb
+++ b/app/models/package_manager/base.rb
@@ -95,7 +95,7 @@ module PackageManager
 
     private_class_method def self.transform_mapping_values(mapping)
       mapping.try do |p|
-        p.compact.transform_values { |v| v.is_a?(String) ? StringUtils.strip_null_bytes(v) : v }
+        p.transform_values { |v| v.is_a?(String) ? StringUtils.strip_null_bytes(v) : v }
       end
     end
 
--- a/app/models/package_manager/base.rb
+++ b/app/models/package_manager/base.rb
@@ -93,12 +93,6 @@ module PackageManager
       find(platform).try(:formatted_name) || platform
     end
 
-    private_class_method def self.transform_mapping_values(mapping)
-      mapping.try do |p|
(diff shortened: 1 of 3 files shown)


app/models/package_manager/base.rb
```diff
@@ -93,12 +93,6 @@ def self.platform_name(platform)
       find(platform).try(:formatted_name) || platform
     end
 
-    private_class_method def self.transform_mapping_values(mapping)
-      mapping.try do |p|
-        p.compact.transform_values { |v| v.is_a?(String) ? StringUtils.strip_null_bytes(v) : v }
-      end
-    end
```
tiegz, line 100: moved this string-cleaning into MappingBuilder below

app/models/package_manager/base.rb
```diff
@@ -93,12 +93,6 @@ def self.platform_name(platform)
       find(platform).try(:formatted_name) || platform
     end
 
-    private_class_method def self.transform_mapping_values(mapping)
-      mapping.try do |p|
-        p.compact.transform_values { |v| v.is_a?(String) ? StringUtils.strip_null_bytes(v) : v }
```
tiegz, line 98: removing `compact` here is the big change -- we'll now be able to update fields like "homepage" to nil if they are nil. These will also be audited now: https://github.com/librariesio/libraries.io/pull/3438/files#diff-611e7045e8b0212d101cd856c335296959519af63b80129f51c246a7bbfe7b91R109

mpace965 (approved):

PR #3442, opened 2024-08-16
Lock RepositoryDownloadWorker while it's executing too

tiegz:
now that we have [audited on Libraries](https://github.com/librariesio/libraries.io/pull/3438), we're noticing a curious race condition situation:

* Repo1 has two projects: Project1 and Project2
* Project1 and Project2 both get pushed by their maintainer at the same time
* Depper picks up the two updates and enqueues two PackageManagerDownloadWorkers jobs for Libraries
* each PackageManagerDownloadWorker job kicks off a Repository update for the same Repository here: https://github.com/librariesio/libraries.io/blob/96225bffe6ff7bf90535f8dd29fa195b5b8993a6/app/models/package_manager/base.rb#L253
* ... so Project1 kicks off RepositoryDownloadWorker
* and while that RepositoryDownloadWorker ^ is running, Project2 kicks off another RepositoryDownloadWorker
* the first RepositoryDownloadWorker updates the Repository#status 
* and then the second RepositoryDownloadWorker also updates the Repository#status
* so we get two Audited::Audit records for the same repo changes

currently we are locking RepositoryDownloadWorker **from** when it enqueues **until** it starts working, but we should be able to prevent this race condition if we lock it **from** when it enqueues **until** it stops working.

### DIFF
--- a/app/workers/repository_download_worker.rb
+++ b/app/workers/repository_download_worker.rb
@@ -2,7 +2,7 @@
 
 class RepositoryDownloadWorker
   include Sidekiq::Worker
-  sidekiq_options queue: :repo, lock: :until_executed
+  sidekiq_options queue: :repo, lock: :until_and_while_executing, lock_ttl: 10.minutes.to_i
 
   def perform(repo_id, token = nil)
     Repository.find_by_id(repo_id).try(:update_all_info, token)
--- a/config/initializers/sidekiq.rb
+++ b/config/initializers/sidekiq.rb
@@ -64,4 +64,15 @@ SidekiqUniqueJobs.configure do |config|
   config.lock_ttl = 1.day.to_i
 end
 
+SidekiqUniqueJobs.reflect do |on|
+  on.lock_failed do |job_hash|
+    message = {
+      message: "Skipping duplicate job",
+      worker: job_hash["class"],
+      args: job_hash["args"],
+    }
+    Sidekiq.logger.info(message)
+  end
+end
+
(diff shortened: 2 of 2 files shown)


app/workers/repository_download_worker.rb
```diff
@@ -2,7 +2,7 @@
 
 class RepositoryDownloadWorker
   include Sidekiq::Worker
-  sidekiq_options queue: :repo, lock: :until_executed
+  sidekiq_options queue: :repo, lock: :until_and_while_executing, lock_ttl: 10.minutes.to_i
```
mpace965, line 5: Will the lock expire at min(job completed time, lock_ttl), or will it always expire at lock_ttl? We could drop some legitimate close together updates if the latter

tiegz: yeah, the docs say `...how long to wait at most before considering a lock to be expired...`, so the "at most" sounds like it usually won't lock the entire TTL. (I'm having a hard time tracking down that code in the gem tho)

it looks like the P90 latency for this worker is 20-40sec, and the p99.9 reaches about 10 minutes. So if I reduce this to `1.minute.to_i` then we should guarantee locks for ~90% of these jobs?

mpace965: We could keep it at 10 and add some logging to keep tabs on the conflicts: https://github.com/mhenrixon/sidekiq-unique-jobs?tab=readme-ov-file#reflections-metrics-logging-etc

mpace965 (approved):

PR #3444, opened 2024-08-16
Fix Repository.check_status, and ensure we run Project callbacks

tiegz:
(followup to https://github.com/librariesio/libraries.io/pull/3437)

forgot to switch this `update_attribute` to `update` in `Repository#check_status`

### DIFF
--- a/app/models/repository.rb
+++ b/app/models/repository.rb
@@ -331,17 +331,17 @@ class Repository < ApplicationRecord
     return true if status == "Hidden" # don't overwrite Hidden repositories
 
     if response.response_code == 404
-      update(status: "Removed", audit_comment: "Response 404") unless private?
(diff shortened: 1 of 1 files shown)


app/models/repository.rb
```diff
@@ -336,7 +336,7 @@ def check_status
       projects.each do |project|
         next unless %w[bower go elm alcatraz julia nimble].include?(project.platform.downcase)
 
-        project.update_attribute(status: "Removed", audit_comment: "Response 404")
+        project.update(status: "Removed", audit_comment: "Response 404")
```
tiegz, line 339: if a Project gets saved as "Removed", we want callbacks to run so that `ProjectUpdatedWorker` runs

pranav (approved):

PR #3454, opened 2024-08-22
PackageManager::Base.update(): kick off CheckStatusWorker whenever we update a Project.

tiegz:
currently libraries updates a Project's `status` via `CheckStatusWorker` in two situations:

1) via `Project#async_sync`, which is triggered via manual syncs via the Web UI
3) `rake projects:check_status` will kick off nightly batches of 150k CheckStatusWorker prioritizing oldest/nil "status_checked_at".


tiegz:
after some thought, the `rake projects:check_status` should probably work. It looks like it was basically broken until last week, but is now checking status for 150k projects nightly. We can wait and see if that continues to drain the queue of Projects.

PR #3461, opened 2024-08-27
Consider privately scoped NPM packages that return 302 as Removed.

tiegz:
there are some packages in Libraries that return `302` when we fetch them from NPM, which means they're not accessible, i.e. not open-source. This starts marking those libraries as "Removed":

```
> curl -I "https://www.npmjs.com/package/@abcdef/ghijk"
HTTP/2 302 
location: /login?next=%2Fpackage%2F%40abcdef%2Fghijk
```

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -668,7 +668,8 @@ class Project < ApplicationRecord
 
     StructuredLog.capture("CHECK_STATUS_CHANGE", { platform: platform, name: name, status_code: response.response_code }) if platform.downcase == "npm"
 
-    if platform.downcase == "packagist" && [302, 404].include?(response.response_code)
+    if platform.downcase.in?(%w[packagist npm]) && [302, 404].include?(response.response_code)
+      # NPM 302 happens with privately scoped packages, which should be considered Removed.
       update(status: "Removed", audit_comment: "Response #{response.response_code}")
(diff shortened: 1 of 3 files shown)


mpace965 (approved):

app/models/project.rb
```diff
@@ -668,7 +668,8 @@ def check_status
 
     StructuredLog.capture("CHECK_STATUS_CHANGE", { platform: platform, name: name, status_code: response.response_code }) if platform.downcase == "npm"
 
-    if platform.downcase == "packagist" && [302, 404].include?(response.response_code)
+    if platform.downcase.in?(%w[packagist npm]) && [302, 404].include?(response.response_code)
+      # NPM 302 happens with privately scoped packages, which should be considered Removed.
```
jhan217, line 672: Is there a difference between 302s that are removed or private? If so, would it better to mark the repository as private / set the project status to 'private'?

tiegz: it's a little ambiguous, for example I found a Project in the database that **used** to be public, but now it 302s. But then I tested with a fake URL and it gave the exact same response. I think NPM wants to obfuscate private packages by just returning the same responses, so there's no way to tell the difference.

jhan217: Yeah looks like the docs don't show any difference between unpublishing and making private. I'm fine treating them the same, but we should probably note this somewhere more public (at least to the PMs/internally).

PR #3474, opened 2024-09-18
PackageManager::NPM: treat "" requirement as "*" when fetching dependencies

tiegz:
we're getting this error whenever an NPM package has dependencies with the requirements `""`:

```
ActiveRecord::RecordInvalid 
Validation failed: Requirements can't be blank
```

treat it as a wildcard ("*") as we do everywhere else (this is [standardized recently in bibliothecary too](https://github.com/librariesio/bibliothecary/blob/main/lib/bibliothecary/dependency.rb#L57))


app/models/package_manager/npm.rb
```diff
@@ -145,16 +145,22 @@ def self.one_version(raw_project, version_string)
         .find { |v| v[:number] == version_string }
     end
 
-    def self.dependencies(_name, version, mapped_project)
+    def self.dependencies(name, version, mapped_project)
```
tiegz, line 148: separate bugfix: since we called this `_name` before, the logger below on line 151 was actually logging the **platform** name instead of the package name.

app/models/package_manager/npm.rb
```diff
@@ -40,7 +40,7 @@ def self.recent_names
     end
 
     def self.project(name)
-      get("http://registry.npmjs.org/#{name.gsub('/', '%2F')}")
+      get("https://registry.npmjs.org/#{name.gsub('/', '%2F')}")
```
tiegz, line 43: separate fix: we were getting 301s for NPM projects this whole time! Avoid that and go direct to HTTPS.

mpace965 (approved):

PR #3475, opened 2024-09-19
ruby 3.2.5

tiegz:
updates Libraries from ruby 3.1.5 to 3.2.5.


app/models/package_manager/packagist.rb
```diff
@@ -97,9 +97,9 @@ def self.versions(raw_project, _name)
     end
 
     def self.acceptable_versions(versions)
-      versions.select do |k, _v|
+      versions.select do |v|
         # See: https://getcomposer.org/doc/articles/versions.md#branches
-        (k =~ /^dev-.*/i).nil? && (k =~ /\.x-dev$/i).nil?
+        (v["version"] =~ /^dev-.*/i).nil? && (v["version"] =~ /\.x-dev$/i).nil?
```
tiegz, line 102: this was a bug that was hidden by the fact that `Hash#respond_to?(:=~)`, but it was caught during this upgrade because `Hash#respond_to?(:=~)` is no longer true.

jhan217 (approved):

PR #3476, opened 2024-09-19
Remove pry gems, in favor of IRB and the native debug gem.

tiegz:


tradiff (approved):

Gemfile
```diff
@@ -73,12 +74,12 @@ gem "rack-cors", require: "rack/cors"
 gem "rails", "7.0.8.1"
 gem "rails-timeago", git: "https://github.com/librariesio/rails-timeago"
 gem "rainbow", "~> 3"
-gem "rb-readline"
```
tiegz, line 76: this should also fix an IRB glitch that @tradiff and I were running into. Going to the beginning of the line wouldn't work for long lines.

PR #3477, opened 2024-09-19
Adds RateLimitService and uses it to throttle NPM status checks

tiegz:
to avoid overloading NPM when we check Project statuses (via CheckStatusWorker -> Project#check_status), this PR...

* adds a `RateLimitService` class based on some other code from @havocp ... it's just a unique key that stores a counter in Redis and covers an interval of time (defined by `period`).
* `Project#check_status`: 
  * uses `RateLimitService` for NPM projects by limiting ourselves to 10 concurrent checks at a time. 
* `CheckStatusWorker`: when `CheckStatusThrottled` is raised, retry in 5 minutes (since it's cheap to check the lock and try again)

### RateLimitService example

```
> rate_limiter = RateLimitService.new(what_to_limit: "foo", limit: 10, period: 1)
> 10.times {  rate_limiter.rate_limited { "no-op" }
=> 10
> 11.times {  rate_limiter.rate_limited { "no-op" }
lib/rate_limit_service.rb:46:in `rate_limited': Rate limit for 'foo' of 10 per 1 seconds exceeded by 1 (RateLimitService::OverLimitError)
        from (irb):4:in `block in <main>'
```


havocp (approved):
makes sense to me - let's try it!

PR #3482, opened 2024-09-27
Project#check_status: change the strategy of rate-limiting ourselves

tiegz:
(followup to https://github.com/librariesio/libraries.io/pull/3477)

this changes up our strategy of fetching NPM package urls:

* fetch the `"retry-after"` value from the NPM response
* instead of using `RateLimitService`, just store `"retry-after"` in Redis with an expiry
* if we are checking an NPM package url, and `"retry-after"` exists, reschedule the request to the proper time
* ensure that we store this value **per host**, since different hosts will receive different values

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -734,14 +734,14 @@ class Project < ApplicationRecord
     if current_retry_after.present?
       # Block this attempt if we have stored a "retry-after" value, per worker box.
       StructuredLog.capture("CHECK_STATUS_FAILURE", { platform: platform, name: name, retry_after: current_retry_after })
-      raise CheckStatusInternallyRateLimited, current_retry_after
+      raise CheckStatusInternallyRateLimited, current_retry_after.to_i
     else
       # If the key isn't set, then we're safe to attempt a request.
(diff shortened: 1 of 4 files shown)


app/models/project.rb
```diff
@@ -704,9 +722,38 @@ def check_status
...
-  rescue RateLimitService::OverLimitError => e
-    StructuredLog.capture("CHECK_STATUS_RATE_LIMITED", { platform: platform, name: name, exceeded_by: e.exceeded_by })
-    raise CheckStatusInternallyRateLimited
+  end
+
+  # Special-case of fetching package url for NPM, where we observe their rate-limiting and raise an error
+  # if we're exceeding their "retry-after" value.
+  def rate_limited_npm_request(url)
+    # This key needs to be host-specific so hosts don't overwrite each other's different values.
+    key = "#{Project::CHECK_STATUS_NPM_RETRY_AFTER}:#{Socket.gethostname}"
```
tiegz, line 731: fetching the hostname here is like 20 microseconds, so we probably don't need to worry about caching it in the process.

tradiff (approved):

PR #3479, opened 2024-09-20
 PackageManager::Conda: treat "" requirement as "*" when fetching dependencies 

tiegz:
similar to https://github.com/librariesio/libraries.io/pull/3474

PR #3483, opened 2024-09-30
 PackageManager::NuGet: treat "" requirement as "*" when fetching dependencies

tiegz:
similar to #3479 and #3474

### DIFF
--- a/spec/models/package_manager/nu_get_spec.rb
+++ b/spec/models/package_manager/nu_get_spec.rb
@@ -368,15 +368,35 @@ describe PackageManager::NuGet do
   describe ".dependencies" do
     let(:name) { "name" }
     let(:version) { "version" }
-    let(:raw_project) do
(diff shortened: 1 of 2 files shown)


spec/models/package_manager/nu_get_spec.rb
```diff
@@ -368,15 +368,35 @@
   describe ".dependencies" do
     let(:name) { "name" }
     let(:version) { "version" }
-    let(:raw_project) do
-      it "fetch the deps" do
```
tiegz, line 372: apparently I accidentally put an `it` block into a `let` a few months ago 🤦🏻

mpace965 (approved):

spec/models/package_manager/nu_get_spec.rb
jhan217: 🙀

PR #3484, opened 2024-10-01
Project: allow npm retries sooner by using ttl instead of get

tiegz:
(followup to https://github.com/librariesio/libraries.io/pull/3482)

* change `Project#rate_limited_npm_request()` to look at the TTL of its host-specific rate-limit key, instead of the value

### Example

if we cache a "retry-after: 60", we'd store the value as 60 and the TTL as 60. But as time goes on, the value is still 60 but the TTL is, say, 13. So if we look at the value, we're waiting longer to retry than we need to.

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -729,9 +729,10 @@ class Project < ApplicationRecord
   def rate_limited_npm_request(url)
     # This key needs to be host-specific so hosts don't overwrite each other's different values.
     key = "#{Project::CHECK_STATUS_NPM_RETRY_AFTER}:#{Socket.gethostname}"
-    current_retry_after = REDIS.get(key)
+    # The value will be the original 'retry-after', and the TTL will be the remaining 'retry-after', so use the latter.
+    current_retry_after = REDIS.ttl(key)
 
-    if current_retry_after.present?
+    if current_retry_after >= 0 # -2 == key does not exist, -1 == key exists without a TTL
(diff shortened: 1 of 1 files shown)


app/models/project.rb
```diff
@@ -729,9 +729,10 @@ def check_status
   def rate_limited_npm_request(url)
     # This key needs to be host-specific so hosts don't overwrite each other's different values.
     key = "#{Project::CHECK_STATUS_NPM_RETRY_AFTER}:#{Socket.gethostname}"
-    current_retry_after = REDIS.get(key)
+    # The value will be the original 'retry-after', and the TTL will be the remaining 'retry-after', so use the latter.
+    current_retry_after = REDIS.ttl(key)
 
-    if current_retry_after.present?
+    if current_retry_after >= 0 # -2 == key does not exist, -1 == key exists without a TTL
```
jhan217, line 735: Can `current_retry_after` ever be nil or empty or some other non-integer?

tiegz: nah, the docs say it'll only return a signed int: https://redis.io/docs/latest/commands/ttl/#resp2resp3-reply

jhan217: ok, just making sure we aren't going to get a pile of errors when their stuff goes down and we end up with a bunch of errors about nils

jhan217 (approved):

PR #3489, opened 2024-10-03
Fix N+1 in VersionUpdater by passing it preloaded Versions

tiegz:


app/models/version.rb
```diff
@@ -38,7 +38,6 @@ class Version < ApplicationRecord
   ].freeze
 
   validates :project_id, :number, presence: true
-  validates :number, uniqueness: { scope: :project_id }
```
tiegz, line 41: I think this validation had come up in a PR earlier, which I can't find now, but this should be fine since there's a UNIQUE index on these columns already

app/models/package_manager/base/version_updater.rb
```diff
@@ -49,7 +51,8 @@ def upsert_version_for_project!
       private
 
       def db_project_version
-        @db_project_version ||= @project.versions.find_or_initialize_by(number: @api_version_to_upsert.version_number)
+        @db_project_version ||= @preloaded_db_versions.find { |v| v.number == @api_version_to_upsert.version_number } ||
+                                @project.versions.new(number: @api_version_to_upsert.version_number)
```
tiegz, line 55: here's the optimization: instead of doing a `SELECT * FROM versions LIMIT 1` for each API version, we're now searching a preloaded list of versions in ruby-land.

pranav (approved):

PR #3493, opened 2024-10-09
Always return objects from PackageManager::Base#mapping() implementations

tiegz:
* `PackageManager::NPM.mapping()`: return a full Hash, even when project is removed
* `PackageManager::Packagist.mapping()`: return a full Hash, even when project is removed
* `PackageManager::Drupal.mapping()`: return a full Hash, even when project is removed
* implement `MappingBuilder` in a few `mapping()` methods that we missed before

the main fix here is to prevent us from returning early from `PackageManager::Base.update()` when the project data is incomplete, because returning early means their `last_synced_at` is never updated, and we'd continuously try to update them.


app/models/package_manager/npm.rb
```diff
@@ -82,11 +82,14 @@ def self.deprecation_info(db_project)
...
-      latest_version = raw_project["versions"].to_a.last[1]
-
-      repo = latest_version.fetch("repository", {})
+      latest_version = if raw_project["versions"].present?
+                         raw_project["versions"].to_a.last[1]
+                       else
+                         {} # "Removed" projects won't have a "versions" Hash
+                       end
+
+      repo = raw_project.fetch("repository", {})
```
jhan217, line 91:
```suggestion
      repo = raw_project.fetch("repository", {}) || latest_version.fetch("repository", {})
```

jhan217 (approved):

app/models/package_manager/npm.rb
tiegz: we gotta account for `repo == {}` too

jhan217 (approved):

PR #3494, opened 2024-10-15
Handle ActiveRecord::RecordNotUnique when creating projects too

tiegz:
this handles a pernicious race condition wherein a Project is created in another worker, right after the current Process has run the `Project.validates :name, uniqueness: true` validation but before it saves the Project.

### DIFF
--- a/app/models/package_manager/base.rb
+++ b/app/models/package_manager/base.rb
@@ -93,16 +93,21 @@ module PackageManager
       find(platform).try(:formatted_name) || platform
     end
 
-    private_class_method def self.ensure_project(mapped_project, reformat_repository_url: false, retries: 0)
-      db_project = Project.find_or_initialize_by({ name: mapped_project[:name], platform: db_platform })
-      db_project.reformat_repository_url if reformat_repository_url && !db_project.new_record?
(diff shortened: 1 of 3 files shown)


app/models/project.rb
```diff
@@ -124,7 +124,6 @@ def initialize(retry_after_seconds = nil)
   delegate :code_of_conduct_url, :contribution_guidelines_url, :funding_urls, :security_policy_url, to: :repository, allow_nil: true
 
   validates :name, :platform, presence: true
-  validates :name, uniqueness: { scope: :platform, case_sensitive: true }
```
tiegz, line 127: cc @havocp since we've talked about redundant AR validations + UNIQUE indices in the past. I personally love the idea of cleaning up these redundancies whenever we see them

app/models/package_manager/base.rb
```diff
@@ -93,21 +93,16 @@ def self.platform_name(platform)
       find(platform).try(:formatted_name) || platform
     end
 
-    private_class_method def self.ensure_project(mapped_project, reformat_repository_url: false)
+    private_class_method def self.ensure_project(mapped_project, reformat_repository_url: false, retries: 0)
```
mpace965, line 96: can `retries` be a local instead?

mpace965 (approved):

PR #3512, opened 2024-11-18
Update to rails 7.1.5

katzj:


tiegz (approved):
I usually do a `rails app:update` when upgrading just to make sure I didn't miss any config changes (although config/environments takes a long time to resolve with our existing modified files). That said, I don't think there are any big breaking changes 🤷🏻

PR #3520, opened 2024-12-05
Redirect various repo views

katzj:
We don't link to our version of info about repos live on the web anymore but there are still occasional times that people land on them manually or from outdated links. Let's just redirect to the repo host for those cases


havocp (approved):
sweet smell of less code

PR #3522, opened 2024-12-06
If there's a later pre-release, don't assume the package is deprecated

katzj:
While the latest stable may be marked deprecated, if there are later pre-releases, assume taht someone mistakenly published the pre-release to the main branch ala npm/attr-accept in Nov 2024


app/models/package_manager/npm.rb
```diff
@@ -73,6 +73,18 @@ def self.deprecation_info(db_project)
         is_deprecated = false
       end
 
+      # if the last stable version is deprecated but there have been pre-releases since, let's assume that
+      # the package isn't deprecated and there was a bad publish, eg attr-accept in Nov 2024
+      last_version = versions.last
+      if is_deprecated && last_version != last_stable_version
+        debugger
```
wenottingham, line 80: :suspect:

PR #3531, opened 2024-12-17
Api::Status#check: Include list of project ids when tracing rendering

tiegz:
mostly interested in reproducing slow rendering times like these:

Screenshot 2024-12-16 at 6 52 59 PM


havocp (approved):
I'm going to bet they have a huge release count

PR #3536, opened 2024-12-18
Add some monitoring around Go host fetching

tiegz:
theory: we may not need the `?go-get=1` fetching, which we're using to lookup Go Modules' aliases/redirects when we can't find them by name in the the database. 

my reasoning is:

* we've cleaned up and standardized Libraries' Go model a lot since [this code was first added](https://github.com/librariesio/libraries.io/pull/2386)
* we're actually picking the VCS repository part of the "?go-get=1" response instead of the actual Go Module name from that response, but we don't store or index Go Modules by their VCS repositories (i.e. this is probably doing nothing for us)

I've added some logging that should help us make a conclusion, which I'll comment on in the diff...


app/models/package_manager/go.rb
```diff
@@ -9,6 +9,10 @@ class Go < Base
     KNOWN_HOSTS = [
       "bitbucket.org",
       "github.com",
+      "gitlab.com",
+      "gitee.com",
+      "gopkg.in",
+      "k8s.io",
```
tiegz, line 15: these are some of the more popular Go hosts in Libraries, in the long-tail following github.com

app/models/package_manager/go.rb
```diff
@@ -243,34 +247,48 @@ def self.project_find_names(name)
...
-          &.last
-          &.sub(/https?:\/\//, "")
-
-        chosen_name = go_import&.start_with?(*KNOWN_HOSTS) ? [go_import] : [name]
+        # TODO: if nothing is being logged in GO_MODULES_PROJECT_FIND_NAMES_SUCCESS
+        # after a while, we might want to start using root_path instead of
+        # repo_url, if anything.
+        root_path = nil
+        vcs = nil
+        repo_url = nil
```
tiegz, line 255: start tracking all 3 values from the "?go-get=1" meta tag, to help prove that this code is not useful

app/models/package_manager/go.rb
```diff
@@ -243,34 +247,48 @@ def self.project_find_names(name)
...
           "GO_MODULES_PROJECT_FIND_NAMES",
           {
             name: name,
-            found_name: go_import,
+            root_path: root_path,
+            vcs: vcs,
+            repo_url: repo_url,
+            found_name: repo_url,
             chosen_name: chosen_name,
+            go_lookup_duration: ms,
```
tiegz, line 278: would be helpful to aggregate how long we're spending in these "?go-get=1" requests overall

app/models/package_manager/go.rb
```diff
@@ -243,34 +247,48 @@ def self.project_find_names(name)
...
         # We can get here from go modules that don't exist anymore, or having server troubles:
         # Fallback to the given name, cache the host as "bad" for a day,
         # log it (to analyze later) and notify us to be safe.
-        Rails.logger.info "[Caching unreachable go host] name=#{name}"
+        StructuredLog.capture(
+          "GO_MODULES_PROJECT_FIND_NAMES_UNREACHABLE_HOST",
+          { name: name }
+        )
         Rails.cache.write("unreachable-go-hosts:#{host}", true, ex: 1.day)
-        Bugsnag.notify(e)
```
tiegz, line 273: no need for a Bugsnag, really

app/models/project.rb
```diff
@@ -542,11 +542,23 @@ def self.find_best!(platform, name, includes = [])
...
+
+    # Track these lookups for a while and see if they actually work. Currently
+    # we're picking the last "repo-url" part of ?go-get=1, which is the VCS url,
+    # but we'd want the first "root-path" part to get actual go module name.
+    if project && platform == "Go"
+      StructuredLog.capture(
+        "GO_MODULES_PROJECT_FIND_NAMES_SUCCESS",
+        { name: name, project_find_names: names, project_id: project.id }
+      )
+    end
```
tiegz, line 559: the absence of these logs should prove that the "?go-get=1" code is not doing anything for us

havocp (approved):

PR #3538, opened 2024-12-21
Maven: Fix property priority

tradiff:
#3537 introduced a regression on properties where the parent would win in a conflict. This PR corrects that regression and adds a spec to confirm the proper behavior.

### DIFF
--- a/spec/fixtures/proto-google-common-protos-0.1.9.pom
+++ b/spec/fixtures/proto-google-common-protos-0.1.9.pom
@@ -36,7 +36,6 @@
   </developers>
   <scm>
     <connection>scm:git:https://github.com/googleapis/googleapis</connection>
-    <url>${scm.url}</url>
   </scm>
   <dependencies>
     <dependency>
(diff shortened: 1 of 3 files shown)


spec/fixtures/proto-google-common-protos-0.1.9.pom
```diff
@@ -36,7 +36,6 @@
   </developers>
   <scm>
     <connection>scm:git:https://github.com/googleapis/googleapis</connection>
-    <url>${scm.url}</url>
```
tradiff, line 39: The fix in this PR broke a test case that was testing scm.url shows up in the homepage. But I believe this was an invalid pom.xml, since it's a variable that refers to itself. If I try to do something similar, mvn throws an error about a recursive element. I've corrected this invalid file, and added another test to confirm `scm.url` is getting reflected in the properties (which is what I think the intent of this was about).

havocp (approved):

PR #3544, opened 2024-12-30
ruby 2.3.6, bundler 2.6.2, add checksums to lockfile, ci simplification

tiegz:
checksums announcement: https://bundler.io/blog/2024/12/19/bundler-v2-6.html

the checksums were added manually to Gemfile.lock via `bundle lock --add-checksums`

### Update bundler to 2.6.1+ locally

`bundle update --bundler`


katzj:
In favor but circle seems unhappy?

tiegz:
@katzj yeah, I was struggling with Circle on this 😭 Might pick it up again later today to retry

.circleci/config.yml
```diff
@@ -1,14 +1,14 @@
...
+  ruby: circleci/ruby@2.3.1
 
 executors:
   # TODO: we don't need ES/PG/Redis for "lint" job, so we could create another
   # stripped-down executor if we want?
   libraries:
     working_directory: ~/libraries
     docker:
-      - image: ruby:3.2.5-slim-bullseye
+      - image: cimg/ruby:3.2-node
```
tiegz, line 11: I think we just needed the custom packages below for node, so this switches to the node image

.circleci/config.yml
```diff
@@ -23,34 +23,16 @@ executors:
...
-
-      - checkout:
-          path: ~/libraries
-
-      - run:
-          name: Prep ruby
-          command: |
-            gem -v
-            bundle -v
-            bundle config path vendor/bundle
```
tiegz, line 50: this is already set by `ruby/install-deps`

havocp (approved):

PR #3545, opened 2024-12-31
PackageManager::Go.project_find_names(): remove this method

tiegz:
(followup to https://github.com/librariesio/libraries.io/pull/3536 )

after monitoring the logs from  that PR ☝🏻 for the past 10 days, it's safe to say we can remove the `PackageManager::Go.project_find_names()` method:

* this code that requests `?go-get=1` and fetches the "repo url" from the `<meta>` tag in the response was added in 2019, before we standardized on `pkg.go.dev` and consolidated all Go Projects on libraries to Go Modules only
* by picking the "repo url" part of the `?go-get=1` response, we're potentially returning info for a **different** Project than we should be. 
  * for instance, `gomodules.xyz/stow` is on `pkg.go.dev` but was not on Libraries, because it was last published in 2020 (before we started scraping `pkg.go.dev`). BUT, the "repo url" from `?go-get=1` for it is `github.com/gomodules/stow`, which IS on Libraries and pkg.go.dev but was last published in 2019! So we're returning old/outdated info for the wrong Project when we could do better by just ingesting `gomodules.xyz/stow` on Libraries (which I did manually) and returning that instead.
* the `?go-get=1` technique AFAICT is really for the Go command to fetch a Go Module's contents when it's behind a proxy. But Libraries is only concerned with Go Modules that are on `pkg.go.dev`, so we don't really need to look for the VCS repository behind them.


app/models/package_manager/go.rb
```diff
@@ -235,67 +213,6 @@ def self.dependencies(name, version, _mapped_project)
...
-        StructuredLog.capture(
-          "GO_MODULES_PROJECT_FIND_NAMES",
-          {
-            name: name,
-            root_path: root_path,
-            vcs: vcs,
-            repo_url: repo_url,
-            found_name: repo_url,
-            chosen_name: chosen_name,
-            go_lookup_duration: ms,
```
tiegz, line 278: these averaged about 1-2 sec, FWIW

tradiff (approved):

PR #3554, opened 2025-03-21
Remove collections route

johnbintz-tidelift:
We don't link to it anywhere anymore.


app/controllers/collections_controller.rb
```diff
@@ -1,24 +0,0 @@
-# frozen_string_literal: true
-
-class CollectionsController < ApplicationController
```
tiegz, line 3: could remove the `app/views/collections` folder too

tiegz (approved):

PR #3557, opened 2025-04-01
Cache Popular Languages and Platforms for Explore Page

mikeyoung85:
These are some fairly expensive queries to populate the lists of languages and platforms on the `/explore` page. These lists should not be changing very often so this PR introduces a caching layer for those lists to help reduce db usage.

### DIFF
--- a/app/controllers/explore_controller.rb
+++ b/app/controllers/explore_controller.rb
@@ -2,11 +2,14 @@
 
 class ExploreController < ApplicationController
   def index
-    @platforms = Project.maintained.group(:platform).order("count_id DESC").limit(28).count("id").keys
-    @languages = Project.maintained.where.not(language: nil).group(:language).order("count_id DESC").limit(21).count("id").keys
+    @platforms = Rails.cache.fetch("explore:platforms", expires_in: 1.day, race_condition_ttl: 2.minutes) do
+      Project.maintained.group(:platform).order("count_id DESC").limit(28).count("id").keys
+    end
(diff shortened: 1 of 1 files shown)


app/controllers/explore_controller.rb
```diff
@@ -2,11 +2,14 @@
 
 class ExploreController < ApplicationController
   def index
-    @platforms = Project.maintained.group(:platform).order("count_id DESC").limit(28).count("id").keys
-    @languages = Project.maintained.where.not(language: nil).group(:language).order("count_id DESC").limit(21).count("id").keys
-
-    project_scope = Project.includes(:repository).maintained.with_description
+    @platforms = Rails.cache.fetch("explore:platforms", expires_in: 1.day, race_condition_ttl: 2.minutes) do
```
mikeyoung85, line 5: We could probably cache these for longer than 1 day if we wanted to, but once a day for the queries seemed reasonable.

app/controllers/explore_controller.rb
```diff
@@ -2,11 +2,14 @@
...
+      Project.maintained.group(:platform).order("count_id DESC").limit(28).count("id").keys
+    end
+    @languages = Rails.cache.fetch("explore:languages", expires_in: 1.day, race_condition_ttl: 2.minutes) do
+      Project.maintained.where.not(language: nil).group(:language).order("count_id DESC").limit(21).count("id").keys
+    end
 
+    project_scope = Project.maintained.with_description
```
mikeyoung85, line 12: This used to also include all the repositories for the `Project` but I don't see any usage of `Repository` fields in the query.

katzj (approved):

PR #3561, opened 2025-04-07
Index dependencies.id_new and backfill all uuids

havocp:
Backfill for this is now done and it should be ready to merge.

### DIFF
--- a/db/migrate/20250403234704_backfill_dependencies_new_id.rb
+++ b/db/migrate/20250403234704_backfill_dependencies_new_id.rb
@@ -1,0 +1,34 @@
+# frozen_string_literal: true
+
+class BackfillDependenciesNewId < ActiveRecord::Migration[7.1]
+  disable_ddl_transaction!
+
+  def change
+    # Step 1: Add a unique index; since all values are null, it should be valid already.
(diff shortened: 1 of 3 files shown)


db/migrate/20250403234704_backfill_dependencies_new_id.rb
```diff
@@ -0,0 +1,34 @@
...
+
+    # Step 2: Fill in all null id_new values with a random uuid using a SQL update statement
+    safety_assured do
+      reversible do |dir|
+        dir.up do
+          execute <<-SQL
+            UPDATE dependencies
+            SET id_new = gen_random_uuid()
+            WHERE id_new IS NULL;
+          SQL
```
tiegz, line 20: Generating these, updating them, and then the index rebuilding seems like it could exceed the CI timeout. Worth doing in a standalone rake task or something like that, for production?

havocp: oh it will definitely exceed the timeout. I'm doing it manually before merging this ...

macowie (approved):

tiegz (approved):

PR #3568, opened 2025-06-11
Project: use PEP530 name normalization for PyPi lookups

tiegz:
Currently Libraries looks up Pypi projects using 3 lookup names: "FOO-BAR_BAZ" (the requested name), "foo-bar-baz", and "foo_bar_baz".

This PR goes further by using [PEP 503's name normalization rules](https://packaging.python.org/en/latest/specifications/name-normalization/): replace any run of "_-." with a single "-" and lower case it.

I've checked every PyPi Project on Libraries right now, and fetching the normalized name from PyPI's API returns the exact same project for the name in our database, which implies that any PyPi project can be queried by its normalized name (at the very least, this can be guaranteed for the "pypi.org/simple/" API, because of PEP 503).

### Changes

* `Project`:
  * adds a `find_all_with_package_manager!` method that does platform-specific alternate name bulk lookup
    * for PyPi, use the PEP503 name normalization regexp on both lookup and database name to match them
      * (we could also avoid the latter by adding a `normalized_name` column but performance on the database lookup seems fine)
    * for everything else do a lowercase'ed name lookup (as we currently do)
  * adds a `find_with_package_manager!` method that just takes the first result from `find_all_with_package_manager!`, or raises RecordNotFound
  * replace the alternate name lookup usage of `project_find_names` with `find_with_package_manager!`
* `PackageStatusQuery`:  (used for bulk project lookup in controllers)
  * replace usage of `project_find_names` with `find_all_with_package_manager!`
* `PackageManager`:
  * `Base`: remove `project_find_names()`, as it's no longer used

### Example

Passing these platform/name pairs to the `/api/check` bulk API searching for [winui3-Microsoft.UI.Xaml.Interop/](https://pypi.org/project/winui3-Microsoft.UI.Xaml.Interop/):
* "pypi"/"winui3-microsoft-ui-xaml-interop"
* "pypi"/"not-found-package"

#### Before

Returns name/canonical_names:

```
* nothing
```

#### After

Returns list of name/canonical_names:

```
[["winui3-microsoft-ui-xaml-interop", "winui3-Microsoft.UI.Xaml.Interop"]]
```

### Existing Limitation

We have a limitation where `Api::BulkProjectController#project_names` combines all the lookup names from `ProjectStatusQuery`, so a search for the project "foo"  with `["FOO", "foo", "foO"]` would return something like `[["FOO", "foo"], ["FOO", "foo"], ["FOO", "foo"]]`

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,7 +7,7 @@ gem "active_model_serializers"
 gem "api-pagination"
 gem "asciidoctor"
 gem "audited"
-gem "bibliothecary", "~> 11.0.0"
+gem "bibliothecary", "~> 12.3.1"
 gem "bitbucket_rest_api", git: "https://github.com/librariesio/bitbucket"
 gem "bootsnap", require: false
 gem "bootstrap-sass"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -121,9 +121,10 @@ GEM
       execjs
     base64 (0.1.1)
     benchmark (0.4.0)
-    bibliothecary (11.0.1)
+    bibliothecary (12.3.1)
       commander
       deb_control
+      json (~> 2.8)
       librariesio-gem-parser
       ox (>= 2.8.1)
       packageurl-ruby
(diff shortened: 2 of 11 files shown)


havocp (approved):
Nice this looks promising!

